PR #2, opened 2019-09-29
Add several Cisco devices

jvanderaa:
Added several Cisco devices.


jeremystretch:
Please note that we are not yet accepting PRs as several details concerning the import logic still need to be worked out.

PR #7, opened 2019-12-26
Add device types1

jvanderaa:


cisco/asr1002-hx.yaml
```diff
@@ -0,0 +1,62 @@
...
+u_height: 2
+is_full_depth: false
+console_ports:
+  - name: con 0
+power_ports:
+  - name: P0
+  - name: P1
+interfaces:
+  - name: GigabitEthernet0/0/0
+    type: SFP (1GE)
```
jeremystretch: Interface types must reference the new (v2.7+) slugs; see the [schema validation](https://github.com/netbox-community/devicetype-library/blob/master/tests/schema.json) for all valid values.

(This should have caused the automated tests to fail, but we weren't catching `.yml` suffixes. This has been fixed.)

cisco/asr1002-hx.yaml
```diff
@@ -0,0 +1,62 @@
...
+    mgmt_only: false
```
jeremystretch: `mgmt_only` is False by default; you can omit this line from non-management interfaces for brevity.

cisco/asr1002-hx.yaml
```diff
@@ -0,0 +1,62 @@
+manufacturer: Cisco
+model: ASR1002-HX
+slug: asr1002-hx
+u_height: 2
+is_full_depth: false
```
jeremystretch: All devices with front-to-back airflow should be considered full-depth. (Installing a device in the same unit on the opposite face would inhibit airflow.)

jeremystretch:
Also, please use `Cisco` for the manufacturer directory (capital C).

cisco/asr1002-hx.yaml
jvanderaa: Set this to true.

cisco/asr1002-hx.yaml
jvanderaa: Updated on all the interfaces, should be looking better.

cisco/asr1002-hx.yaml
jvanderaa: Makes sense. Sorry for the PR in that setup. Went too quick.

jeremystretch:
@jvanderaa Looks like you need to update your fork. There should be a `device-types` directory that holds all the manufacturers. Once you've done that, the validation tests should run as expected.

jeremystretch:
This PR is superseded by #16

PR #9, opened 2019-12-26
Create Device Type Cisco ASR1001-HX

collin-clark:


device-types/Cisco/AIR-AP2802I-A-K9.yaml
```diff
@@ -0,0 +1,11 @@
+manufacturer: Cisco
+model: AIR-AP2802I-A-K9
+slug: air-ap2802i-a-k9
+u_height: 0
+is_full_depth: false
+interfaces:
+  - name: gi0
```
jeremystretch: This should be the full interface name (i.e. `GigabitEthernet0`).

device-types/Cisco/ASR1001-HX.yaml
```diff
@@ -0,0 +1,51 @@
...
+    type: 1000base-t
+  - name: GigabitEthernet0/0/4
+    type: 1000base-t
+  - name: GigabitEthernet0/0/5
+    type: 1000base-t
+  - name: GigabitEthernet0/0/6
+    type: 1000base-t
+  - name: GigabitEthernet0/0/7
+    type: 1000base-t
+  - name: Te0/1/0
```
jeremystretch: Same as above; be sure to use full interface names consistently.

device-types/Eaton/5PX3000RTN.yaml
```diff
@@ -0,0 +1,29 @@
...
+    type: nema-5-20r
+  - name: Primary Group - Outlet2
+    type: nema-5-20r
+  - name: L5-30R
+    type: nema-5-30r
+power-port:
+  - name: L530R
+    type: nema-5-30p
+device_bays:
+  - name: EBM
```
jeremystretch: This device type does not appear to house child devices. (Remember that a child device is one with its own independent control/management plane.)

device-types/Cisco/ASR1001-HX.yaml
collin-clark: Updated Interface Names

device-types/Cisco/AIR-AP2802I-A-K9.yaml
collin-clark: Update interface name

device-types/Eaton/5PX3000RTN.yaml
collin-clark: Removed bay

device-types/Cisco/ASR1001-X.yaml
```diff
@@ -0,0 +1,37 @@
...
+  - name: con 0
+    type: rj-45
+power-ports:
+  - name: PSU0
+    type: iec-60320-c14
+    maximum_draw: 250
+  - name: PSU1
+    type: iec-60320-c14
+    maximum_draw: 250
+device-bay:
```
jeremystretch: I'm not familiar with this device, but this appears to be a regular expansion slot (not modeled) and not a device bay.

collin-clark: Maybe I don't understand what is meant by a device bay. I see a device bay as a slot in a network device that can take a module such as a WIC/HWIC/NM/NIM/SPA that provides additional interfaces (types and quantities). Is that incorrect usage?

device-types/Cisco/AIR-AP2802I-A-K9.yaml
```diff
@@ -0,0 +1,11 @@
+manufacturer: Cisco
+model: AIR-AP2802I-A-K9
+slug: air-ap2802i-a-k9
+u_height: 0
+is_full_depth: false
+interfaces:
```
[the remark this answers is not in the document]
collin-clark: I don't have access to a autonomous AP to verify the interface nomenclature so I will delete this device type.

device-types/Cisco/ASR1001-X.yaml
jeremystretch: Every component is described in [the docs](https://netbox.readthedocs.io/en/stable/core-functionality/devices/#device-bays):

> Device bays represent the ability of a device to house child devices. For example, you might install four blade servers into a 2U chassis. The chassis would appear in the rack elevation as a 2U device with four device bays. Each server within it would be defined as a 0U device installed in one of the device bays. Child devices do not appear within rack elevations or the "Non-Racked Devices" list within the rack view.
>
> Child devices are first-class Devices in their own right: that is, fully independent managed entities which don't share any control plane with the parent. Just like normal devices, child devices have their own platform (OS), role, tags, and interfaces. You cannot create a LAG between interfaces in different child devices.
>
> Therefore, Device bays are not suitable for modeling chassis-based switches and routers. These should instead be modeled as a single Device, with the line cards as Inventory Items.

PR #13, opened 2019-12-27
Juniper ex2200 48t

caskings:
Juniper EX2200 switch non poe version


device-types/Juniper/EX2200-48T-4G.yaml
```diff
@@ -0,0 +1,114 @@
...
+  - name: ge-0/0/44
+    type: 1000base-t
+  - name: ge-0/0/45
+    type: 1000base-t
+  - name: ge-0/0/46
+    type: 1000base-t
+  - name: ge-0/0/47
+    type: 1000base-t
+  - name: ge-0/1/0
+    type: 1000base-sfp
```
jeremystretch: `1000base-sfp` should be `1000base-x-sfp`

PR #15, opened 2019-12-27
Add Juniper QFX5110-48S-4C

steffann:
Juniper automatically renames interfaces based on the transceiver installed. A 10G `xe-` port will therefore automatically be renamed to `ge-` the moment a 1G transceiver is plugged in. I included the default names for transceivers with the maximum speed installed, assuming that someone who buys an 48x10G switch will actually use it as 10G.

### DIFF
--- a/device-types/Juniper/QFX5110-48S-4C.yaml
+++ b/device-types/Juniper/QFX5110-48S-4C.yaml
@@ -1,0 +1,121 @@
+manufacturer: Juniper
+model: QFX5110-48S-4C
+slug: qfx5110-48s-4c
+interfaces:
+  - name: fxp0
+    type: 1000base-t
+    mgmt_only: true
+  - name: xe-0/0/0
+    type: 10gbase-x-sfpp
+  - name: xe-0/0/1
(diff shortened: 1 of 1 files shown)


steffann:
If I did this one correctly I'll add more QFX switch models

device-types/Juniper/QFX5110-48S-4C.yaml
```diff
@@ -0,0 +1,121 @@
...
+    type: 10gbase-x-sfpp
+  - name: xe-0/0/45
+    type: 10gbase-x-sfpp
+  - name: xe-0/0/46
+    type: 10gbase-x-sfpp
+  - name: xe-0/0/47
+    type: 10gbase-x-sfpp
+  - name: xe-0/0/48
+    type: 100gbase-x-qsfp28
+  - name: xe-0/0/49
```
jeremystretch: Shouldn't the 100GE interfaces be `et-X/Y/Z?`

steffann: Oops, well spotted!

PR #16, opened 2019-12-27
Fixed up directory structure on my local branch

jvanderaa:
This should be the updated version with all of the changes requested. We can delete the other PR.


jeremystretch:
Looks like tests are failing on some invalid interface types (you can see the exact failures in the [test logs](https://github.com/netbox-community/devicetype-library/pull/16/checks?check_run_id=365235407) on GitHub).

device-types/Cisco/asr1002-hx.yaml
```diff
@@ -0,0 +1,46 @@
...
+    type: 1000base-x-sfp
+  - name: GigabitEthernet0/0/4
+    type: 1000base-x-sfp
+  - name: GigabitEthernet0/0/5
+    type: 1000base-x-sfp
+  - name: GigabitEthernet0/0/6
+    type: 1000base-x-sfp
+  - name: GigabitEthernet0/0/7
+    type: 1000base-x-sfp
+  - name: Te0/0/0
```
jeremystretch: Needs to be the full form of the interface name (i.e. `TenGigabitEthernet`).

device-types/Cisco/c9300-24p.yaml
```diff
@@ -0,0 +1,62 @@
+manufacturer: Cisco
+model: C9300-24P
+part_number: C9300-24P
+slug: c9300-24p
+u_height: 1
+is_full_depth: false
```
jeremystretch: Devices with front-to-back airflow should be considered full-depth.

device-types/Cisco/asr1002-hx.yaml
jvanderaa: Makes sense. I was going back and forth in my mind on these. A `show ip interface brief` would show it in the shortened fashion (not sure why). Versus the actual show interface. Will get that fixed up and checked on the other ones tonight.

PR #20, opened 2019-12-27
Create N7K-7706.yaml

collin-clark:


trrunde:
> Supervisor modules (N77-SUP2E, N77-SUP3E) are not device types. Device types define entirely self-contained devices.

How should linecard/supervisor modules be defined if they are not a device type?

collin-clark:
> Supervisor modules (N77-SUP2E, N77-SUP3E) are not device types. Device types define entirely self-contained devices.

Understood. Is it a future plan to also YAML'ize inventory items as well? 

Also I think some (many?) are using parts of Netbox different than what you anticipated. For example I will build a device type of a chassis. I will also build device types of the supervisor and line cards. I make the chassis a parent and it has X number of device bays (slots). I then child the supervisor and install it in the chassis. It may be wrong, but it works pretty good. I can see what each line card has for interfaces, connect to them, etc. The only downfall is that the interfaces are not seen directly under the chassis device. I have to click on the device type name under the device bay to view them. In the Supervisors above I assume they would be put under inventory of the chassis? How would I then assign an IP and connect it to the rest of the network?

jeremystretch:
> I will also build device types of the supervisor and line cards

This is using NetBox incorrectly. This PR is not the place to have this discussion. If you need assistance figuring out how to model things, please post to the NetBox mailing list.

jeremystretch:
Closing this PR for now. Please feel free to resubmit once the invalid device types have been removed.

PR #24, opened 2019-12-28
Juniper srx

trrunde:
Juniper SRX240H2

### DIFF
--- a/device-types/Cisco/N5K-C5548UP.yaml
+++ b/device-types/Cisco/N5K-C5548UP.yaml
@@ -1,0 +1,81 @@
+manufacturer: Cisco
+model: Nexus 5548UP
+slug: n5k-c5548up
+part_number: N5K-C5548UP
+interfaces:
+  - name: mgmt0
+    type: 1000base-t
+    mgmt_only: true
+  - name: Ethernet1/1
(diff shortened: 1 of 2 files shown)


device-types/Juniper/SRX240H2.yaml
```diff
@@ -0,0 +1,47 @@
...
+    type: 1000base-t
+  - name: ge-0/0/15
+    type: 1000base-t
+power-ports:
+  - name: PSU0
+    type: iec-60320-c14
+console-ports:
+  - name: Console
+    type: rj-45
+device-bays:
```
jeremystretch: PIMs are not devices and thus PIM slots are not device bays.

device-types/Cisco/N5K-C5548UP.yaml
```diff
@@ -0,0 +1,81 @@
+manufacturer: Cisco
```
jeremystretch: This definition was submitted in PR #23. When you create new branches, be sure to base each of them off of the `master` branch to avoid redundant changes.

trrunde: sorry about that, will check better in future.

PR #31, opened 2020-01-01
Added some cisco NCS boxes

trrunde:


device-types/Cisco/NCS-5501-SE.yaml
```diff
@@ -0,0 +1,106 @@
+manufacturer: Cisco
+model: NCS 5501 SE
```
jeremystretch: From Cisco's docs, it looks like these model numbers should have dashes (same for all the NCS definitions):
```suggestion
model: NCS 5501-SE
```

trrunde: updated model name.

PR #35, opened 2020-01-03
Update CRS328-4C-20S-4S+RM.yaml

tioan:
correct port type of combo ports


jeremystretch:
Neither type is more "correct" than the other, as either can be valid. A user can modify the interface types during or after import as needed.

PR #41, opened 2020-01-03
APC pdus

ryanmerolle:
Models Added:
- AP8868
- AP8886
- AP8932

Additionally had to update the json.schema to include for the above models:
- California Style Connector type to power-ports and power-outlets
- rj12 type for console-ports


ryanmerolle:
main netbox PR submitted under netbox-community/netbox#3844

jeremystretch:
Thanks for updating the schema as well!

PR #64, opened 2020-02-25
Additions of Avocent, Cisco ACI, and Meraki 

jemery27:


jeremystretch:
Validation is failing because you've defined an empty subdevice role:

```
subdevice_role: ''
```

Just omit these and other fields you've set to empty strings: If they have no value there's no reason to include them.

jemery27:
Ok cool thanks - just FYI the reason I included them is I created the templates in netbox and exported.

jeremystretch:
``
  allocated_draw: null
```

Same issue: If the field has no value, omit it from the definition. Otherwise it will fail validation.

jemery27:
Thank you

PR #79, opened 2020-03-24
Add Cisco Catalyst Digital Building Switch

kaechele:
These devices are a bit special. See the comments in the files.

Signed-off-by: Felix Kaechele <fkaechel@cisco.com>


jeremystretch:
> actual height of the rack mount chassis is 5U but it fits 5 CDB horizontally

This needs to be modeled as a child device inside the 5U parent chassis

kaechele:
I fixed this by adding the chassis as well as adding the subdevice roles.
I had to put the power port numbers in quotes as the schema only allows strings. But I tried to match the writing on the physical device as closely as possible:
CDB-MNT-RACK5-C14

PR #82, opened 2020-03-27
added one 7048 and two 7050 model switches

wbrandst:
add yaml for Arista 7048T-A 
add yaml for Arista 7050S-52 
add yaml for Arista 7050S-64


ryanmerolle:
@wbrandst apologies for the delay here.  The Arista 7048 and 7050 models look fine, but it seems you included some other models in your PR.  Please clarify if you are looking for those to be added to the PR.  If so, we should reference those as being added.

ryanmerolle:
@wbrandst please advise

PR #91, opened 2020-03-30
Add 2 SuperMicro boards

garbled1:
These give you the base server so you can modify from there.

### DIFF
--- a/device-types/SuperMicro/X8DTH-iF.yaml
+++ b/device-types/SuperMicro/X8DTH-iF.yaml
@@ -1,0 +1,26 @@
+manufacturer: SuperMicro
+model: X8DTH-iF Server
+slug: x8dth-if
+u_height: 4
+is_full_depth: true
+console-ports:
+  - name: VGA
+    type: other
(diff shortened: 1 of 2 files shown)


MartinVerges:
does it make sense to write `u_height: 4`? I think there is something like a need to specify a chassis to have the size as well.

garbled1:
I feel like the important bit of a system is all the various network connections and USB and whatnot.  The case it's sitting in is just a box.  So the idea here is you take the board, import it, and then modify it locally before using.  (like fix u_height and power, or maybe you add a NIC card in, etc etc.  PC's are far more modular)

I'm not sure it's really reasonable to do a chassis and then load the board as a child.  That makes sense, but it would be obnoxious to work with in practice.

DanSheps:
We are only going to import fully fleshed out devices at this point I think.  If you want to apply specific models to this I will leave it open for you to alter.

garbled1:
Yeah, this one can probably be closed out..  it might not make sense to import.

PR #102, opened 2020-04-15
Add Cisco ASR RSP (A9K-RSP880-LT-TR)

YAPNie:


ryanmerolle:
This is an ASR inventory item and not a chassis or device that can run on its own.

PR #103, opened 2020-04-16
Add Cisco ASR Line Card (A9K-24X10GE-1G-TR)

YAPNie:


ryanmerolle:
This is an ASR inventory item and not a chassis or device that can run on its own.

PR #108, opened 2020-04-24
Correct Interface 15 for APC AP8853

mciuraru:


ryanmerolle:
Fixed already in #94 which was just merged.

PR #112, opened 2020-04-25
Add multiple device types (Cisco, Infoblox, Fortigate, HPE)

deku-m:
I made a couple of device types which we use..


deku-m:
Can the "+"  be supported in naming and model/part-number/slug?

ryanmerolle:
@deku-m I love the contribution here.  It's hard to review given all the different models you have bulked into one PR.  Would you be able to separate this out logically so I can review and approve quickly for each grouping.

deku-m:
> 
> 
> @deku-m I love the contribution here. It's hard to review given all the different models you have bulked into one PR. Would you be able to separate this out logically so I can review and approve quickly for each grouping.

I think i need to create new branches for each new device-type then isnt it?

deku-m:
Making a new one as per PR.

PR #116, opened 2020-04-29
Fixing interface port 43 numbering

rjdani:
Port 43 is mis-numbered which causes an error when trying to import


ryanmerolle:
Fixed already in #94 which was just merged.

PR #123, opened 2020-05-08
Add Fortinet FG-50E

tnagafuchi:


ryanmerolle:
@tnagafuchi apologies for the delay here. The FG-50E looks fine, but it seems you included some other models in your PR. Please clarify if you are looking for those to be added to the PR. If so, we should reference those as being added.

tnagafuchi:
Other models included in PR have been deleted.

PR #126, opened 2020-05-12
Add Juniper EX2300-C-12P/T

avermeer-tc:
Adding support for the EX2300-C-12P and 12T (PoE and Non PoE)

Spec sheet: https://www.juniper.net/assets/us/en/local/pdf/datasheets/1000580-en.pdf


mpeterson-tc:
I believe the units also have a USB console port

PR #127, opened 2020-05-19
Add Cisco WS-C4900M switch

BarbarossaTM:
This adds the chassis of the Cisco Catalyst 4900M switches.

### DIFF
--- a/device-types/Cisco/ws-c4900m.yaml
+++ b/device-types/Cisco/ws-c4900m.yaml
@@ -1,0 +1,45 @@
+manufacturer: Cisco
+model: Catalyst C4900M
+slug: ws-c4900m
+part_number: WS-C4900M
+u_height: 2
+is_full_depth: true
+subdevice_role: parent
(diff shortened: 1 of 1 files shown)


ryanmerolle:
Is this really a parent device and full depth?  This does not seem correct from the product page.

BarbarossaTM:
Hi @ryanmerolle 

you are right that the chassis as-is isn't full depth. But as this thing is a heavy beast it has rack rails like the nexus boxes which also have to be put into the end of the rack (see [0]). Therefore I made it full depth as the RU is blocked on both ends.

[0] https://www.cisco.com/c/en/us/td/docs/switches/lan/catalyst4900/4900M/installation/guide/4900Mod/04instal.html 

Best
Max

BarbarossaTM:
Hi @ryanmerolle,

do I take the rvisions needed tag correctly, that you want me to change the full depth to false?

Best
Max

PR #130, opened 2020-05-19
Add Netonix WISP switch WS-24-400A

BarbarossaTM:
This device has 4 dual-purpose ports (21-24) which can be 1000BaseT or SFP. I chose to model them as 1000BaseT as the default.

Are there any guidelines about that? Is there a better way?

### DIFF
--- a/device-types/Netonix/WS-24-400A.yaml
+++ b/device-types/Netonix/WS-24-400A.yaml
@@ -1,0 +1,91 @@
+manufacturer: Netonix
+model: WS-24-400A
+slug: ws-24-400a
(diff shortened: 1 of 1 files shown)


ryanmerolle:
@BarbarossaTM for those sort of dual purpose ports we just use the type of the sfp and leave it up to the user.  We also just add a comment next to said port(s).  I was quick to merge this in cleaning up the backlog.  I updated your model in commit https://github.com/netbox-community/devicetype-library/commit/c4b54753134952b1d2102d84a62430f0a1e7dbfd

PR #131, opened 2020-05-19
Add AP8865

ryanmerolle:
- Added APC PDU AP8865


ryanmerolle:
Relates to netbox-community/netbox#4665

PR #140, opened 2020-05-24
5130-48G-PoE+-4SFP+ (370W) EI

constantpinger:
This is my first contribution. I'm adding a single one to make sure I'm working to the expected model. Please feel free to provide ways to improve as I'll hopefully submit a lot more devices.

Note1: HP changed to HPE for the networking kit a couple of years ago. There might be confusion in the manufacturer name.

Note2: Full model name is **5130-48G-PoE+-4SFP+ (370W) EI** but I'm unsure what would be the best filename

I have left allocated draw null as I'm guessing that is site specific?


constantpinger:
failed tests, closing to investigate

PR #144, opened 2020-06-01
Add two ActionTec DSL modems.

garbled1:
There really is no good interface type for POTS so I added them as 100's.

### DIFF
--- a/device-types/ActionTec/C1000A.yaml
+++ b/device-types/ActionTec/C1000A.yaml
@@ -1,0 +1,24 @@
+manufacturer: ActionTec
+model: C1000A
+slug: c1000a
+u_height: 0
+is_full_depth: false
+power-ports:
+  - name: Power
+    type: nema-5-15p
+console-ports:
+  - name: USB
+    type: usb-a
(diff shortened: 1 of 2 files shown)


wtinetworkgear:
We defined our POTS modem lines as


  - name: Modem
    type: rj-11?




________________________________

garbled1:
That's not an interface type though, it's only available if I pick something like console port or front/rear port.  I kinda wanted to be able to assign IP's and such to it.   Sadly there are a few interface types missing for older equipment.  :)  I should file something on that...

DanSheps:
Please change the interface type to "other"

garbled1:
I don't see other listed in tests/schema.json.  Doesn't that have to be updated first?

DanSheps:
It looks like it needs to be updated, perhaps @ryanmerolle can handle that.

ryanmerolle:
@DanSheps @garbled1 I added the missing "other" interface as long as any other missing interfaces to the schema.json file.  You can make your updates and I will merge the PR.

PR #154, opened 2020-06-18
Creation of HPE folder and added 3 device-types

deku-m:
Added in the folder HPE several device-types:
- c7000.yaml
- HPE-SY480.yaml
- synergy-12000.yaml

Which include a chassis/blade (parent/child)

### DIFF
--- a/device-types/HPE/HPE-SY480.yaml
+++ b/device-types/HPE/HPE-SY480.yaml
@@ -1,0 +1,6 @@
+manufacturer: HPE
+model: HPE-SY480
+slug: HPE-SY480
+u_height: 0
+is_full_depth: true
+subdevice_role: child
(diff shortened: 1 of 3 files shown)


device-types/HPE/HPE-SY480.yaml
```diff
@@ -0,0 +1,6 @@
+manufacturer: HPE
+model: HPE-SY480
+slug: HPE-SY480
+u_height: 0
```
ryanmerolle, line 4: I am confused by this model.  From the docs it looks like almost the form factor of a blade (from a blade chassis). I guess you would line them up on a rack shelf?

deku-m: This is the blade yes, you put them in the parent chassis.

ryanmerolle:
I need coffee before I go through the backlog of PRs.  Thanks for correcting me @deku-m

PR #157, opened 2020-06-18
Revert "Add Netonix WISP switch WS-24-400A"

ryanmerolle:
Reverts netbox-community/devicetype-library#130

Merged too quickly.  Need to change the optional ports to SFP as opposed to hardwired base-t since we have the option for either.


ryanmerolle:
disregard

PR #180, opened 2020-07-23
added BIG-IP 5250v

andyewing:
This adds F5 as a Manufacturer and the BIG-IP 5250v platform as a devicetype. It should be fairly easy to use this devicetype to add additional BIG-IP models.


andyewing:
Can you tell me more about which checks failed? Is it because F5 doesn't exist as a manufacturer yet?

andyewing:
It seems pretty important to get F5 and the BIG-IP product line into the community library. Can I get more information as to why the check failed?

ryanmerolle:
I had to rerun your PR's test given the original commit was so old the logs were erased. If you looked at the log, it mentioned github actions / pytest did not like that interfaces were being recognized as a float data type.  Try quotes around the interface names to force python to see it as a string.

If you click the test report you would see this error:
`/home/runner/work/devicetype-library/devicetype-library/tests/test_definitions.py:54: Failed: device-types/F5/BIG-IP_5250v.yaml failed validation: 1.1 is not of type 'string'`

nahun (approved):
Added quotes around the interface names

PR #184, opened 2020-07-29
added SRX-1500 to device library

sy-na-github:
added srx-1500 to device library.  Used the 2 PSUs variant as the logic is that its easier to remove stuff than it is to add.


ryanmerolle:
I get what you are saying about PSUs being easier to remove than add, but the logic to date is to model all the various model number variants with only the included parts.  If people purchase/install extra parts, then they would add accordingly.

I would recommend we keep this model since that is how we have tried to keep this consistency to this point.

sy-na-github:
understood.  removed the extra psu.

PR #204, opened 2020-08-14
Dell poweredge

mddeff:
Proposed additions for Dell PowerEdge servers.  A few items of note:

- The PCI-E configurations (Half vs Full slot) are not an exhaustive list of available configurations; especially with the 7x0 and 8x0 series. It's what I've seen as the most common ones.  Users can alter templates to suit.
- The 4 NICs included on the daughter card are just basic 1000Base-T NICs in the template, but dozens of configurations are available from Dell (Half SFP+, Half Copper, etc).  Users can alter templates to suit.
- LFF indicates 3.5" HDD, SFF indicates 2.5" HDD/SSD as per Dell's naming conventions.
- Templates were tested against `2bf63b1c8bd9 (v2.7.8)`


mddeff:
Looks like this is failing because of
```
tests/test_definitions.py::test_definition[device-types/Cisco/N5K-C5648Q.yaml] FAILED [ 63%]
```
Which is currently in master (and not modified by my changes).  How should this be addressed?

ryanmerolle:
Per the NetBox docs:

> Device bays are not suitable for modeling line cards (such as those commonly found in chassis-based routers and switches), as these components depend on the control plane of the parent device to operate. Instead, line cards and similarly non-autonomous hardware should be modeled as inventory items within a device, with any associated interfaces or other components assigned directly to the device.

These bays should be used for blade chassis and other items were a discreet device with its own direct management that is housed by another device.

DanSheps:
As there has been no activity on this in awhile, I am going to close this PR.  You are free to re-open this PR once you have made the changes requested.

PR #208, opened 2020-08-20
Add device-type for Aruba Mobility controller 3400

CedricMoreau:
Add device-type for Aruba Mobility controller 3400 (4 ports gig0/0/0 to 0/0/3 dual personnality (Base T or Base X), 1 console port, 1 power supply)

### DIFF
--- a/device-types/Aruba/ArubaMC-3400.yaml
+++ b/device-types/Aruba/ArubaMC-3400.yaml
@@ -4,22 +4,18 @@ slug: aruba3400
 is_full_depth: no
 u_height: 1
 interfaces:
-  - name: (BASE-T)gigabitethernet 0/0/0
-    type: 1000base-t
   - name: (BASE-X)gigabitethernet 0/0/0
     type: 1000base-x-sfp
-  - name: (BASE-T)gigabitethernet 0/0/1
-    type: 1000base-t
+    #comment: Combo port either 1000base-x-sfp or 1000base-t
   - name: (BASE-X)gigabitethernet 0/0/1
     type: 1000base-x-sfp
-  - name: (BASE-T)gigabitethernet 0/0/2
-    type: 1000base-t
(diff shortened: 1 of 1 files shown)


device-types/Aruba/ArubaMC-3400.yaml
```diff
@@ -0,0 +1,29 @@
...
+  - name: (BASE-X)gigabitethernet 0/0/1
+    type: 1000base-x-sfp
+  - name: (BASE-T)gigabitethernet 0/0/2
+    type: 1000base-t
+  - name: (BASE-X)gigabitethernet 0/0/2
+    type: 1000base-x-sfp
+  - name: (BASE-T)gigabitethernet 0/0/3
+    type: 1000base-t
+  - name: (BASE-X)gigabitethernet 0/0/3
+    type: 1000base-x-sfp
```
ryanmerolle, line 22: interface names should match exactly what you would see in the configuration/gui/label on the device.  BASE-T is a description of the port type.

alagoutte: Hi ryan, what the best for define combo interface ?

ryanmerolle: You mean interfaces that can be either onboard rj45/1000BASE-T vs 1000base-x-sfp?  We just call them 1000base-x-sfp.  You can put a comment under the port like `   #comment: Dual function interface sfp or onboard rj45`

alagoutte: it will no be better if it is possible to set it is a shared/combo port ?

ryanmerolle: Shared combo ports that can do RJ45 or some form of SFP plugable are just modeled as the sfp version for now.  You can add a comment to each port to idnidcate it is a shared/combo, but the netbox data model does not support importing the interfaces the way you currently modeled this.

PR #218, opened 2020-09-13
Create SRX-4100.yaml

sy-na-github:
added srx-4100 

this model comes by default with 2 power supplies.


sy-na-github:
Not sure why my second commit is failing.  Any help?

PR #222, opened 2020-09-14
Add Dell PowerConnect-8024

iguy0:
Dell PowerConnect-8024 and Delta Networks ET-DT7024. They share the same hardware.

### DIFF
--- a/device-types/Dell/PowerConnect_8024.yaml
+++ b/device-types/Dell/PowerConnect_8024.yaml
@@ -1,0 +1,63 @@
+manufacturer: Dell
+model: PowerConnect 8024
+slug: powerconnect_8024
+u_height: 1
+is_full_depth: false
+console-ports:
+  - name: Serial
+    type: db-9
+power-ports:
+  - name: Power
+    type: DPSN-300DB
+interfaces:
+  - name: Te1/0/1
+    type: 10GBase-SR
+  - name: Te1/0/2
+    type: 10GBase-SR
+  - name: Te1/0/3
+    type: 10GBase-SR
+  - name: Te1/0/4
+    type: 10GBase-SR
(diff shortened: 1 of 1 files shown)


device-types/Dell/PowerConnect_8024.yaml
```diff
@@ -0,0 +1,70 @@
...
+    type: 10gbase-tx
+  - name: Te1/0/22
+    type: 10gbase-tx
+  - name: Te1/0/23
+    type: 10gbase-tx
+  - name: Te1/0/24
+    type: 10gbase-tx
+  - name: Management
+    type: 100base-tx
+    mgmt_only: true
```
ryanmerolle, line 70: The validation test is failing because we do not have a "10gbase-tx" interface type defined.  I think you meant this to be "10gbase-t".  Fix the issue and I will merge this PR.

rfdrake: There are a couple of other possible issues with this template.  I have a PowerConnect 8024 so I thought I would try this.

I believe the SFPP and 10gbase-T ports are reversed.  My own switch has 20 10gbase-T ports and 4 sfpplus.
Also, it's nitpicky but the Management interface is 1000base-tx.

It's possible there are different PC8024 models.   This is what it shows in show version for mine:

```
System Object ID.................. 1.3.6.1.4.1.674.10895.3023
System Model ID................... PC8024
Machine Type...................... Powerconnect 8024
```

ryanmerolle: @rfdrake do you want to take this PR over?

ryanmerolle:
This is stale.  Please reopen if you want to work on this.

PR #250, opened 2020-10-21
Add Device Type for the Ubiquiti ER-X

bonzi:
Adds support for the [Ubiquiti ER-X](https://www.ui.com/edgemax/edgerouter-x/)

Passes pytest on my local machine as GitHub Actions and branches don't play well with each other.

### DIFF
--- a/device-types/Ubiquiti/er-x.yaml
+++ b/device-types/Ubiquiti/er-x.yaml
@@ -6,6 +6,7 @@ u_height: 1
 is_full_depth: false
 power-ports:
 - name: 12V Power Supply
+  type: None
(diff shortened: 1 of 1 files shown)


bonzi:
The current iteration of devicetype-library lacks a Null type for supplies that don't have an applicable type under NetBox. For example, the ER-X uses a 12V barrel jack to power it, which NetBox doesn't have as a choice.

Supply Types are not mandatory as of v2.9.7 of NetBox. See below
image

bonzi:
Added Virtual `switch0` interface, now should be feature complete for most standard use cases.

PR #251, opened 2020-10-23
Added various WTI devices

wtinetworkgear:
Added the WTI device CPM-1600-1-EC, VMR-HD4D16H, NBB-20VD20-1, NPS-8H20-ATS-1, CMP-800-1-ECA


ryanmerolle:
Did you mean to add all 5 devices in this commit which only mentions CPM-1600-1-EC in the title?

wtinetworkgear:
Thanks for the catch, we added those various devices for a customer that requested and it was bit of a rush. We changed the title to reflect a more generic commit.

PR #253, opened 2020-11-08
Ubiiquiti unifi

jaco1982:
Added definitions for current Unifi generation 1 switches. These are distinct models from the Ubiquiti EdgeSwitches


jaco1982:
It appears the checks failed due to a missing newline, but I can't see any clear indication which file this was in. Can anyone shed some light on this?

PR #260, opened 2020-11-18
Aruba mobilitymaster

mtinberg:
Add Aruba controllers, Mobility Master VMs and 72xx hardware models that were not yet in the library.


nahun:
The MM VMs probably shouldn't be added as device types since they would be created as Virtual Machines.

mtinberg:
Done

nahun:
Oh I'm just a contributor like you. I'll be dumping a bunch of models soon as well for them to review. You actually added some that I was about to create myself so thanks!

PR #268, opened 2020-11-18
Add Cisco Nexus 77xx family Chassis

mtinberg:
Add Cisco Nexus 77xx Chassis models.


DanSheps:
I think there are 3 PSU slots on the 7706 @mtinberg

mtinberg:
> 
> 
> I think there are 4 PSU slots on the 7706 @mtinberg

I think you are right

mtinberg:
https://www.cisco.com/c/en/us/products/collateral/switches/nexus-7000-series-switches/data_sheet_c78-728187.html

PR #277, opened 2020-11-21
Add various Aruba and legacy HPE devices

nahun:
Add some Aruba controllers and switches. Also added a couple legacy ProCurve HP switches.
- 2920-24G
- 2920-48G-PoE+
- 2920-48G
- 2930F-48G-4SFP+
- Aruba7005
- Aruba7010
- Aruba7205
- ProCurve 2810-24G
- ProCurve 2810-48G

### DIFF
--- a/device-types/Aruba/Aruba7005.yaml
+++ b/device-types/Aruba/Aruba7005.yaml
@@ -1,0 +1,19 @@
+manufacturer: Aruba
+model: Aruba7005
+slug: aruba7005
+is_full_depth: no
+u_height: 1
+interfaces:
+  - name: gigabitethernet 0/0/0
+    type: 1000base-t
+  - name: gigabitethernet 0/0/1
+    type: 1000base-t
+  - name: gigabitethernet 0/0/2
+    type: 1000base-t
+  - name: gigabitethernet 0/0/3
+    type: 1000base-t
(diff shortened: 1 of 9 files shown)


device-types/Aruba/Aruba7005.yaml
```diff
@@ -0,0 +1,19 @@
+manufacturer: Aruba
+model: Aruba7005
+slug: aruba7005
+is_full_depth: no
```
alagoutte, line 4: missing power ?

nahun: Honestly, I couldn't figure out what to put for power since I think its an external power supply so most likely a DC barrel connector. I've never even seen a 7005, we only have one in an office thats on a different continent than me.

alagoutte: There is port 3 (or 0) can accept PoE input... 

and for external power, it is from 7005 Ordering guide... _AP-AC-12V30A 12V/30W AC/DC Desktop Style 1.7/4.0/9.5mm Circular 90 Deg Plug DoE Level VI Adapter_

nahun: You'll want to create another pull request to add the power info for the 7005.

PR #290, opened 2020-12-06
Adding ISR897VA.yaml

fraserc182:
Adding in cisco ISR897VA router.


fraserc182:
Any hints on why this has failed validation would be good.
Can't see any issues with the yaml myself.

mayafox:
From the top level of your local run pytest and it will provide a detailed failure log.

fraserc182:
@mayafox thanks for the help, got it sorted now.

PR #342, opened 2021-02-22
Nokia 7210 sas

dejantep:
add support for Nokia 7210 SAS series


dejantep:
Why is this failing?

ryanmerolle:
I think it is the half u height.

PR #348, opened 2021-02-27
update: README and move cisco c3750x to correct dir

ndom91:
Updated README to add the best devicetype importer I've found yet [minitriga/Netbox-Device-Type-Library-Import](https://github.com/minitriga/Netbox-Device-Type-Library-Import). 

Also moved that lone cisco device type template to the correct subdirectory. Not sure how it landed in the root of the repo..


ryanmerolle:
Hello @ndom91 I actually had a big PR to fix duplicate keys, update the README, add some better quality checks, and etc.  It seems like your PR is now conflicting.  If you address this I will merge.

ndom91:
Thanks for the quick response! Looks like you already moved that random cisco device template file that landed in the root of the repo haha

ryanmerolle:
I should have said this earlier, but there are a lot of users who are not quite as savvy linux/python etc.  You may just want to add a blurb that this repo does not support said script, and all instructions can be found at said repo.

Thinking out loud this might also be worth adding to the wiki when if/when we build that out.

ndom91:
@ryanmerolle yeah good call, I also have a PR open on that repo (I've gotten to know the maintainer a bit actually after having dug into it a bit haha) to clean up its readme, make it easier to get into, etc. See: https://github.com/ndom91/Netbox-Device-Type-Library-Import/tree/update-readme-env

PR #354, opened 2021-03-11
Adding Arista model 7020SR-32C2

Jake3292:


Jake3292:
Adding an Arista 7020SR-32C2

PR #376, opened 2021-04-06
Add Tripplite SMART1500LCD UPS

bobjacobsen:
Add Tripplite SMART1500LCD UPS with info from [manufacturer info page](https://www.tripplite.com/smartpro-lcd-120v-1500va-900w-line-interactive-ups-avr-2u-rack-tower-lcd-usb-db9-serial-8-outlets~SMART1500LCD) and [datasheet](https://assets.tripplite.com/product-pdfs/en/smart1500lcd.pdf)

### DIFF
--- a/device-types/TrippLite/SMART1500LCD.yaml
+++ b/device-types/TrippLite/SMART1500LCD.yaml
@@ -7,7 +7,7 @@ comments: |
 
   Dimensions: 438 x 89 x 267 mm (17.25" x 3.5" x 10.5") Weight: 13.6 kg
 u_height: 2
-is_full_depth: true
+is_full_depth: false
 console-ports:
   - name: RS232
     type: de-9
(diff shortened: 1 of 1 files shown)


AevumDecessus:
Looks good to me, only point of difference is that this is NOT a full depth device, it's only 10.5" deep, and only attaches to one side of a rack

bobjacobsen:
Thanks!  Will update. Wasn't sure exactly what full-depth meant.  Is it about the back attachment?

AevumDecessus:
Full Depth means you can't attach something to the same elevation height on the other side of the rack, usually due to the device going all the way from front to back, or rails for things like a server that connect to both the front and back posts of a 4 post rack. The other reason to mark something full depth is if you need unrestricted airflow and can't put something on the other side of the rack due to that

PR #377, opened 2021-04-08
Create huawei-s5720-56c-ei-48s-ac.yaml

krasotinpa:


m2martin:
@ryanmerolle I've seen that the merged file is not located in the corresponding vendor folder and does not follow the verndor's file name convention.
I suggest moving the file into the vendor folder and renaming the file based on the existing device files. (Houskeeping)

PR #378, opened 2021-04-08
Added devices

KaleBrink:
Added 4 devices. 2 from Solid Optics and 2 from Dell. These configurations reflect what we have in de data center.


ryanmerolle:
Those interface names in parentheses, are they the physical labels, the configuration names, or something else? We generally want to use the config displayed name for the interface name, and labels for if the port physically is labeled differently.

KaleBrink:
The interface names in parentheses are the slot names where the interface card is placed in the blade chassis. The interface names outside the parentheses are the names you see in the cli.
Attached is a foto of the actual configuration we have in our datacenter. I hope this clarifies things for you. If this is not how you want it, please let me know how to add a configuration like this.

ryanmerolle:
> The interface names in parentheses are the slot names where the interface card is placed in the blade chassis. The interface names outside the parentheses are the names you see in the cli.
> Attached is a foto of the actual configuration we have in our datacenter. I hope this clarifies things for you. If this is not how you want it, please let me know how to add a configuration like this.
> 

It is not about how I want it, just trying to keep things consistent.  I think you could put that in the label given the tool tip of the label in the GUI states label is for the physical label.  We try to keep interface name the logical name seen in the config.

As an additional FYI, any card or module that does not come with a device/chassis is left out of these models.

PR #381, opened 2021-04-13
Add Cisco devices C1000-8P-E-2G-L, C1000-24P-4G-L and bunch of Cisco SMB devices

qaxi:


ryanmerolle:
I see some devices with port channels defined.  Is that how the devices are configured out of the box?

qaxi:
> I see some devices with port channels defined. Is that how the devices are configured out of the box?
Yes, exactly.

When you list interfaces _(show interfaces status)_ there are Port-channel interfaces present. 

CiscoSMB firmware is similar but in some things really different from cisco IOS (napalm (https://github.com/napalm-automation-community/napalm-s350) and Ansible (https://github.com/ansible-collections/community.ciscosmb) has separate driver for it ...

qaxi:
May be I misundestund term "allocated_draw" ...
On PoE switch maximum_draw is switch consumption without PoE + all PoE it can provide, right?
Is "allocated_draw" equivalent of PoE switch consumption itself?

ryanmerolle:
If I recall correctly, allocated is just the base level you can assume at minimal run.  Max is the max a psu can pull.

qaxi:
> If I recall correctly, allocated is just the base level you can assume at minimal run. Max is the max a psu can pull.

Thank you.
Aalues in this pull request are correct, because documentation provided by Cisco say:

- regular switches do not usually show minimum consumption, just maximum
- PoE switches usually show consumption with any PoE delivered (allocated) and with maximum PoE delivered (maximum)

PR #384, opened 2021-05-03
Ubiquiti er es usg

redkite127:
Updated style guidance because we cannot respect it if we want to comply with yamlfmt enforced by the pre-commit script.

Updated/added Ubiquiti devices:

- ER-X
- ER-X-SFP
- USG
- ES-10X
- ES-10XP
- ES-12F


redkite127 (commented):
(and also fix previous commit...)

PR #392, opened 2021-05-17
Create ATN910I-D.yaml

martinum4:
My first device type, hope this is ok.


ryanmerolle:
Your ATN910I-D.yaml is failing because allocated_power is set to null.

Per the [schema.json rules](https://github.com/netbox-community/devicetype-library/blob/15c65de97a0c8a25d874641cb937ee63a04c8e5d/tests/schema.json#L238) if it is defined, it should be an integer.

PR #397, opened 2021-05-23
MikroTik CRS125-24G-1S-RM (Rackmount-Version)

julianze:
Added new definition of CRS125-24G-1S, rackmount-Version


ryanmerolle:
This is still in draft mode. Set it to ready for review, and I will merge it.

PR #416, opened 2021-06-19
Added Stratix 5700 Variants

BrandonRedder:
Added all variants of Rockwell Automation's Stratix 5700 industrial managed switch.


BrandonRedder:
Missed the uniqueness requirement on model number; need to correct and resubmit.

PR #425, opened 2021-07-06
Create cisco-ncs540-acc-sys.yaml

krasotinpa:
Cisco NCS540 device type


ryanmerolle:
I would the empty description and label fields

PR #438, opened 2021-07-31
Add Juniper SRX650

parkervcp:
This would add the Juniper SRX650.
This is the default config with no extras such as the 2x10G and 16x1G switch XPIMs


dmcken:
Hi, anything happening with this pull request?

dmcken:
Already added via PR #632

PR #457, opened 2021-09-01
Add RA Stratix 5400 Switches

BrandonRedder:
This PR adds Rockwell Automation's Stratix 5400 industrial switches (Cisco IE "private label").


BrandonRedder:
Correcting device bays.

PR #467, opened 2021-09-07
Add Panduit Copper / Fiber Patch Panels

BrandonRedder:
This PR adds a collection of Panduit's copper (Mini-Com) and fiber (Opticom) patch panels and related components.


BrandonRedder:
Need to correct RU's on child part numbers. Will resubmit after fix.

PR #505, opened 2021-10-13
Netgear m4250

nasi87:
It's my second try to contribute **Netgear M4250** devices. 
Hopefully it's correct now.


nasi87:
I ran "pre-commit run --all" which apparently fixed empty line errors and indentation.
I don't understand why exported files from my Netbox instance are not accepted anyways but that's another discussion I guess.

nasi87:
Why did it fail? What does exit code 1 mean?

FloEisen:
Just a hint from my side:
- subdevice_role can either be "parent" or "child" but not empty
- allocated_draw can be removed if there is no value, otherwise it throws errors with pytest because of the missing type (integer expected)

You also can check your syntax by running pytest locally.

nasi87:
Thanks for your help.

But now I have to ask for some background information: I've build and exported these models with Netbox WebGUI of a running Netbox 3.0 installation. So they are kind of working for me. But it's not possible to import them somewhere else?
I already splitted the single exported file to different files for each model and ran "pre-commit" - but now I have to run another script (pytest) and make changes manually to every file? Where do I find said pytest? Where is a documentation about all properties and their permitted values? Why isn't Netbox checking these things before exporting?

Sorry for asking questions which may be obvious to other people but I'm no developer at all and it's my first pull request on GitHub.

PR #508, opened 2021-10-19
Create S6730-H24X6C.yaml

walday:
S6730-H24X6C (24*10GE SFP+ ports, 6*100GE QSFP28)


walday:
> Interface type needs to be 100gbase-x-qsfp28 and you have 100gbase-x-qsfpp28.

thanks, this was a very unexpected error

PR #537, opened 2021-11-04
add legacy ubiquiti device templates

bile0026:
Add legacy ubiquiti gear templates. Older style bullet, rocket, and nanostation devices.


ryanmerolle:
any reason why you updated other non-ubiquiti device templates?

bile0026:
> any reason why you updated other non-ubiquiti device templates?

No, that was unintentional. I will work to revert those.

PR #545, opened 2021-11-08
Update C9200-48T.yaml

walday:
subdevice_role: parent


michaelklose:
Does it really make sense to add subdevice_role to this device? Based on the documentation (https://netbox.readthedocs.io/en/stable/core-functionality/device-types/) device-bays should not be used for switch uplink modules.

walday:
ok! delete bays for 4 device

PR #555, opened 2021-11-08
Update C9200-48P.yaml

walday:
subdevice_role: parent


ryanmerolle:
@walday I'm somewhat removed from Cisco these days, but how is this a subdevice_role of parent?  Seems like a 1u fixed form chassis.  I thought parent devices house child devices not modules.

PR #565, opened 2021-11-18
Adding Huawei OptiXtrans DC908

suom1:


suom1:
@ryanmerolle Do you have any clue why pytest fails on `iec-60320-c19` but `iec-60320-c14` works fine? 
`iec-60320-c19` is used in many other models in this repo and is included in `tests/schema.json` under `power-outlet`.

Is it me doing something wrong?

FloEisen:
As far as I know iec-60320-c19 is a power outlet but not a valid power port (inlet).
It is like the Plug on a cable which gets plugged into the device.
So the corresponding device power port (inlet) would be the iec-60320-c20

suom1:
> As far as I know iec-60320-c19 is a power outlet but not a valid power port (inlet). It is like the Plug on a cable which gets plugged into the device. So the corresponding device power port (inlet) would be the iec-60320-c20

Ohh.. You're right. Thanks for pointing out @FloEisen

PR #571, opened 2021-11-22
Rename Rohde und Schwarz folder

FloEisen:
Renamed the "Rohde und Schwarz" vendor folder in order to avoid errors due to whitespaces


ryanmerolle:
What white space errors?  If there are some, why not surround the value in quotes?

FloEisen:
For example with the Python import script, where whitespaces are used as delimiter between vendors within the "vendors" env-variable. Quotes didn't work for me here.

I know, it would be cleaner to change the script to use other delimiter.
But another reason is to get rid of the "und" as it is a German word and not the best for the international name of Rohde & Schwarz.

ryanmerolle:
The only issue is that the device will now show up in netbox associated to a manufacturer name that will look odd with no space when there should be.  How about we quote it with the & and spaces?

FloEisen:
Yes, good point! I tested it successfully with the &. Commited the changes
Thanks!

PR #578, opened 2021-11-23
Dell N3248TE-ON network switch

sc68cal:
This is based off the spec sheet at

https://www.delltechnologies.com/asset/en-us/products/networking/technical-support/dell-networking-n3200-powerswitch-specsheet.pdf

### DIFF
--- a/device-types/Dell/Networking_N3248TE-ON.yaml
+++ b/device-types/Dell/Networking_N3248TE-ON.yaml
@@ -1,0 +1,184 @@
+---
+manufacturer: Dell
+model: N3248TE-ON
+slug: n3248te-on
+u_height: 1
+is_full_depth: true
+console-ports:
+  - name: Console
+    type: rj-45
+power-ports:
+  - name: PS1
+    type: iec-60320-c14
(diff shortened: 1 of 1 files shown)


device-types/Dell/Networking_N3248TE-ON.yaml
```diff
@@ -0,0 +1,183 @@
+manufacturer: Dell
+model: N3248TE-ON
+slug: n3248te-on
+u_height: 1
+is_full_depth: true
+console-ports:
+  - name: Console
+    type: rj-45
+power-ports:
```
sc68cal, line 9: I'm a little unsure about these values, since the following are listed in the spec sheet:

| Power cords |
----|
C15 to NEMA 5-15, 1.8M (N3208PX-ON only)
C13 to NEMA 5-15, 3M (all other N3200 platforms)
C13 to C14, 2M (all other N3200 platforms)

Also I'm not sure about the `maximum_draw` and `allocated draw` since the spec sheet lists 550W power supplies, but then a maximum power consumption in watts of 212

PR #583, opened 2021-12-01
Main

KaleBrink:
Added new devices
Raritan DSX2-16
Fortinet FG-80F


danner26:
Do you still need this? @KaleBrink

PR #596, opened 2021-12-17
Create RPS-C3K-PWR-750WAC

manolab:
Initial rendering of the RPS. It needs to added with DC ports to switch...


danner26:
Do you still need this? @manolab

manolab:
No, thank you!

PR #599, opened 2021-12-17
Create CCR2004-1G-12S+

manolab:


dmcken:
Just looking at it, there is only one CCR2004-1G-12S+ which is CCR2004-1G-12S+2XS and its already added.

PR #609, opened 2021-12-23
Adding Mellanox SB7800

vpalacio:
new device type Mellanox SB7800


michaelklose:
As this is a Infiniband switch I think infiniband-edr is the correct port type.

PR #621, opened 2022-01-06
Fix #616 Unknown properties are ignored during schema validation

### LINKED ISSUE #616: Unknown properties are ignored during schema validation
I have done some experiments with the new module-bays option and noticed, that unknown/misspelled properties are currently ignored during schema validation and doesn't raise a failure.

This can be fixed by adding "additionalProperties": false, to the schema.json file.

Currently I have noticed the following unknown properties in use:
comments
comment
power-port
console-server-port
console_ports
power_ports
full_depth

For reference I added my pytest results with modified schema.json.
[results.txt](https://github.com/netbox-community/devicetype-library/files/7797637/results.txt)

I can open a pull requests to fix this if the issue is accepted.

michaelklose:
- Added "additionalProperties": false to schema.json
- Added comments property to schema.json
- Fixed all misspelling properties in files

### DIFF
--- a/device-types/Aruba/MC-3200.yaml
+++ b/device-types/Aruba/MC-3200.yaml
@@ -7,16 +7,16 @@ u_height: 1
 interfaces:
   - name: (BASE-X)gigabitethernet 0/0/0
     type: 1000base-x-sfp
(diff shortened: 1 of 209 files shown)


ryanmerolle:
Nice PR!  Thanks!

PR #622, opened 2022-01-11
Update Ubiquiti USW-Aggregation and UDM-Pro

aphistic:
I was adding these two devices to a new NetBox install today and noticed a few discrepancies in the definitions.

**USW-Aggregation**
I was unable to find a 10-port version of this switch and the datasheet I found from 2020 only lists an 8-port variation, which is the only one I could find. I updated the device to only include 8 ports now.

**UDM-Pro**
The primary change here is to mark it as not a full-depth device. As long as I was updating it I thought I'd add the airflow and update the model name to match the data sheet's name and add the part number.

**Airflow Direction in Schema**
I added the "airflow" field to the device type schema. I added airflow direction to the UDM-Pro on a whim, so this was required to allow the tests to pass.

I may have more updates coming and I know I'll have a definition for the USW-Pro-Aggregation to add soon as I have one of those coming!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *.pyc
 .idea
 .DS_Store
+venv/
--- a/device-types/Ubiquiti/UDM-Pro.yaml
+++ b/device-types/Ubiquiti/UDM-Pro.yaml
@@ -1,7 +1,9 @@
 ---
 manufacturer: Ubiquiti
-model: UDM-PRO
+model: Dream Machine Pro
 slug: udm-pro
+airflow: front-to-rear
+part_number: UDM-Pro
 comments: |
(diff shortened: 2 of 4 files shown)


ryanmerolle:
When ready take this out of draft mode

aphistic:
Thanks for the ping! I was expecting a notification once the checks finished/passed but I guess I didn't get one. :)

aphistic:
The biggest thing I have a question on here is I changed the UDM-Pro to not full-depth, but considering airflow and airflow direction maybe that shouldn't be the case?

ryanmerolle:
I get your logic, but if the device is not full depth, then it should not be modeled as such.

ryanmerolle:
Thanks for the PR and updating the schema to match the core project!

PR #627, opened 2022-01-22
Add ubnt af 5xhd

dmcken:
Adding the Ubiquiti airFiber 5XHD

https://dl.ubnt.com/datasheets/airfiber/airFiber_5XHD_DS.pdf

Used ieee802.11ax as it doesn't allow wireless parameters to be specified despite the fact that it is a LTE-U device.

Has been tested to import fine into my v3.1.6 instance of netbox.


dmcken:
Also added "bridge" as an acceptable device interface type.

PR #630, opened 2022-01-25
Create D-link_DGS-1210-52.yaml

walday:


ryanmerolle:
Label signifies the label on the physical device.

PR #632, opened 2022-01-26
Add Juniper SRX650

dmcken:
Import into current stable successful and tests passed.

### DIFF
--- a/device-types/Juniper/SRX650.yaml
+++ b/device-types/Juniper/SRX650.yaml
@@ -1,0 +1,68 @@
+---
+manufacturer: Juniper
+model: SRX650
+slug: srx650
+u_height: 2
+is_full_depth: false
+comments: |
+  [Juniper SRX650 Data Sheet](https://www.juniper.net/content/dam/www/assets/datasheets/us/en/security/srx-series-services-gateways-branch-datasheet.pdf)
+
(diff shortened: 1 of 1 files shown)


ryanmerolle:
I’d recommend deleting keys that have empty quotes set (‘’)

dmcken:
Removed.

ryanmerolle:
I missed this before, but labels signify an actual physical label on the device.  Comments or descriptions may be better to describe an interface

dmcken:
I can change the labels to just the numbers for the PIM slots:
https://www.juniper.net/assets/img/products/image-library/srx-series/srx650/srx650-front-high.jpg - PIM1 is just 1 as the physical label.

There is no label for the power supply slots or the SRE slots.

Looking at:
https://github.com/dmcken/devicetype-library/blob/master/README.md - Do module bays have comments or description fields?

ryanmerolle:
I even just use `# Comment: ` to show these sort of notes where a comment key is not supplied.

Given PIM1 is technically a label to your point you could use that.  I have seen people use labels as a comment field, and I wanted to make sure that was not your intent here.

ryanmerolle:
let me know when you a ready to merge

dmcken:
Its fine to merge now.

PR #637, opened 2022-01-27
Add missing types to schema.json

michaelklose:
Add missing types to schema.json and modified order to match https://github.com/netbox-community/netbox/blob/develop/netbox/dcim/choices.py


ryanmerolle:
Thanks!  We are usually a step behind what is merged into netbox core.

PR #646, opened 2022-02-01
Add F5 i2000 Series

DominicHar:
Add i2600 and i2800 devices


DominicHar:
Fixed Line endings and resubmitt

Issue #604, opened 2021-12-18
Expand the repo to include module types (coming in NetBox v3.2)


jeremystretch:
NetBox v3.2 will introduce [support for device modules](https://github.com/netbox-community/netbox/issues/7844) (e.g. line cards). I'd like to add a second root directory to this repo named `module-types`, which will mirror what exists today for device types.

The YAML definitions for module types are still TBD, and won't be finalized until NetBox v3.2.0 is released, but it should closely replicate the format used for device types. We'll likely need to refactor the tests though.

jeremystretch:
NetBox v3.2 also introduces support for [inventory item templates](https://github.com/netbox-community/netbox/issues/8118), which we should add here as well.

PR #659, opened 2022-02-11
Module types

jeremystretch:
### Closes: #604

- Move the devicetype schema to `schemas/devicetype.json`
- Add `schemas/moduletype.json`
- Move the component scehmas to a common file at `schema/components.json`
- Refactor `definitions_test.py` to support testing both device and module types (and to support the new external component references)
- Add the first module type definitons


jeremystretch:
We still need to clean up the test for duplicates, but that can be handled separately.

PR #675, opened 2022-02-20
device type: Juniper NFX250

FlorianHeigl:
The NFX serves as a 'network virtual function' platform.
It's basically a VM host plus autoprovisioning and a lot of virtual networking options to allow chaining multiple network VMs behind each other.

### DIFF
--- a/device-types/Juniper/NFX250.yaml
+++ b/device-types/Juniper/NFX250.yaml
@@ -9,19 +9,13 @@ comments: 'Juniper NFV Platform for CPE install'
 console-ports:
   - name: Console
     type: rj-45
-    label: ''
-    description: ''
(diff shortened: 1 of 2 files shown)


dmcken:
Here is the error when running the tests:

device-types/Juniper/NFX250.yaml failed validation: None is not of type 'integer'

Failed validating 'type' in schema['properties']['power-ports']['items']['properties']['allocated_draw']:
    {'type': 'integer'}

On instance['power-ports'][0]['allocated_draw']:
    None

dmcken:
You need to set it to a number or just exclude that field if you don't know.

FlorianHeigl:
Thank you!
I'll remove it.

Issue #575, opened 2021-11-23
The name of this barrel plug


FacundoAcevedo:
Hi, I have the feeling that this is not the correct place for this question, but since I've been creating some devices an answer will improve the quality of my contributions.
I'm looking for the name in Netbox of this kind of power connector:
![](https://mla-s1-p.mlstatic.com/608409-MLA31787200012_082019-F.jpg)

Thanks

FloEisen:
For this kind of plugs I used `dc-terminal` as type

FacundoAcevedo:
Thank you very much

PR #679, opened 2022-02-21
Add Cisco Meraki MS120-8 switches

Nothing4You:


Nothing4You:
MS120-8/MS120-8LP have an external psu, MS120-8FP has an internal PSU.

Based on #575 I guess `dc-terminal` is the way to go for the external ones.

As per https://documentation.meraki.com/MS/MS_Installation_Guides/MS120-8_Series_Installation_Guide#MS120-8FP_Series_back_panel this should be `iec-60320-c14` on the FP one.

PR #687, opened 2022-02-23
Extend tests to check for null/empty values

### LINKED ISSUE #686: Extend tests to check for null/empty values
We should extend the validation tests to flag any null or empty attribute definitions. For example, `description: ''` is unnecessary, since the attribute will default to an empty string. Explicitly specifying an empty value merely consumes space unnecessarily.

michaelklose:
Fixes #686
- yamllint: enable empty-values
- definitions_test.py: add test for empty quotes
- clean up device-type files


jeremystretch:
Excellent, thanks for taking this!

PR #697, opened 2022-03-04
Add CheckPoint SG3600 and SG6400

madecto:
Manufacturer: Check Point
Model: SG3600

Manufacturer: Check Point
Model: SG6400

Manufacturer: Check Point
Model: 4-1F-C

Manufacturer: Check Point
Model: 4-10F-C

Manufacturer: Check Point
Model: 8-1C-C

Manufacturer: Check Point
Model: 1500/3600/3800-RM-DUAL


jeremystretch:
Hi @madecto, thanks for this! It looks like CI is failing on some type validation. Would you mind changing the labels to strings please? (For example, `label: 1` should become `label: '1'`) Thanks!

madecto (commented):
Thanks for the hint @jeremystretch. This should do the trick.

PR #595, opened 2021-12-17
Create ASR-920-24SZ

manolab:

PR #700, opened 2022-03-08
Add asr 920 24sz

dmcken:
Cleaned up https://github.com/netbox-community/devicetype-library/pull/595


ryanmerolle:
I think you should delete the `subdevice_role: parent`

If I recall parents would have discrete devices in them with their like blade chassis.

PR #718, opened 2022-03-24
Add juniper QFX3500-48S4Q

dmcken:


jeremystretch:
Validated this locally to work around bug with linter configuration. Good to go, thanks!

PR #719, opened 2022-03-24
Update stale.yml

dmcken:
Disabled check for truthyness

Taken from:
https://github.com/adrienverge/yamllint/issues/430


jeremystretch:
Hey @dmcken thanks for looking into this! I just happened to push a fix as well (ended up just linting the definitions directories). Should be working again now.

PR #735, opened 2022-04-06
Create PowerEdge_R750.yaml

bbrk364:
adding the Dell R750

### DIFF
--- a/device-types/Dell/PowerEdge_R750.yaml
+++ b/device-types/Dell/PowerEdge_R750.yaml
@@ -1,0 +1,56 @@
+---
+manufacturer: Dell
+model: PowerEdge R750
+slug: dell_poweredge_r750
+comments: 'Dell EMC PowerEdge R750 Technical Specifications https://dl.dell.com/content/manual45025655-dell-emc-poweredge-r750-technical-specifications.pdf?language=en-us&ps=true' 
+u_height: 2
+is_full_depth: true
+subdevice_role: parent
+console-ports:
+  - name: Optional Rear Serial
+    type: de-9
+  - name: Front VGA
+    type: other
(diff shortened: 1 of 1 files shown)


device-types/Dell/PowerEdge_R750.yaml
```diff
@@ -0,0 +1,56 @@
...
+    type: usb-a
+console-server-ports:
+  - name: Front USB1
+    type: usb-a
+  - name: Front Dedicated iDRAC direct USB
+    type: usb-a
+  - name: Rear USB1
+    type: usb-a
+  - name: Rear USB2
+    type: usb-a
```
DanSheps, line 26: We are no longer modelling USB ports unless they function solely as a console port.

The iDRAC port should be modelled as a interface as that is how it functions

device-types/Dell/PowerEdge_R750.yaml
```diff
@@ -0,0 +1,56 @@
...
+  - name: Optional OCP NIC
+  - name: Optional BOSS-2 Removed
+  - name: Optional PCIe-Gen4 1
+  - name: Optional PCIe-Gen4 2
+  - name: Optional PCIe-Gen4 3
+  - name: Optional PCIe-Gen4 4
+  - name: Optional PCIe-Gen4 5
+  - name: Optional PCIe-Gen4 6
+  - name: Optional PCIe-Gen4 7
+  - name: Optional PCIe-Gen4 8
```
DanSheps, line 56: Should be modelled as a module bay

DanSheps (changes requested):

danner26:
Do you still need this? @bbrk364

PR #736, opened 2022-04-06
Create UCSB-5108-AC2-UCS 6324.yaml

bbrk364:
for Cisco UCS 5108 Blade Server Chassis w/ Cisco UCS 6324 Fabric Interconnect for Mini Express FlexPod

### DIFF
--- a/device-types/Cisco/UCSB-5108-AC2-UCS 6324.yaml
+++ b/device-types/Cisco/UCSB-5108-AC2-UCS 6324.yaml
@@ -1,0 +1,98 @@
+---
+manufacturer: Cisco
+model: UCSB-5108-AC2-6324
+slug: ucsb-5108-ac2-6324
+part_number: UCSB-5108-AC2-6324
+comments: 'Cisco UCS 5108 Blade Server Chassis w/ Cisco UCS 6324 Fabric Interconnect for Mini Express FlexPod'
+u_height: 6
+is_full_depth: true
+subdevice_role: parent
+airflow: front-to-rear
+console-ports:
+  - name: IOM A Con0
+    type: rj-45
+  - name: IOM A usb0
+    type: usb-a
+  - name: IOM B Con1
+    type: rj-45  
+  - name: IOM B usb1
+    type: usb-a
+power-ports:
+  - name: psu-1
+    type: iec-60320-c20
+    maximum_draw: 2500
+  - name: psu-2
+    type: iec-60320-c20
+    maximum_draw: 2500
+  - name: psu-3
+    type: iec-60320-c20
+    maximum_draw: 2500
+  - name: psu-4
+    type: iec-60320-c20
+    maximum_draw: 2500
+interfaces:
+  - name: IOM A mgmt0
+    type: 1000base-t
+    mgmt_only: true
+  - name: IOM A eth1/1
+    type: 10gbase-x-sfpp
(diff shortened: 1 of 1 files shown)


ekrichbaum:
I like that you modelled the IOM slots as device bays (probably should be module bays now) and the IOMs should be modelled as modules, removing those from the 5108 chassis already here.  The mini fabs should be a device/module similar to a 2208 or 2204, etc. at that point. This would allow for the ports to be added outside of the raw chassis, which is more elegant. The item up there is with the 2208 installed so could be removed and redone.

bbrk364:
The UCS 5108 w/ two Fabric-Extender-1/UCS 6324 because of the version for Netbox =< 3.18

ekrichbaum:
You have 2 extra spaces after rj-45 on line 17 which is stopping the lint from passing.  That model still isn't quite right even with older netbox, but functional for you I would guess. Because the IOM is modular, I'm against having a device type with the components already installed for my purposes but I can see that being useful in your local case. Not sure that it belongs in the library like that because of the corner case.

ryanmerolle:
I agree with @ekrichbaum. This should not get merged as is.

device-types/Cisco/UCSB-5108-AC2-UCS 6324.yaml
```diff
@@ -0,0 +1,98 @@
...
+    maximum_draw: 2500
+  - name: psu-2
+    type: iec-60320-c20
+    maximum_draw: 2500
+  - name: psu-3
+    type: iec-60320-c20
+    maximum_draw: 2500
+  - name: psu-4
+    type: iec-60320-c20
+    maximum_draw: 2500
```
DanSheps, line 32: Should add module bays for this and make these modular

device-types/Cisco/UCSB-5108-AC2-UCS 6324.yaml
```diff
@@ -0,0 +1,98 @@
...
+airflow: front-to-rear
+console-ports:
+  - name: IOM A Con0
+    type: rj-45
+  - name: IOM A usb0
+    type: usb-a
+  - name: IOM B Con1
+    type: rj-45  
+  - name: IOM B usb1
+    type: usb-a
```
DanSheps, line 19: This should be part of the FI child, not the chassis

device-types/Cisco/UCSB-5108-AC2-UCS 6324.yaml
```diff
@@ -0,0 +1,98 @@
...
+    mgmt_only: false
+  - name: IOM B Br-eth5/2
+    type: 10gbase-x-sfpp
+    mgmt_only: false
+  - name: IOM B Br-eth5/3
+    type: 10gbase-x-sfpp
+    mgmt_only: false
+  - name: IOM B Br-eth5/4
+    type: 10gbase-x-sfpp
+    mgmt_only: false
```
DanSheps, line 87: This should be part of the FI child.

DanSheps (changes requested):
I think we are going to need to reject this.  This is just a combined SKU and not an actual individual device.  You should break out the FI and chassis if you want this merged.

danner26:
Do you still need this? @bbrk364 I see there is still a lot of changes that need to be made. If you are willing to work on it please reopen the PR and we will review it once ready

PR #742, opened 2022-04-08
Changed modular Allied Telesis PSU devices

m2martin:
Changed Allied Telesis devices with modular PSUs to use device bays and added PSU products.


ryanmerolle:
@jeremystretch this is an interesting approach with module model used by PSU.

My first reaction is that this is not the intended design, but please confirm.

device-types/Allied Telesis/x950-28XSQ.yaml
```diff
@@ -104,5 +97,16 @@ interfaces:
...
+    position: 'B'
   - name: Slot 1
     position: '1'
+inventory-items:
+  - name: Fan-Module A
+    manufacturer: Allied Telesis
+    part_id: AT-SBxFAN05
+  - name: Fan-Module B
+    manufacturer: Allied Telesis
+    part_id: AT-SBxFAN05
```
DanSheps, line 112: Interestingly enough, these should probably be modelled as a module, however you cannot as we don't support inventory items on the ModuleType model.  Perhaps this should be a FR.

DanSheps (approved):

device-types/Allied Telesis/x950-28XSQ.yaml
m2martin: Didn't model them as modules because in my opinion modules are active parts inheriting some kind of entities. Fans are just exchangeable parts wich do not inherit any interfaces or settings - they are there as standard but may be exchanged by another serial number...

Didn't I get something wrong in the past discussions?

DanSheps: Nope, everything is good for now, it was just a comment.

I think we can merge this in.  I will likely do it on Monday unless someone gets to it first.

DanSheps (approved):

PR #751, opened 2022-04-10
Update UCSB-B200-M5.yaml

bbrk364:
updating w/ the device-bays

### DIFF
--- a/device-types/Cisco/UCSB-B200-M5.yaml
+++ b/device-types/Cisco/UCSB-B200-M5.yaml
@@ -1,8 +1,16 @@
 ---
 manufacturer: Cisco
-model: UCSB-B200-M5
-slug: ucsb-b200-m5
+model: UCS B200-M5
+slug: b200-m5
 part_number: UCSB-B200-M5
 u_height: 0
 is_full_depth: true
 subdevice_role: child
+  - name: Console connector
+    type: other
+device-bays:
+  - name: Drive bay 1  
+  - name: Drive bay 2
+  - name: mLOM
+  - name: Front mezzanine
+  - name: Rear mezzanine


ekrichbaum:
I can always override it again but, can I ask why you changed the model name and the slug?

image

For me, this affects automation because it is what UCS Manager uses for that model.

device-types/Cisco/UCSB-B200-M5.yaml
```diff
@@ -1,8 +1,16 @@
...
 is_full_depth: true
 subdevice_role: child
+  - name: Console connector
+    type: other
+device-bays:
+  - name: Drive bay 1  
+  - name: Drive bay 2
+  - name: mLOM
+  - name: Front mezzanine
+  - name: Rear mezzanine
```
DanSheps, line 16: Should likely be converted to modules

DanSheps (changes requested):
Please check those device bays and perhaps change to module bays

device-types/Cisco/UCSB-B200-M5.yaml
ekrichbaum: I wouldn't think so since they aren't going to have interfaces etc. that change.  The M5 (M6, M4, M3) etc. do have those types of things.  Do you model all of your harddrives to warrant making that a device bay?  The mezzanines could be changed (different RAID card, different VIC etc.).  We don't model down to that layer and just note it as an inventory item.  They'll be something there for sure but is it necessary to track out separately?

[https://www.cisco.com/c/en/us/products/collateral/interfaces-modules/unified-computing-system-adapters/datasheet-c78-741130.html]

The VIC1440 be an example.  Yes, it takes that mezzanine position but nothing physical about ports/etc, just a change in how the internal connections to the fabrics are handled.  Yes, the IOMs with the ports make perfect sense to be modules now because it could be x 40G or x 10G.  Here, I don't see it other than device bay if you wanted to track every subpart on every device.

DanSheps: My point is, it likely shouldn't be listed in here unless it is an actual sub-device.

If you don't do it as modules, they should be modelled as inventory items, not device bays.

From docs:

> Device bays represent a space or slot within a parent device in which a child device may be installed. For example, a 2U parent chassis might house four individual blade servers.

And:

> Child devices are first-class Devices in their own right: That is, they are fully independent managed entities which don't share any control plane with the parent. Just like normal devices, child devices have their own platform (OS), role, tags, and components. LAG interfaces may not group interfaces belonging to different child devices.

Finally, the note on that page: 

> Device bays are not suitable for modeling line cards (such as those commonly found in chassis-based routers and switches), as these components depend on the control plane of the parent device to operate. Instead, these should be modeled as modules installed within module bays.

ekrichbaum: Yes, I agree. I was referring to your comment that they should likely be converted to modules. I wouldn't (and didn't) list them at all. Like I said, I don't feel the need to model those parts and just note them other ways if they deviate from our standards.  Out of several hundred blades, very few deviate for me,

DanSheps: Yep, but we want these models to conform with documentation, so all of those device bays should likely be removed.

DanSheps:
After further consideration, we are going to reject this PR out of hand.  It departs too much from the standard that is set by the other UCS blades and rack servers.

device-types/Cisco/UCSB-B200-M5.yaml
DanSheps: @ekrichbaum Thanks for the comments, after further review I think we should just reject this one out of hand.

ekrichbaum: agreed.

PR #757, opened 2022-04-10
Add F5 BIG-IP 2000s and 2200s.

YAPNie:
The both devices look the same but the second one has a hardware accelerator inside.


YAPNie:
I am not a programmer and did not understand how to make requests (from CONTRIBUTING.md). Too complicated for me. So I would better to delete this wrong stuff. Hope, no one notices it.

PR #770, opened 2022-04-13
Arista DCS-7368 got two Management interfaces

c-po:


ryanmerolle:
Technically its the supervisor module that has them.  Would you want to introduce the proper modules now that NetBox 3.2 supports modeling modules?

c-po:
@ryanmerolle sure will do.

Is there also planned support to define the line-card power consumption?

From https://www.arista.com/assets/data/pdf/Datasheets/7368X4-Datasheet.pdf


image

ryanmerolle:
That is not a feature currently requested by the user community.

ryanmerolle:
- netbox-community/netbox#9246

c-po:
Extended the chassis with position for the slot IDs.

image

This resolves the wrongly opened issue https://github.com/netbox-community/netbox/issues/9246

PR #792, opened 2022-04-30
Updated Cisco ASR1001-X

abhi1693:
added missing interfaces

### DIFF
--- a/device-types/Cisco/ASR1001-X.yaml
+++ b/device-types/Cisco/ASR1001-X.yaml
@@ -21,6 +21,18 @@ interfaces:
     type: 1000base-t
   - name: GigabitEthernet0/0/5
     type: 1000base-t
+  - name: GigabitEthernet0/1/0
+    type: 1000base-t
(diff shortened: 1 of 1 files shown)


ryanmerolle:
Is there a link to the specifications we could reference?

ekrichbaum:
I only have 2 left on the network, so that doesn't look right. That should be from a module in a module bay.
image

https://www.cisco.com/c/en/us/products/routers/asr-1001-x-router/index.html

I do seem to remember there was a model that had different GE ports but it isn't in my brain anymore. The device type already there should probably instead get a modulebay installed for the SPA and then the SPA modeled.  Or if there is a specific different version of the box then a new model created.

ryanmerolle:
That is what I was thinking,  The original contributor added the device without the module bay populated, which is correct.

DanSheps:
Yup, these interfaces are modules, not fixed interface.

abhi1693:
I'm closing since I followed the original contributor but seems like that's the wrong way to make changes.

PR #797, opened 2022-05-03
Remove questionable use of USB ports in console

DanSheps:
Currently, there are a number of USB ports being modelled in the console port/console server port section of devices where they are clearly not being used for console access (external USB storage, USB HID, USB Peripherals, etc).

This PR removes all of those as that is not the intended usage of the console port/console server port section.  Modelling for these "special" cases can be done via other methods (interfaces, inventory items, etc) by the end user.

If there is a need to model these specific use cases,  in more detail then what is allowed via those methods, then a FR should be opened against the NetBox project to propose a implementation.

### DIFF
--- a/device-types/Dell/PowerEdge_R640.yaml
+++ b/device-types/Dell/PowerEdge_R640.yaml
@@ -7,21 +7,10 @@ is_full_depth: true
 console-ports:
   - name: Serial
     type: de-9
-  - name: Optional USB Port
-    type: usb-a
   - name: Front VGA
     type: other
   - name: Rear VGA
     type: other
-console-server-ports:
-  - name: Front USB1
-    type: usb-a
-  - name: Front iDRAC Direct micro port
-    type: usb-a
-  - name: Rear USB1
-    type: usb-a
-  - name: Rear USB2
-    type: usb-a
 power-ports:
   - name: Power 1
     type: iec-60320-c14
--- a/device-types/Dell/PowerEdge_R650.yaml
+++ b/device-types/Dell/PowerEdge_R650.yaml
@@ -12,17 +12,6 @@ console-ports:
     type: other
   - name: Rear VGA
     type: other
-  - name: Internal USB1
-    type: usb-a
-console-server-ports:
-  - name: Front USB1
-    type: usb-a
-  - name: Front Dedicated iDRAC direct USB
-    type: usb-a
-  - name: Rear USB1
-    type: usb-a
-  - name: Rear USB2
-    type: usb-a
 power-ports:
   - name: Power 1
     type: iec-60320-c14
--- a/device-types/Dell/PowerEdge_R740.yaml
+++ b/device-types/Dell/PowerEdge_R740.yaml
@@ -12,19 +12,6 @@ console-ports:
     type: other
   - name: Rear VGA
     type: other
-console-server-ports:
-  - name: Front USB1
-    type: usb-a
-  - name: Front USB2
-    type: usb-a
-  - name: Front Dedicated iDRAC direct USB
-    type: usb-a
-  - name: Rear USB1
-    type: usb-a
-  - name: Rear USB2
-    type: usb-a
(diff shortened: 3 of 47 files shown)


device-types/Dell/PowerEdge_R640.yaml
```diff
@@ -7,21 +7,10 @@ is_full_depth: true
...
-  - name: Optional USB Port
-    type: usb-a
   - name: Front VGA
     type: other
   - name: Rear VGA
     type: other
-console-server-ports:
-  - name: Front USB1
-    type: usb-a
-  - name: Front iDRAC Direct micro port
```
DanSheps, line 19: Should be modelled as a interface as it provides an IP address

device-types/Dell/PowerEdge_R650.yaml
```diff
@@ -12,17 +12,6 @@ console-ports:
     type: other
   - name: Rear VGA
     type: other
-  - name: Internal USB1
-    type: usb-a
-console-server-ports:
-  - name: Front USB1
-    type: usb-a
-  - name: Front Dedicated iDRAC direct USB
```
DanSheps, line 20: Should be modelled as a interface as it provides an IP address

device-types/Dell/PowerEdge_R740.yaml
```diff
@@ -12,19 +12,6 @@ console-ports:
     type: other
   - name: Rear VGA
     type: other
-console-server-ports:
-  - name: Front USB1
-    type: usb-a
-  - name: Front USB2
-    type: usb-a
-  - name: Front Dedicated iDRAC direct USB
```
DanSheps, line 20: Should be modelled as a interface as it provides an IP address

device-types/Extreme Networks/5520-24T.yaml
```diff
@@ -11,13 +11,6 @@ console-ports:
     type: rj-45
   - name: Console_USB
     type: usb-micro-b
-console-server-ports:
-  - name: USB-Storage1
```
DanSheps, line 15: Storage is not to be modelled as a console server port

device-types/Extreme Networks/5520-24T.yaml
```diff
@@ -11,13 +11,6 @@ console-ports:
...
-    type: usb-a
-    description: 'for management or external USB flash'
-  - name: USB-Storage2
```
DanSheps, line 18: Storage is not to be modelled as a console server port

device-types/Extreme Networks/VSP-7400-48Y-8C.yaml
```diff
@@ -8,9 +8,6 @@ is_full_depth: true
 console-ports:
   - name: Console
     type: rj-45
-console-server-ports:
-  - name: USB-Storage
```
DanSheps, line 12: Storage is not to be modelled as a console server port

FlorianHeigl:
we should strive to conserve energy in these times.
a PR that has no major benefit but offloads redoing existing work to thousands of end users is not a step forward for humanity.
if it had introduced a better fitting model or carefully reviewed those affected items and migrated them, it would be a good thing. as it is: brings no benefit.

where's even the consideration for i.e. 
* a USB port that is absolutely only used to do a `configure usb` automatically on boot?
* OS that have a `line` for those USB devices?
* alcatel switches that use the USB port with bluetooth dongles to expose a wireless serial console?
* use with i.e. mikrotik Woobm-USB?
* Lenovo Servers with dual-use Craft USB Ports where you in fact run the KVM on your phone connecting over this USB port 

To claim any of that to be "questionable" when you see there was lots of people asking themselves the same question and judging that it is more sensible to go ahead and model those ports makes no sense. One should assume that others might be better informed than you any given day.


Reconsider, try provide reasoning for any single USB port that you want to remove.
If you can't , then this set of changes should be reverted. This wide-sweeping "it is questionable" is a reasonable call to action, but does not justify more than looking into the issue *in detail*.

PR #803, opened 2022-05-11
Create WS-C2950-24.yaml

ozairakhlaq:
Create definition for Cisco Catalyst 2950 24 port Switch.


jeremystretch:
Whoa, a blast from the past! :laughing:

Issue #821, opened 2022-05-25
Can APC NMC modules be converted from devices to modules?


twilley:
APC NMC modules are currently modeled as devices. Their definitions can be found in the following files:
- device-types/APC/AP9617.yaml
- device-types/APC/AP9618.yaml
- device-types/APC/AP9619.yaml
- device-types/APC/AP9630.yaml
- device-types/APC/AP9631.yaml
- device-types/APC/AP9640.yaml
- device-types/APC/AP9641.yaml

The device definitions look good, but adding them to a UPS poses some challenges. 

- Both devices must be created, with unique names and all the other required fields, before the NMC can be installed into the UPS.
- UPSes and NMCs are currently two discrete devices and I think a device/module relationship would be more appropriate.

For example, Consider the following device/module models:

UPS device model:
```
---
manufacturer: APC
model: SMT1500RM2U
slug: smt1500rm2u
part_number: SMT1500RM2U
u_height: 2
is_full_depth: true
console-ports:
  - name: Serial
    type: rj-45
  - name: USB
    type: usb-b
power-ports:
  - name: Source
    type: nema-5-15p
power-outlets:
  - name: Group 1 Outlet 1
    type: nema-5-15r
  - name: Group 1 Outlet 2
    type: nema-5-15r
  - name: Group 1 Outlet 3
    type: nema-5-15r
  - name: Group 2 Outlet 1
    type: nema-5-15r
  - name: Group 2 Outlet 2
    type: nema-5-15r
  - name: Group 2 Outlet 3
    type: nema-5-15r
module-bays:
  - name: NMC Slot
```
NMC module model:
```
---
manufacturer: APC
model: AP9641
part_number: AP9641
u_height: 0
is_full_depth: false
comments: UPS Network Management Card 3 with Environmental Monitoring
subdevice_role: child
interfaces:
  - name: Network
    type: 1000base-t
    mgmt_only: true
```

The UPS is the device, and the NMC can be installed in the NMC Slot module-bay.

I can provide a pull request if this is a good idea.

PR #822, opened 2022-05-26
add some APC UPS, XLBP, XFMR device models, convert NMC from device to module models

twilley:
I opened Issue #821 because I think APC NMC modules are better modeled as modules than devices now that NetBox supports them. This pull request includes that conversion, as well as some new APC device models that support these modules.

### DIFF
--- a/device-types/APC/APTF10KT01.yaml
+++ b/device-types/APC/APTF10KT01.yaml
@@ -1,0 +1,22 @@
+---
+manufacturer: APC
+model: APTF10KT01
+slug: aptf10kt01
+part_number: APTF10KT01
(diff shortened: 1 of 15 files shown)


jeremystretch:
Looks like some of the power ports are missing a type. Could you add these please?

twilley:
I think you're talking about the "XLBP Source" and "XLBP Outlet" power ports. I left those blank because I could not find the appropriate type for those at this time and because it appears to not be a required field.

These power ports are for expanding the battery capacity of UPS. Perhaps a "Battery Expansion" power port type could be created?

twilley:
For now, I might use the "DC Terminal" power port type for these.

PR #823, opened 2022-05-27
add or update several Juniper models

twilley:
I'm adding some models that I'd need (EX3300 and QFX5100 variants). Added u_height values, console ports, and interfaces to some existing models. Removed some incorrect interfaces (e.g. the EX3400 does not have four 40G ports.), and explicit "mgmt_only: false" on certain ports.

### DIFF
--- a/device-types/Juniper/EX2200-48T-4G.yaml
+++ b/device-types/Juniper/EX2200-48T-4G.yaml
@@ -2,6 +2,7 @@
 manufacturer: Juniper
 model: EX2200-48T-4G
 slug: ex2200-48t-4g
+u_height: 1
 interfaces:
   - name: ge-0/0/0
     type: 1000base-t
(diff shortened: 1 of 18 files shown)


jeremystretch:
```
=================================== FAILURES ===================================
____ test_definitions[device-types/Juniper/QFX5100-96S-8Q.yaml-schema1054] _____
device-types/Juniper/QFX5100-96S-8Q.yaml device type has duplicate slug "qfx5100-48s-6q"
======================= 1 failed, 1770 passed in 24.01s ========================
```

@twilley looks like there's a duplicate slug. I think this should be `qfx5100-96s-8q`.

twilley:
@jeremystretch Thanks for fixing that. It looks like I also failed to update the part_number for this model. It should look like:
```
model: QFX5100-96S-8Q
slug: qfx5100-96s-8q
part_number: QFX5100-96S-8Q
```

PR #826, opened 2022-05-29
Fix filename for N3K-C3064Q-10GE.yaml

abhi1693:


DanSheps:
N3K-C3064Q-10GX is a valid SKU, why do you think this needs to change?

abhi1693:
Because, it doesn't match with the part number that's defined inside. Even @ryanmerolle mentioned on the slack that it's a typo.

PR #833, opened 2022-06-03
Add some Cambium Networks Fixed Wireless gear

syncopsta:
Some fixed wireless gear from Cambium Networks


ryanmerolle:
The WiFi 7 gear could not even be imported into NetBox, as the ieee802.11ay spec has not been added to core.

Would you like to take a stab at raising an issue and PR against the main NetBox report to add ieee802.11ay in the same files that ieee802.11ax is defined?

danner26:
Do you still need this? @syncopsta

PR #836, opened 2022-06-14
Added 10/100 2530's, corrected model no in 2530-48G

mo-g:
Added 2530-24-PoE and 2530-48 non-poe. Still currently-supported devices which I'm using operationally!


mo-g:
@DanSheps Any problems with my "handiwork"? :)

ryanmerolle:
Seems like there is a conflict with device-types/HPE/2530-24-PoEP(J9779A).yml

You should delete your version and update the existing.

PR #854, opened 2022-07-11
More Aruba 5400 modules, and Leviton Keytone Port Patch Panels

dbusse-wcr7:


DanSheps:
They are both ended with a CRLF.  You need to convert that to a LF.  I would recommend grabbing notepad++ and fixing up all three of those.

ryanmerolle:
https://blog.boot.dev/clean-code/line-breaks-vs-code-lf-vs-crlf/#:~:text=The%20Quick%20Fix%20for%20%E2%80%9CEnd%20of%20line%20character%20is%20invalid%E2%80%9D&text=At%20the%20bottom%20right%20of,has%20the%20correct%20line%20breaks. if you use VS Code

ryanmerolle:
Now that I corrected the files to LF, you the tests proceeded and you are failing on something new.

dbusse-wcr7:
Thanks for the help! It should be ready to merge now.

PR #862, opened 2022-07-17
Add Corning Initial Edge Patch Panels and Modules

brainsfull:
Add Devices:

* EDGE-01U
* EDGE-01U-SP
* EDGE-02U
* EDGE-04U

Add Modules:

* ECM-UM-12-04-89G
* ECM-UM-12-05-93Q
* ECM-UM-12-05-93T
* ECM-UM-12-05-93V


brainsfull:
Will rework and update based on failed checks.

PR #868, opened 2022-07-26
Added Cisco Nexus 9332D-GX2B

andreassb:
Added Cisco Nexus 9332D-GX2B

### DIFF
--- a/device-types/Cisco/N9K-C9332D-GX2B.yaml
+++ b/device-types/Cisco/N9K-C9332D-GX2B.yaml
@@ -1,0 +1,87 @@
+---
+manufacturer: Cisco
+model: Nexus 9332D-GX2B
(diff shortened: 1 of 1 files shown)


device-types/Cisco/N9K-C9332D-GX2B.yaml
```diff
@@ -0,0 +1,87 @@
...
+u_height: 1
+is_full_depth: true
+console-ports:
+  - name: console
+    type: rj-45
+power-ports:
+  - name: Power Supply 1
+    type: iec-60320-c16
+  - name: Power Supply 2
+    type: iec-60320-c16
```
DanSheps, line 15: In my opinion, these should be in a module and these should be replaced with module bays

danner26:
Do you still need this? @andreassb

PR #869, opened 2022-07-26
Added model for Raritan DSX2-16

Houtek:
I've added the Raritan DSX2-16 (https://www.raritan.com/assets/ram/resources/data_sheets/raritan-ds-sxii.pdf)

### DIFF
--- a/device-types/Raritan/DSX2-16.yaml
+++ b/device-types/Raritan/DSX2-16.yaml
@@ -1,0 +1,68 @@
+---
+manufacturer: Raritan
+model: RT-DSX2-16
+slug: rt-dsx2-16
+u_height: 1
+is_full_depth: false
+console-ports:
+  - name: ADMIN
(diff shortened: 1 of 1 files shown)


device-types/Raritan/DSX2-16.yaml
```diff
@@ -0,0 +1,68 @@
...
+  - name: DVI
+    type: other
+  - name: Terminal
+    type: rj-45
+  - name: USB1
+    type: usb-a
+  - name: USB2
+    type: usb-a
+  - name: USB3
+    type: usb-a
```
DanSheps, line 19: I am assuming these are used for Keyboards/Mice as this is a KVM.  This is not a valid use of the "console ports" for the device type library

DanSheps (changes requested):

danner26:
Do you still need this? @Houtek

PR #879, opened 2022-08-15
Add/update Cisco Meraki MS250 switches and associated PSUs

Nothing4You:
I'm not entirely sure about the PSU config, but it seems that this is also being done this way for some other devices already.
One PSU is bundled with the switch and a second one can be installed for redundancy.
There's no requirement to always have a specific one, as long as at least one is present.
Both of them are modular.

### DIFF
--- a/module-types/Cisco/Meraki-MA-PWR-1025WAC.yaml
+++ b/module-types/Cisco/Meraki-MA-PWR-1025WAC.yaml
@@ -1,0 +1,9 @@
+---
+manufacturer: Cisco
+model: MA-PWR-1025WAC
+part_number: MA-PWR-1025WAC
+comments: '[1025W power supply unit product information](https://meraki.cisco.com/product/switches/switches-accessories/switches-accessories-power/1025-watt-power-supply/)'
+power-ports:
+  - name: PSU{module}
+    type: iec-60320-c14
+    maximum_draw: 1025
--- a/device-types/Cisco/Meraki-MS250-24.yaml
+++ b/device-types/Cisco/Meraki-MS250-24.yaml
@@ -1,15 +1,15 @@
 ---
 manufacturer: Cisco
-model: MS250-48FP
-slug: ms250-48fp
-part_number: MS250-48FP-HW
+model: MS250-24
+slug: ms250-24
+part_number: MS250-24-HW
 u_height: 1
 is_full_depth: false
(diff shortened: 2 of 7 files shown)


module-types/Cisco/Meraki-MA-PWR-1025WAC.yaml
```diff
@@ -0,0 +1,9 @@
+---
+manufacturer: Cisco
+model: MA-PWR-1025WAC
+part_number: MA-PWR-1025WAC
+comments: '[1025W power supply](https://meraki.cisco.com/product/switches/switches-accessories/switches-accessories-power/1025-watt-power-supply/)'
```
jsenecal, line 5: Wouldn't be the text for this link better as "Documentation" ?

jsenecal: Or even:
```markdown
1025W Power Supply for MS250-48FP
[Documentation](https://meraki.cisco.com/product/switches/switches-accessories/switches-accessories-power/1025-watt-power-supply/)
```

DanSheps: It isn't just documentation.  That is the accessory product page.  It probably could be more descriptive however

DanSheps: I would do the same as the switches.

"1025W power supply overview and specification"

Nothing4You: it's unfortunately the only piece of information that I could find for this at all.
there doesn't seem to be any published specs at all for the PSUs, besides the provided wattage.
the supported models list isn't exhaustive either, as for example the 250W PSUs are explicitly mentioned for MS250-24-HW and MS250-48-HW in https://meraki.cisco.com/lib/pdf/meraki_datasheet_ms250.pdf, yet the product page only lists MX250 and MX450.
what do you think about a more generic text like `1025W power supply product information`?
the `overview and specifications` part comes from the title of the associated page for the switches, which the PSUs don't really have.

Nothing4You: like this?

jsenecal (approved):
Looks very good to me 👍

PR #889, opened 2022-08-25
Update components.json

dmcken:
Added DOCSIS and GPON interface types.


schema/components.json
```diff
@@ -371,6 +371,13 @@
                         "t3",
                         "e3",
                         "xdsl",
+                        "docsis",
+                        "gpon",
+                        "xg-pon",
+                        "xgs-pon",
+                        "ng-pon2",
+                        "epon",
+                        "10g-epon",
```
DanSheps, line 380: Did not notice this before I brought in the pon ports.

Could you remove and then repush?  We still need DOCSIS added anyways

DanSheps (changes requested):
Did not notice this before merging my changes into master

dmcken:
Getting an error with all of the ADVA interfaces, I think its because they are all numeric and its expecting a string?

DanSheps (approved):

DanSheps (approved):

PR #902, opened 2022-09-06
Add cisco ubr 10012

dmcken:
Addition of base chassis.

Will add power, PRE, uplink and RF modules later.


device-types/Cisco/UBR-10012.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - name: rf-slot9
+  - name: rf-slot10
+  - name: rf-slot11
+  - name: rf-slot12
+  - name: power-slot1
+  - name: power-slot2
+  - name: pfe-slot1
+  - name: pfe-slot2
+  - name: timing-slot1
+  - name: timing-slot2
```
DanSheps, line 26: Position is missing for all of these.  Those should be entered to allow dynamic name creation.

DanSheps (changes requested):

DanSheps (approved):

PR #903, opened 2022-09-06
Update USW-Pro-48-PoE.yaml

suhl79:
Added PoE definitions and the secondary power interface


device-types/Ubiquiti/USW-Pro-48-PoE.yaml
```diff
@@ -3,118 +3,388 @@ manufacturer: Ubiquiti
...
+  label: ''
+  description: ''
+- name: Universal Input
+  type: iec-60320-c14
+  maximum_draw: 660
+  allocated_draw: null
+  label: ''
+  description: ''
+interfaces:
+- name: Port 1 (PoE)
```
DanSheps, line 27: I don't believe these port names are correct, I suspect they take the form of 0/# in the CLI.  Since we are updating the PoE we should also ensure port names are correct.

DanSheps (changes requested):

danner26:
Do you still need this? @suhl79

PR #904, opened 2022-09-08
Update PowerEdge_R640.yaml

dmcken:
Fix error with import:
[{"__all__":["Subdevice role of device type (PowerEdge R640) must be set to \"parent\" to allow device bays."]},{"__all__":["Subdevice role of device type (PowerEdge R640) must be set to \"parent\" to allow device bays."]},{"__all__":["Subdevice role of device type (PowerEdge R640) must be set to \"parent\" to allow device bays."]}]

### DIFF
--- a/device-types/Dell/PowerEdge_R640.yaml
+++ b/device-types/Dell/PowerEdge_R640.yaml
@@ -31,7 +31,10 @@ interfaces:
   - name: iDRAC9
     type: 1000base-t
     mgmt_only: true
-device-bays:
+module-bays:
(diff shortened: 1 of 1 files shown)


DanSheps:
This should be rewritten to use modules instead of device bays

DanSheps (changes requested):
Rewrite for modules instead of device bays

dmcken:
Drop the parent then?

dmcken:
I ask since this will then require positions which I don't know.

DanSheps (approved):

PR #913, opened 2022-09-15
Added Device WS-C2960XR-24PS-I

Corruptsector:
Added Device WS-C2960XR-24PS-I


device-types/Cisco/WS-C2960XR-48PS-I.yaml
```diff
@@ -0,0 +1,76 @@
...
+  - name: GigabitEthernet1/0/28
+    type: 1000base-x-sfp
+console-ports:
+  - name: con 0
+    type: rj-45
+power-ports:
+  - name: PSU0
+    type: iec-60320-c14
+  - name: PSU1
+    type: iec-60320-c14
```
DanSheps, line 76: PSU's should be modelled as module bays

DanSheps (changes requested):

device-types/Cisco/WS-C2960XR-48PS-I.yaml
Corruptsector: Hey Dan, I just followed how the existing 2960XRs are setup. Do you want me to create the PSU modules as well? Should i update the other XRs to be the same?

JimmyKip: Hi - would it be worth updating the contribution guide to make sure this is a clearly communicated standard? I don't disagree with the thinking in principle but there are many PR, including recent ones (eg #846 ) where the PSU have not been done as bays.

We just need to make sure its consistent & squares with the current note in the contribution guide:

_"Include only components which are fixed to the chassis. Optional modular components should be omitted from the definition. (Note that this does not exclude field-replaceable hardware that is expected to always be present, such as power supplies.)"_

DanSheps: > Hey Dan, I just followed how the existing 2960XRs are setup. Do you want me to create the PSU modules as well? Should i update the other XRs to be the same?

If you want to, yes, that would be ideal

DanSheps: > Hi - would it be worth updating the contribution guide to make sure this is a clearly communicated standard? I don't disagree with the thinking in principle but there are many PR, including recent ones (eg #846 ) where the PSU have not been done as bays.

I agree however sometimes the PSUs are fixed, sometimes they are modules.

I will see what I can do with the contrib guide tonight.

> We just need to make sure its consistent & squares with the current note in the contribution guide:
> 
> _"Include only components which are fixed to the chassis. Optional modular components should be omitted from the definition. (Note that this does not exclude field-replaceable hardware that is expected to always be present, such as power supplies.)"_

Corruptsector: I have updated that device, and was half way through adding some more. If youre happy with this ill update the other XR switches, make the power supply modules and then got some more devices to add.

device-types/Cisco/IE-4000-16GT4G-E.yaml
```diff
@@ -0,0 +1,51 @@
+---
```
DanSheps, line 1: Missing PSU it looks like.  From the datsheet, looks to be modular

device-types/Cisco/IE-4000-4S8P4G-E.yaml
```diff
@@ -0,0 +1,43 @@
+---
```
DanSheps, line 1: Possibly missing PSU on this one too.

DanSheps (changes requested):
Looks like 2 are missing PSU's.  If you want to add the module bays I can add the PSU's as a new module-type.

device-types/Cisco/IE-4000-4S8P4G-E.yaml
Corruptsector: These IE Switches run of external DC and do not have PSUs

danner26:
Do you still need this? @Corruptsector

PR #923, opened 2022-09-26
Create SO-DWDM-MUX-40CH-MON-1310W

hhansen06:
added SO-DWDM-MUX-40CH+MON+1310W


danner26:
Do you still need this? @hhansen06

PR #926, opened 2022-09-28
Update ASR1001-X.yaml

goebelmeier:
Refinements for ASR1001-X

### DIFF
--- a/device-types/Cisco/ASR1001-X.yaml
+++ b/device-types/Cisco/ASR1001-X.yaml
@@ -1,37 +1,102 @@
 ---
 manufacturer: Cisco
-model: ASR1001-X
-slug: asr1001-x
+model: ASR 1001-X
+slug: asr-1001-x
 part_number: ASR1001-X
-u_height: 1
-is_full_depth: true
-interfaces:
-  - name: GigabitEthernet0
-    type: 1000base-t
-    mgmt_only: true
-  - name: GigabitEthernet0/0/0
-    type: 1000base-t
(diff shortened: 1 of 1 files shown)


goebelmeier:
Oh, that's interesting. Should _all_ FRUs be modeled as a module? So far I understand that things like line cards are modeled as modules, but things like power supplies and fans are not.

DanSheps:
All should as they are swappable so it makes more sense to model them as modules so that they can be swapped out easily

device-types/Cisco/ASR1001-X.yaml
```diff
@@ -1,37 +1,102 @@
 ---
 manufacturer: Cisco
-model: ASR1001-X
-slug: asr1001-x
+model: ASR 1001-X
+slug: asr-1001-x
```
DanSheps, line 4: We shouldn't change this

device-types/Cisco/ASR1001-X.yaml
```diff
@@ -1,37 +1,102 @@
...
-    type: 10gbase-x-sfpp
+u_height: 1.0
+is_full_depth: false
+subdevice_role: ''
+airflow: front-to-rear
+comments: ''
 console-ports:
-  - name: con 0
-    type: rj-45
+- name: Console
```
DanSheps, line 12: Per the bot, needs indentation.  Please fix all locations.

DanSheps (changes requested):

danner26:
Do you still need this? @goebelmeier

PR #932, opened 2022-10-04
Added new Device Type: XBox-One-SeriesS.yaml

jhueppauff:


DanSheps:
I need to ask...

Why?

jhueppauff:
I have two of those documented through netbox - but I see your point, outside of a home lab or a streaming service the use is properly limited.

PR #937, opened 2022-10-11
Add Huawei S5731-S24T4X and modules

quequiel:
- Add Huawei S5731-S24T4X switch 
- Add Huawei S5735-L24P4X-A1 switch
Create Huawei in module-types and add
- FAN-023A-B (fan module) 
- PAC600S12-DB (Power supply module)


danner26:
Do you still need this? @quequiel

PR #945, opened 2022-10-22
Cisco 2821 and 2800 modules

quequiel:
- Add Cisco ISR 2821 
- Add Cisco VIC2-4FXO card for 2800 family
- Add Cisco AIM-VPN/EPII-Plus module for 2800 family
- Add Cisco PVDM2-64 DSP module for 2800 family


danner26:
Do you still need this? @quequiel

PR #899, opened 2022-09-03
Adding Brand-Rex Patch Panels and Keystone Modules

arcticash:
Created a popular Brand-Rex patch panel and its associated Keystone Modules

PR #950, opened 2022-11-03
Added Brand-Rex Patch Panels and Keystone Modules

arcticash:
Created a popular Brand-Rex patch panel and its associated Keystone Modules

_Re-opening of Pull Request #899 which was closed due to inactivity_


arcticash:
Is there anyone who can trigger the workflows on this PR please, it is currently waiting approval to run the tests

> First-time contributors need a maintainer to approve running workflows.

arcticash:
Can someone run the tests against this PR please?

danner26:
Hey @arcticash, everyone else has been busy but I am here now to work on these.. I checked your files and it looks like you are missing the ending newline. This will cause validation to fail. Tag me once you update the PR so I can re-review! Thanks

Also, as a tip for the future, if you use pre-commit as described in the readme, it will fix these issues for you. If you have any questions feel free to let me know

danner26:
@arcticash So I was thinking about it, unless the patch panel has some special functionality besides look, why dont we omit your implementation and just use the generic 24 port copper patch panel? We are going to be duplicating a lot of data if we start adding every single brand of patch panel and all of the varients they come in. What do you think?

danner26:
Also, what is the benefit of modelling every keystone jack as well? NetBox already has the idea of connected or not. I dont see the value in adding all of the different brand of patch panels and keystones when a generic representation should be enough in my opinion

danner26:
Hello @arcticash , are you able to work on this?

arcticash:
Sure, there is nothing specific about the patch panels themselves I just couldn't find anything suitable back when I started on the custom implementation, and then was trying to follow the guides about modelling specific components in their own files separately as they are removable.

Interestingly the files did have newlines at the ends as I did run the pre-commit however it had a moment and didn't seem to like being run on windows so may have done something weird with the files, oh well guess there not needed anymore anyway!

danner26:
So I do not think this implementation is necessary, @arcticash. The Generic folder already contains patch panels with 24 ports

You may open an issue in this repository if you believe there is a specific need for it, and we can discuss it with the other maintainers. However, as I mentioned, there is already such a device within the generic folder, so I do not believe that a new device is necessary just to accommodate angled versions (you could just put angled in the name)

PR #959, opened 2022-11-10
Add Cisco NCS-5508 and several line cards

ds2600:


danner26:
Do you still need this? @ds2600

PR #948, opened 2022-10-29
Improve existing Cisco C1000 switches and add missing ones

jnovak-netsystemcz:
Added all missing Cisco C1000 and C1000FE switches. Synchronized definition of Cisco C1000 switches.

PR #971, opened 2022-11-28
Add Generic half depth patch panels

jnovak-netsystemcz:
Added half depth variants of all generic patch panels
Airflow set to passive


danner26:
Hello, could you please update this PR to better describe all of the changes. For documentation purposes I think it would be good to have a more detailed change log

danner26:
Also @jnovak-netsystemcz if you take a look at the actions run, you have multiple issues. Please correct those so we can re-review

jnovak-netsystemcz:
I know. By mistake I pushed part of changes from #948. i'm waiting for merge of #948 (which is fine) to sync my fork and correct this.
BTW I don't understand why #948 is not merged when it is OK and approved...

danner26:
I will work on #948 now

danner26:
@jnovak-netsystemcz #948 has been merged

jeremystretch:
IMO we don't need full- and half-depth variants for each of these, given that they're generic representations. Why not just tweak the existing device types to be half-depth?

jnovak-netsystemcz:
We have both types in use (full/half depth) and we are checking whether all our devices matches its templates. When we tweak it, it complains over and over...
I believe other users might touch this issue too.

danner26:
@jnovak-netsystemcz so going with what you have, is there no full version of the `SC-24-double-port_fiber_patch_panel_half.yaml` and `SC-24-port_fiber_patch_panel_half.yaml`?

PR #974, opened 2022-12-01
Adding Fortinet FortiGate 600F and 601F devices

neilschelly:
These devices are identical as far as Netbox is concerned, like the 600E and 600F before. The 601F just includes some on-board SSD storage.


danner26:
Hey @neilschelly, looks like you forgot to change the slug on your fg-600f, currently it has `slug: fg-601f`

Can you please fix and push an update?

neilschelly:
@danner26 Sorry about that - copy/paste error, clearly. It should be good now. :crossed_fingers:

PR #999, opened 2022-12-27
Pre-commit formatting fixes

danner26:
I noticed a lot of files were failing pre-commit runs so I went through one by one and fixed any issues after the run


vittore:
Thanks. Done.

PR #1001, opened 2022-12-27
Added Wall boxes with 2, 3, 4 and 6 UTP plugs.

jnovak-netsystemcz:
Added Wall boxes with 2, 3, 4 and 6 UTP plugs.
They can be used for documenting physical wiring.

### DIFF
--- a/device-types/Generic/wall-box-2-utp.yaml
+++ b/device-types/Generic/wall-box-2-utp.yaml
@@ -2,7 +2,7 @@
 manufacturer: Generic
 model: Wall box, 2 UTP plugs
 slug: wall-box-2-utp-plugs
-u_height: 1.0
+u_height: 0.0
 is_full_depth: true
 airflow: passive
 front-ports:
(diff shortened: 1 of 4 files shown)


danner26:
A rack mounted wall box? I am confused what this actually is
Can you send a link to what you are representing here?

jnovak-netsystemcz:
No, it is non-rack device. We use plenty of different types so I created it as generic. There is example:

https://www.vseprowifi.cz/utp-kabely-konektory-zasuvky-w-star-zasuvka-utp-2xrj45-na-zed-narazeci-cat5e-kompletni-na-stenu-p-805.html

danner26:
You have `u_height: 1.0` set on these. If they are not rack mounted then they should be `u_height: 0`

jnovak-netsystemcz:
I see. Set to 0.

danner26:
I think you should omit both `is_full_depth` and `airflow`. Airflow isnt really important with these wall mounted utp ports, and is_full_depth defaults to true anyway.

jnovak-netsystemcz:
Done

danner26:
@jnovak-netsystemcz I was talking with someone else, it makes sense to actually mark these as `is_full_depth: false` since I guess if they get placed in a rack theyre not a full depth item. I am sorry about that.. Once more change and we should be good to go

jnovak-netsystemcz:
I agree with it. Done.

danner26:
Changes merged! Thanks @jnovak-netsystemcz

PR #1005, opened 2022-12-30
Create TL-SG105E.yaml

jeffersonraimon:


danner26:
@jeffersonraimon please add the power port(s)
Also correct me if I am wrong, but this is not rack mountable so it should have `u_height` set to `0`

jeffersonraimon:
Done, check again pls @danner26

danner26:
@jeffersonraimon please correct the power port type, this unit uses an external power adapter that delivers DC current. So please make the type `dc-terminal`

PR #1007, opened 2022-12-30
Cisco 2821

quequiel:
Add Cisco 2821 and modules


danner26:
This looks good, if you are ready please mark it as ready for review

danner26:
Please add `airflow` if you can

PR #1011, opened 2022-12-30
Added RB5009UPr+S+IN

Wlanfr3ak:
New Type with all Details

### DIFF
--- a/device-types/MikroTik/RB5009UPr+S+IN.yaml
+++ b/device-types/MikroTik/RB5009UPr+S+IN.yaml
@@ -1,0 +1,50 @@
+---
+manufacturer: MikroTik
+model: RB5009UPr+S+IN
+slug: rb5009upr_s_in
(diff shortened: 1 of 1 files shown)


Wlanfr3ak (commented):
Fixed Typo

Wlanfr3ak (commented):
Fixed Typos around PoE

danner26:
Please set `u_height`, `is_full_depth`, and `airflow` as applicable

Wlanfr3ak:
Device is without Rack Mount so no of the 3 values is used (passive cooling)

danner26:
Then `u_height` should be set to `0` as it defaults to 1
`is_full_depth` should be set to `false` as it defaults to true
`airflow` should be set to `passive`

Wlanfr3ak (commented):
Added more Details as wished

PR #1017, opened 2022-12-30
Update RB2011UiAS-RM.yaml with more Details and PoE

Wlanfr3ak:
Added more Details and the PoE Details to the Device-Type


Wlanfr3ak (commented):
Typo

PR #1027, opened 2023-01-04
AIP/ CSC SSM updates for the ASA

sebrupik:
Various tweaks to the old but gold ASA5500 line of firewalls. Added module bays where present and included applicable SSM/ SSC module types.

### DIFF
--- a/device-types/Cisco/ASA5505.yaml
+++ b/device-types/Cisco/ASA5505.yaml
@@ -16,35 +16,26 @@ power-ports:
 interfaces:
   - name: Ethernet0/0
     type: 100base-tx
-    mgmt_only: false
   - name: Ethernet0/1
     type: 100base-tx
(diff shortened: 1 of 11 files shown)


danner26:
Hey @sebrupik thanks for the fixes and additions. `100base-t` is not a valid type of interface. Please use `100base-tx` as it is the only interface type availible at the moment

We have a [PR](https://github.com/netbox-community/netbox/pull/11377) open with the upstream NetBox repo to add other 10/100 adapters, including `100base-fx (fast ethernet over fiber), 100base-lfx (proprietary fast ethernet over fiber), and 100base-t1 (10/100 single pair)`

If you need another adapter we will have to edit that PR or create a new one for the type you need. I think in this case if they are regular ethernet ports, `100base-tx` is what you want

danner26 (approved):

PR #1044, opened 2023-01-09
Patch 3

LHBL2003:
Add —- to begin


danner26:
Hey @LHBL2003 I fixed the indentation issue for you
Also, mgmt-only defaults to false, so no need to define it

danner26:
Are the adapters in the device config actually referred to as `1`, `2`, etc? Or are they referred to somewhat like a Cisco switch port? The name of the interface should reflect exactly how the config references them, for example: `GigabitEthernet1/0/1`, `GigabitEthernet1/0/2`, etc

danner26:
Hey @LHBL2003, I fixed all of the issues. Once you verify the port names we can merge this

LHBL2003:
On the web it is represented like this:
GE1/0/13

For example, in the secret Extended CLI, the command is written like this:
interface Gigabit Ethernet 1/0/13

The output in the CLI looks like this:
Gigabit Ethernet1/0/13

See also:
https://www.pbdigital.org/post/2019-07-22-vlans-hp-v1910-cli/

Sorry I'm at home and don't have my computer with me :D

danner26:
@LHBL2003 All good. It looks like `interface GigabitEthernet 1/0/6` is used to configure the interfaces, so let's go with that

danner26:
Thanks for your contribution! @LHBL2003

PR #1045, opened 2023-01-10
Add APC SMT2200RMI2UNC

tschuxxi:
APC SMT2200RMI2UNC UPS system

### DIFF
--- a/device-types/APC/SMT2200RMI2UNC.yaml
+++ b/device-types/APC/SMT2200RMI2UNC.yaml
@@ -45,4 +45,3 @@ power-outlets:
     power_port: Source
 module-bays:
   - name: SmartSlot
-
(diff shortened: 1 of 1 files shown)


tschuxxi:
Did I misinterpret the contributing guidelines here? They state: End each definition file with a blank line

danner26:
@tschuxxi That is absolutely correct, but if you look in your file, you actually have 2 newlines.
image

Your best bet is to use `pre-commit run` when staging your changes, which will fix/notify you of issues like this. You can also use `pre-commit run -a` which will run the checks against the entire repo. Just be careful if something else needs refactoring to not commit it in the same PR by accident.

danner26:
Thanks for your contribution @tschuxxi

PR #1046, opened 2023-01-10
Added Mellanox, NEC, Netgear, and Yamaha switches/routers

reishoku:
This PR contains some new switches/routers from Mellanox, NEC, YAMAHA, NETGEAR.

Sorry for its (large) size. It is mostly because Mellanox and YAMAHA switches had numerous ports.

-----
Signed-off-by: KOSHIKAWA Kenichi <reishoku.misc ~@~ pm.me>


danner26:
Please remove all definitions of `mgmt_only: false`. These default to false, so no need to explicitly define them.
Normally I would go ahead and fix it for you but I noticed a bunch of your files have them, so it would be easier for you to do so. Thanks!

reishoku:
Thanks for the comment. I assume I fixed.

reishoku:
Sorry. It seems that I mistakenly modified some files... Will fix ASAP.

reishoku:
Checked the changes I made.
I assume I'm ready.

danner26:
@reishoku Thanks so much for the changes. Just a heads up I am swamped at work so timing might be delayed in getting back to you/making changes/requesting revisions

That being said, I just want to verify a few things as I have not worked with all of these devices before. Can you please confirm:

1. For the Mellanox devices, in the configuration are the ports configured using `swp1` or do they have full names such as the Cisco `GigabitEthernet1/0/1`? If you could post a sample config of a few interfaces for verification that would be helpful as well. Otherwise the Mellanox device definitions look good!
2. Same question for the NEC devices, in the configuration are they actually called `GigaEthernet2 port 1` for example? A sample of your config would be appreciated for this as well. Otherwise they look good!
3. The Netgear GS116 is an unmanaged switch, no? I haven't used one of these in eons lol. The last time I worked with a NetGear pro series like this (managed though), I believe the interfaces were called `g1`, `g2`, etc but that was in the web config. Do these actually have a config to reference? Last I checked there is no web interface, am I correct? I am not against calling them this in the case that there is no actual config to reference since it aligns with the other already defined devices in the repo.
4. Please add the power for the Netgear GS116. You can use type `
5. Same question as number 3 for the Netgear GS116E.
6. Same question as number 3 for the Netgear JGS524.
7. Same question about the port names for all of the Yamaha devices. In the configuration are they actually called `port1.1` for example? A sample of your config would be appreciated for this as well.
8. For the Yamaha `SWX2221P-10NT.yaml`, `port1.9` and `port1.10`, what is the default port speed? We generally make sure these configs have the default values, and then the person using the device after import can change them if that suits their configuration. Can you confirm the default speed for these ports are `10gbase-t`?
8.a Same question for the Yamaha `SWX2310P-10G.yaml`
8.b Same question for the Yamaha `SWX2310P-18G.yaml`

danner26:
Hello @reishoku , just nudging you again to see if you have availability to work on this. I would like to get it merged before the PR gets marked stale

reishoku:
Sorry for the late reply, and thank you for your time to review.

> For the Mellanox devices, in the configuration are the ports configured using swp1 [...]

I afraid to say that I do not own/have circumstance(s) to use these Mellanox switch(es). Devicetype configurations were created based entirely on datasheets.
However, according to the documentation published by NVIDIA( https://docs.nvidia.com/networking-ethernet-software/cumulus-linux-53/Layer-1-and-Switch-Ports/Interface-Configuration-and-Management/ ), it looks like interfaces are named `swp1 swp2 [...]`.

> 2. Same question for the NEC devices, in the configuration are they actually called GigaEthernet2 port 1 for example? [...]

I own NEC UNIVERGE IX2215, and yes, for `GigaEthernet` part, the interfaces are actually called `GigaEthernet` . For port numbers part, I will fix them ( `GigaEthernet2 port 1` will be called as `GigaEthernet2.1` ).

> 3. The Netgear GS116 is an unmanaged switch [...]
> 5. [...]
> 6. [...]

Yes, you are right; will fix (align with already defined configs).

> 4. Please add the power for the Netgear GS116. You can use type `

May I ask which type you were intended? `dc-terminal` ?

> 7. Same question about the port names for all of the Yamaha devices.  [...] are they actually called port1.1 for example? [...]

Yes. You can access documentation published by YAMAHA ( e.g. http://www.rtpro.yamaha.co.jp/RT/manual/swx2210/cmdref/interface/basic/show_interface.html ) and search `show interface` and check port name(s) if needed.

> 8. For the Yamaha SWX2221P-10NT.yaml, port1.9 and port1.10, what is the default port speed?

I cannot find any specific reference to the default port speed. However, considering YAMAHA switches automatically negotiates the speed by default, the default port speed will be 10G.

> Can you confirm the default speed for these ports are 10gbase-t?

Users choose which they use RJ-45 port or SFP+ port, so there is no "confirmation" ( user chooses the port, the chosen port will automatically link up ), I assume.

danner26:
Thanks for getting back to me @reishoku! Sorry about number 4, the type should be `dc-terminal` indeed

Everything else sounds good! Let me know when your changes are implemented (or request another review) and I will re-review.

reishoku:
@danner26 , sorry for taking time. I updated files. I think I'm ready.

PR #1048, opened 2023-01-11
Create CRS226-24G-2S+RM.yaml Device Type model

Wlanfr3ak:
Create new Device Type for Mikrotik CRS226-24G-2S+RM

### DIFF
--- a/device-types/MikroTik/CRS226-24G-2S+RM.yaml
+++ b/device-types/MikroTik/CRS226-24G-2S+RM.yaml
@@ -1,0 +1,72 @@
+---
+manufacturer: MikroTik
+model: CRS226-24G-2S+RM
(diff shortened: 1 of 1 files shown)


danner26:
Hey @Wlanfr3ak thanks for the contribution
Can I just confirm that in the configuration, the interfaces are referenced as `ether2` for example? If you could post a sample if your configuration that would be appreciated. Thanks!

Wlanfr3ak:
Mikrotik has on every device the same name scheme in standard. You can rename it but in default configuration it start with ether1, ether2 etc. Reference: https://wiki.mikrotik.com/wiki/Manual:Interface/Ethernet

danner26:
Perfect, thanks!

PR #1052, opened 2023-01-11
Added stack ports

amyers1616:
This device has 2 stack ports so I've added them here.


danner26:
Hey @amyers1616 are these stack ports built in by default or are they modular? If they are modular, please move both of the stack ports to a [module-bays](https://github.com/netbox-community/devicetype-library#module-bays) and add the relevant stacking kit to the module-types folder

If they are built in just let me know and we should be good to go. Thanks!

amyers1616:
Hi @danner26 - they are built-in. It looks like from this (https://www.cisco.com/web/ANZ/cpp/refguide/hview/switch/3750.html) that all the 3750 models have 2 stack ports since they provide only 1 rear view for all the systems.

danner26:
Perfect, thanks!

PR #1055, opened 2023-01-16
Create XGM3S-2xf.yaml

renaniam:
Template for XGM3S-2xf - Dual SFP+ Module for X460-48T

PR #1058, opened 2023-01-18
@renaniam XGM3S-2xf creation - module types

danner26:
Assisting @renaniam with moving this module definition they created to the module-types
Reference is #1055


danner26:
@renaniam Please review this and make sure it is ready to be merged, I have created it in the module-types folder

danner26:
@renaniam I fixed the extra variables that do not belong in module-types. Once again though, are the ports actually referred to as `53` and `54`? Or is this just the position you have them installed in?

I have named them `{module}-port1` and `{module}-port2` for the time being. If you could send an output of the config for these interfaces that would work as well

renaniam:
@danner26

In case of X460-48T theres 52 GE Ports (48 GE Cooper + 4 GE SFP), but there's others Summit models with less than 48 ports.

Diferentily of others devices, when you put a module there's no slot number, in that Extreme Switch model are just add other interfaces.

So, without module we willl find 52 ports GE, when you put this module will be incriased more 2 interfaces numbered by 53 and 54.

There's a XGM3SB-4sf, in that case will be incriased more 4 ports 53, 54, 55 and 56. 

In case of SummitStack Module or SummitStack-v80 Module always be the unreacheable ports as 57 and 58 because is not in interface list.

Thanks for all!

danner26:
Hm, I am thinking of how to handle this then @renaniam 
If you slot this into a device with 24 ports, what happens? Are they still referenced as `53`, `54` or are they now `25` and `26`? Sorry for the confusion.

renaniam:
Unfortunately I'm don't had one to make a test, but I suppose that's correct.

danner26:
I dont have a great way to model this then, since they do not have a unique name this is going to be a tough one. I will add the `help wanted` tag and hopefully someone with one of these modules can give us an example output. For now I will mark this PR as a draft as well

danner26:
```
weight: 12.01
weight_unit: lb
```

Issue #1060, opened 2023-01-18
Update device type creation/update guidelines for device and module weight


xkilian:
### Documentation update.
Device types and modules can now support the weight value starting in netbox 3.4.
Include the weight in the device-template examples and module examples.

danner26:
The PR has been merged! Thank you for bringing this up

PR #1061, opened 2023-01-19
Added weight to device and module types

danner26:
Added the required validation information for the addition of weight as described in #1060


danner26:
@xkilian this PR will allow for weight to be a valid definition. If you are utilizing [Netbox-Device-Type-Library-Import](https://github.com/minitriga/Netbox-Device-Type-Library-Import), they will need to update their code to import this data

PR #1067, opened 2023-01-26
Extreme 9k

mayafox:
Added in Extreme 9920, 8520, and 8720 products, fixed incorrect wattage for 7400-48y, and removed mentions of AC from the model and parts for the MLXe and SLX 9xxx boxes.


mayafox:
Any guidance on resolving this? last I checked 0.01 would go into 10.11 just fine, and the yaml matches up with the sample output format shown in the main readme.md

```
weight:
  - value: 10.11
    unit: kg
```
device-types/Extreme Networks/8520-48Y-8C.yaml failed validation: 10.11 is not a multiple of 0.01

Failed validating 'multipleOf' in schema['properties']['weight']['items']['properties']['value']:
    {'minimum': 0, 'multipleOf': 0.01, 'type': 'number'}

On instance['weight'][0]['value']:
    10.11

danner26:
Hey @mayafox, the weight definition is brand new and we are clearly still tackling some issues around that. I will use this as a testing ground to fix that issue as well. Could you please invite me to your fork of this repo? That way I can commit directory to it from vscode. I would appreciate it!

danner26:
@mayafox I have merged PR #1069 which resolves your issue. We can use `multipleOfPrecision` instead of `multipleOf` which resolves the issue. I have tested this with `10`, `10.1`, `10.11`, etc.

danner26:
@mayafox I started to update some of the weights but I do not know the actual values. I have merged the master branch back into yours, so you should be able to correct the weight values now. Can you please update the weights?

Please also make sure the weights are as accurate as possible, and in a `comment` on the device, please add a link to the device documentation. Thanks!

mayafox:
@danner26 
Changes in support of fixing the weight look good!
Device documentation should be the hardware information URL, and not the OS admin guides I presume?

danner26:
Hardware info url yes, but if you have both feel free to throw them both
in!

cwegener:
@danner26  Just a quick note that you didn't actually "fix" the original issue with the `multipleOf` keyword. In #1069 you have unknowingly disabled the `multipleOf` keyword.

There is no such thing as a `multipleOfPrecision` keyword in the JSON Schema specification.

danner26:
@cwegener Thanks for the heads up, I am guessing that [this issue](https://github.com/python-jsonschema/jsonschema/issues/679) isnt an actual solution then. Any idea why it works correctly then?

Do you have a proper solution for this?

cwegener:
> Do you have a proper solution for this?

Nope. I just noticed the obvious error and thought I point it out.

I'm in a similar situation like yourself. I'm aware of how floating point arithmetic works in Python (and any other language using binary for floating point) but also have no experience on how to properly apply libraries like `jsonschema` as part of the whole ecosystem.

My case is even more complex as I ran into this issue in the context of a distributed system where the JSON Schema spec is the 'connecting' tissue but the implementations that send data to each other are blissfully unaware of each other. (https://github.com/singer-io/getting-started/blob/master/docs/SPEC.md)

danner26:
Gotcha, I appreciate you pointing it out @cwegener ! Here's to hoping we find a solution

danner26:
@cwegener thoughts on using `divisibleBy: 0.01` rather than `multipleOf`?

danner26:
@mayafox we fixed the decimal issue, please try applying the proper weight values again. Thank you for your patience

mayafox:
@danner26 Done!

danner26:
I am a little confused, what is going on with all of the renames such as the `BR-MLXE-16-MR2-AC` being renamed to `BR-MLXE-16-MR2`? I see the [BR-MLXE-16-MR2-AC](https://www.terabitsystems.com/brocade/routers/br-mlxe-16-mr2-x-ac) listed here. 

We shouldnt be removing or renaming device definitions, you should add new files for different devices. Unless I am missing something. Can you explain? @mayafox

mayafox:
> 
So there's a couple issues here with the -AC models.  I chatted with one of my colleagues on adding in all the various Power supply (AC/DC) and air-flow combinations that are listed, and they viewed it as a moot point and not worth the effort (not sure they fully understood the question either). I also asked you what the projects preference was, but you missed that question as well.   I don't mind doing the work to add in the variations, but a clear direction would be good.

danner26:
Sorry about that, I totally missed that questions. So the only thing the -AC is used for is when the AC power module is installed? If that is the case, then I agree it is a moot point. I think modelling them without the AC makes sense, and then the modules that you install (AC or DC power supply) is what will determine which power type it uses. Is that understanding where your changes came from?

mayafox:
> 
Yes, and for some boxes (MLX and the 9920) it is possible to have both AC and DC power supplies installed and in service. Airflow is generally uniform, and can impact the power supplies choices as well, but most of the platforms allow you to switch by swapping the AC PSU for DC, and swapping fans and PSU's to change the air flow. It's a mess of permutations I would rather not poke if I can avoid it. 
For information purposes I am modeling them with the correct wattages, and AC connections and front to rear airflow.

danner26:
So I see what the mess is here, this is quite the issue. Let me ask you this, if I were to go to the vendor and ask for a BR-MLXE-16-MR2-AC, would they provide me with the unit and the AC power supplies? Like do they actually list for sale `BR-MLXE-16-MR2-AC` that includes all of the kit, plus the `BR-MLXE-16-MR2` which doesnt include the PSU? Or do you have to buy either a `BR-MLXE-16-MR2-AC` or a `BR-MLXE-16-MR2-DC`?

Because this library acts as a source of truth for a specific device, I do not think we should be removing/renaming legitimate models even if they do have the option to be changed on the fly. If a product can be purchased by default under the `BR-MLXE-16-MR2-AC` sku, it should have its own file. Just as if you can buy the product without the AC/DC PSUs, you should also have another file that is `BR-MLXE-16-MR2`. 

Does that make sense? Unless you cannot buy a `BR-MLXE-16-MR2-AC` and you can only buy a `BR-MLXE-16-MR2`, then I think ultimately you should revert the renames/changes in those files.

mayafox:
@danner26 
The TLDR answer is yeap it makes sense and I will update the platform files accordingly.  Thanks for providing a clear direction on what's expected.

"Let me ask you this, if I were to go to the vendor and ask for a BR-MLXE-16-MR2-AC, would they provide me with the unit and the AC power supplies?" Based of the ordering information page from the data sheet, yeap, they used to list the bare chassis sans mgmt, and power, but that's been removed, 

```BR-MLXE-16-MR2-M-AC
MLXe-16 AC system with one MR2 (M) management module, three high-speed switch fabric modules, four 1,800W
AC power supplies, two exhaust fan assembly kits, and air filter. Power cord not included.
BR-MLXE-16-MR2-M-DC
MLXe-16 DC system with one MR2 (M) management module, three high-speed switch fabric modules, four 1,800W
DC power supplies, two exhaust fan assembly kits, and air filter. Power cord not included.
BR-MLXE-16-MR2-X-AC
MLXe-16 AC system with one MR2 (X) management module, three high-speed switch fabric modules, four 1,800W
AC power supplies, two exhaust fan assembly kits, and air filter. Power cord not included.
BR-MLXE-16-MR2-X-DC
MLXe-16 DC system with one MR2 (X) management module, three high-speed switch fabric modules, four 1,800W
DC power supplies, two exhaust fan assembly kits, and air filter. Power cord not included.
```

mayafox:
@danner26 
Does this fall in line with what you expect?

danner26:
Hey @mayafox take your time.. let me know when you are ready for me to review it again
And yes that looks like what I am thinking, whatever is the default configurations from the vendor is what we should define. The idea is that this is a repo that you can import the standardized models based off of, and then after imported and created in NetBox, the user can modify the device in NetBox from the default configuration to model the real world. Let me know if you need any clarification, I hope I am making sense

danner26:
```
weight: 12.01
weight_unit: lb
```

mayafox:
Closing for rework of the code.

PR #1071, opened 2023-01-30
JSON Schema Weight Validation Fix

### LINKED ISSUE #1068: Weight Validation
During validation of a new addition seen in PR #1067, @mayafox noticed that using a weight of `10.11` would return an invalid configuration, when in fact `10.11` is a multipleOf `0.01`.

danner26:
Testing a fix for the multipleOf issue using info in https://github.com/python-jsonschema/jsonschema/issues/1039

Not currently working correctly, though the developer says it should be. We will need to diagnose this further
Afterwards it will close #1068

### DIFF
--- a/schema/components.json
+++ b/schema/components.json
@@ -605,7 +605,7 @@
                 "value": {
                     "type": "number",
                     "minimum": 0,
-                    "multipleOfPrecision": 0.01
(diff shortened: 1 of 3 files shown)


danner26:
@jsenecal was able to fix this in #1073 so this can be closed

PR #1073, opened 2023-01-31
Fix float validation using decimal.Decimal

### LINKED ISSUE #1068: Weight Validation

jsenecal:
- Introduce the usage of `decimal.Decimal` in order to catch inconsistencies with operations against floats in Python
- Created `DecimalSafeLoader` to deserialize YAML floats as `decimal.Decimal`
- Added file handler for `RefResolver` in order to also cast `decimal.Decimal` to floats loaded through a $ref.

Closes #1068

### DIFF
--- a/schema/components.json
+++ b/schema/components.json
@@ -605,7 +605,7 @@
                 "value": {
                     "type": "number",
                     "minimum": 0,
-                    "multipleOfPrecision": 0.01
+                    "multipleOf": 0.01
                 },
                 "unit": {
                     "type": "string",
--- a/tests/definitions_test.py
+++ b/tests/definitions_test.py
@@ -1,11 +1,14 @@
+import decimal
 import glob
 import json
 import os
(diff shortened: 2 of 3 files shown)


danner26:
Just ran some tests against this PR, looks like it solves the issues along with the `10.11` pesky one we were fiddling with
Everything looks good to me, are we ready to merge this to master or is there anything else you need to do? @jsenecal

jsenecal:
> Just ran some tests against this PR, looks like it solves the issues along with the `10.11` pesky one we were fiddling with Everything looks good to me, are we ready to merge this to master or is there anything else you need to do? @jsenecal

All good! This was fun to reverse-engineer!

PR #1074, opened 2023-01-31
Create FS FMU-MC042733 CWDM Module

Wlanfr3ak:
Create FS FMU-MC042733 CWDM Module as child object

Parent is: https://github.com/netbox-community/devicetype-library/blob/master/device-types/FS/FMU-1UFMX-N.yaml

Reference:
https://www.fs.com/products/42972.html

### DIFF
--- a/device-types/FS/FMU-MC042733.yaml
+++ b/device-types/FS/FMU-MC042733.yaml
@@ -1,0 +1,30 @@
+---
+manufacturer: FS
+model: FMU-MC042733
+slug: fmu-mc042733
+comments: '[4 Channels 1270-1330nm, LC/UPC, Dual Fiber, Low Insertion Loss CWDM Mux Demux, FMU Plug-in Module](https://www.fs.com/products/42972.html)'
+part_number: '42972'
+u_height: 0
+is_full_depth: false
+subdevice_role: child
+front-ports:
+  - name: '1270'
+    type: lc
+    rear_port: Line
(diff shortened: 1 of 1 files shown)


device-types/FS/FMU-MC042733.yaml
```diff
@@ -0,0 +1,30 @@
+---
+manufacturer: FS
+model: FMU-MC042733
+slug: fmu-mc042733
+comments: '[4 Channels 1270-1330nm, LC/UPC, Dual Fiber, Low Insertion Loss CWDM Mux Demux, FMU Plug-in Module](https://www.fs.com/products/42972.html)'
+part_number: '42972'
```
danner26, line 6: According to the website, the FS P/N is `FMU-MC042733`. I see `#42972` in reference to the FS product page.. Since the FS part number is actually `FMU-MC042733` I think your file should reflect this

danner26 (commented):
Everything else looks good, please review my comment and let me know what you think

device-types/FS/FMU-MC042733.yaml
Wlanfr3ak: Look at the Picture from them: left side same modelname and left side this part number. Its correct so, its confused me too ;-)

danner26: Yup I see what you mean, okay lets roll with that

PR #1088, opened 2023-02-03
ZPE Nodegrid Gate SR addition

biwhite:
Extra device, serial/ethernet out of band device combined


danner26:
Hello @biwhite thank you for your contribution
Please add `zpe_` as a prefix to the `slug`. Also, please make the `slug` lowercase

Other than that, everything looks good. Is there anything else you need to do?

biwhite:
Tweaked Slug as requested.  Seems inconsistent across vendors/devices as to which include vendor name or not?

PR #1089, opened 2023-02-03
Create a template for Inspur NF8260M6

sulik76:
Create a template for Inspur NF8260M6 server platform


danner26:
Hello @sulik76 thank you for your contribution

VGA ports are not really meant to be modelled. Console ports is supposed to be for console access, i.e. a db9 port on a switch.
Same goes for USB ports that are not console ports

CPUs are not meant to be modelled in NetBox currently. We are in the process of talking about adding storage management, maybe one day more granular device management. If you really want to model these, I would suggest inventory items after you import this (do not add them in this definition file).
Please remove CPU and DIMM definitions from the file.

sulik76:
Hello @danner26 
I corrected the template according to your instructions.

PR #1095, opened 2023-02-05
Few changes proposal

requenag:
Hi,

I would like to propose the few changes:
- Detox device-types folder. (reason: Netbox is a Linux based solution and might not like the '&' and space around folder/files. The sooner we fix this the better we will collaborate in my opinion)
- Detox device-modules folder. (same as above)
- Adding a device-type-imgs to add front/rear image ready to be imported on Netbox. (just to share all the one I have)
- Adding Intel NIC cards (Modules approach).

Regards,


danner26:
Hello @requenag we are not quite ready for the images yet. We are still discussing internally about where to properly store these images (most likely will be its own repo). Once we are ready for that, we will let you know

In the meantime, please remove images from your PR. Once you are ready let me know and I will review any changes you may have

requenag:
Hi Daniel,

Thank you for your reply.

I haven't find a simple way to modify this Pull Request the way you would
like.
Can I propose you cancel this Pull Request and I will make the 2 detox and
the proposed changes on a separate PR This will be more convenient for both
of us.


Regards,

danner26:
That would work better, I will close the PR now and you can implement a different PR for the modifications you want to add now

PR #1098, opened 2023-02-06
Create QSW-6900-32H.yaml

krasotinpa:
QSW-6900-32H switches are data center oriented high density 40G/100G access switches, providing 40G/100G access, low latency, and complete data center features.


danner26:
Thank you for your contribution!

PR #1099, opened 2023-02-06
Create QSW-6900-56LF.yaml

krasotinpa:
QSW-6900-56LF switches are data center oriented high density 10G access switches, providing 10G/100G access, low latency, and complete data center features.

### DIFF
--- a/device-types/QTECH/QSW-6900-56LF.yaml
+++ b/device-types/QTECH/QSW-6900-56LF.yaml
@@ -1,0 +1,193 @@
+---
+manufacturer: QTECH
+model: QSW-6900-56LF
+part_number: QSW-6900-56LF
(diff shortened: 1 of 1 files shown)


danner26:
Hello, all looks good. What is the meaning of the labels if I may ask?
Also, do you have `airflow` and `weight` information? If so, please populate that. Also, if you have a link to the datasheet we would appreciate you providing that in a `comments`. Thanks!

krasotinpa:
Label - An alternative physical label identifying the interface.
I define it as smal mark near the interface or console port on front or rear panel.

danner26:
Is the label specific to you? Or is the label a standard for this device across the industry? If it is specific to you/your org please remove it. Otherwise sounds good and we can merge this after I have your confirmation

PR #1102, opened 2023-02-06
Add Huawei S5700-28X-LI-AC switch

madecto:
Add Huawei S5700-28X-LI-AC switch


danner26:
Hello, all looks good. 
Also, do you have `weight` information? If so, please populate that. Otherwise just let me know. Thanks!

PR #1103, opened 2023-02-06
Adding equipment from the manufacturer DATACOM.

rogeriomendesp:


rogeriomendesp:
Hi, help.

I have this equipment to add https://www.datacom.com.br/en/produtos/gpon/dm4618 , but it has a slot, this slot cryo device bay or module bay?

What about the board, do I create it as a child device type and the chassis as a parent?
https://www.datacom.com.br/uploads/product/file/78e9b89bab08dcd67045f3e3f4d4c9ae.pdf

danner26:
Since this is still a WIP, I have converted this to be a draft PR

Here is the best way to think of it @rogeriomendesp, Device bays represent a space or slot within a parent device in which a child device may be installed. For example, a 2U parent chassis might house four individual blade servers. The chassis would appear in the rack elevation as a 2U device with four device bays, and each server within it would be defined as a 0U device installed in one of the device bays. Child devices do not appear within rack elevations or count as consuming rack units. A good example of this is a Dell server module that slots into a VRTX chassis. The VRTX chassis is a 4U device and the server module slots into that chassis device.

A module type represents a specific make and model of hardware component which is installable within a device's [module bay](https://docs.netbox.dev/en/stable/models/dcim/modulebay/) and has its own child components. For example, consider a chassis-based switch or router with a number of field-replaceable line cards. Each line card has its own model number and includes a certain set of components such as interfaces. Each module type may have a manufacturer, model number, and part number assigned to it. A module bay can also be used for things like nic cards, modular PSUs, etc.

It looks like the DM4618 is a device. According to the website it has `an expansion slot that allows you to equip it with 32 more GPON ports`. This makes me believe that the expansion part would be a module.

rogeriomendesp:
Hi,

I would like a status of this request.

Thanks.

rogeriomendesp:
Hi,

I would like a status of this request.

Thanks.

danner26:
Hello, have you made the requested updates yet? Also, please ensure all checks are passing.

Once you are ready for this to be reviewed, please mark this PR ready for review (it is currently in a draft status).

danner26 (changes requested):
Hello, thanks for this. Some notes about changes that I believe may be applicable to most if not all of your added files

1. Slugs: all slugs should have the manufacturers name appended. I.E. `datacom-`
2. Please add `weight` and `weight_unit` to any devices which you can find that information
3. Please add `airflow` to any devices which you can find that information

PR #1104, opened 2023-02-07
Detox device-types and module-types folder for Linux user

requenag:
Hi,

Understand this is a big change request, but we should detox the following folders for Linux users:
- device-types
- module-types

Having the right naming convention is essential when you call API and those filename under Ansible automation for example.
As netbox is a Linux based solution, we should try to avoid special characters like & and space preferably.

This is a proposal to detox those 2 folders (detox -r . used)

Thank you.

Regards,


jeremystretch:
> Having the right naming convention is essential when you call API and those filename under Ansible automation for example.

Is there a reason you can't just quote or escape the file path? I'm also a Linux user and this has never been an issue for me.

jsenecal:
I'm with @jeremystretch could you please provide a breaking use-case ?

requenag:
Hi,

Understand the quoting mechanisms but let's say the user is not quote this
(by mistake) and someone a tiny silly and add a {command &} as part of the
filename.

My constructive feedback I am trying to make here is don't we want to be
secure and compliant and make al the chance on our side to make Netbox
usage safe?

-
https://www.cyberciti.biz/faq/linuxunix-rules-for-naming-file-and-directory-names/
- https://tldp.org/LDP/abs/html/special-chars.html

danner26:
I would like to add to what the other maintainers have said, and first note that I use NetBox (and Linux) on a daily basis. This has never proven to be an issue for myself or my colleagues (including those that are new to Linux). This change would break downstream dependencies (namely the popular NetBox Python import, even though it is not inherently supported here, we attempt to keep things compatible). My personal opinion is that we should not proceed with this, but I will leave the final decision to @jeremystretch

jeremystretch:
I don't see any compelling justification for the proposed change. As mentioned, it's trivial to properly handle these file paths properly, and the onus falls to the individual user to do so. Thanks for the suggestion but we're going to decline.

PR #1105, opened 2023-02-07
Add Zyxel VMG1312-B10A

madecto:


danner26:
Hello, all looks good. Do you have `weight` information? If so, please populate that. If not, just let me know. Thanks!

PR #1108, opened 2023-02-07
Adding Intel NIC cards.

requenag:
Hi,

This is a pull request to add Intel NIC card as modules for server compute.

Regards,


danner26:
Hello, all looks good except for the comments. Can you please make those markdown links? The format `[text](url)` will do this. You will just need to wrap that in single quotes (`'`)

Thanks!

PR #1110, opened 2023-02-08
Add Catalyst WS-C2960XR-24TD-I

squintfox:


danner26:
@squintfox please add `weight` if you can find it, and also please link the datasheet in the `comments`. Thank you!

squintfox:
@danner26 Updated as requested

danner26:
Are the PSUs hot swappable? I don't have this exact unit in my lab so I cannot really check. If they are hot swappable they should be defined as module bays rather than PSUs themselves

squintfox:
@danner26 Yup, you're right.  I updated it as modular.

PR #1111, opened 2023-02-08
Add Fortinet FEX, FG-VM and FS

squintfox:
Added new FortiExtender, FortiGate and FortiSwitch models.


danner26:
Please add `fortinet-` as a prefix to the slug on each device. Also, please add the `weight`, `airflow`, and include a link to the datasheet within the `comments` if you can. Let me know once that is done and I will re-review. Thank you!

squintfox:
@danner26 Updated as requested.

danner26:
Hey, a few more things. First, I am sorry I didnt notice these earlier. 
The `FG-VM-OPC` is a virtual machine SKU, an thus is not a physical device. This does not belong in this library. I have removed the VM file for you.
The `FEX-40D`, you have `FX04DA` which appears to be a ROM revision. This repository is not meant for this. If you want to track rom revisions I would suggest you add an inventory item. I have fixed this in the other devices as well. The best way to know which is the correct name is to look at the physical switch, it usually has the model name printed on it. 
image

The `FS-1048E` and `FS-448D-FPOE` appears to have hot swappable AC power supplies. These should be modelled as module bays and then you should add the module-type definitions if those files arent already created

The `FS-448D-POE`, does the RPS connector on the switch itself come pre-installed? Or do you have to purchase a module for the switch to use it? If it is preinstalled, then it is fine. Otherwise please make this a module bay

danner26:
If I incorrectly put any datasheets, please feel free to replace them. Try to make sure they are from the manufacturer and not random websites though. Thanks!

squintfox:
Hey @danner26 ,

1) If that's what's been decided, you should probably purge FG-VM-AZURE as well.

2) Fortinet sold the different models with different radios for different geos.  They are not ROM revisions.  You can see the models here.  https://www.avfirewalls.com/datasheets/FortiExtender.pdf  I don't really agree that the marketing material should take precedence to engineering information about how the hardware models identify themselves.  That's what part numbers are, marketing/sales material.

4) The connector is built in.  The RPS is a separate box you have to buy to supply DC power to the connector.  I don't think I'd consider it a module.

I only included the sub-par links because Fortinet does not seem to maintain the old ones on their website anymore (for the FEX or the switches).  I agree that it's not ideal, but that's all I could find.

In any case, this has deviated from my work pretty significantly so I'm not going to do a back and forth thing.  If it's how you want it, totally keep it, if not, go ahead and close it out.

danner26:
1. Absolutely agree, if you could please remove it as part of this PR
2. THe original values are not the actual model names. They corelated to the ROM revision according to one of the fortinet hosted PDFs (forgot to link it earlier). The model names/part numbers I see are `FortiExtender 20D` with a part number of `FEX-20D`, `FortiExtender 100B ` with a part number of `FEX-100B`, `FortiExtender 40D-NAM` with a part number of `FEX-40D-NAM`, `FortiExtender 40D-INTL` with a part number of `FEX-40D-INTL`, and `FortiExtender 40D-AMEU` with a part number of `FEX-40D-AMEU`. This information is directly presented on the datasheet you provides and the ones on the Fortinet website. If you can show me a document with this numbers, I will chat internally about how to handle this otherwise. 
3. Great, then you can ignore the module statement. It defintiely isnt a module

In regards to the datasheets, I see what you mean. You can omit any non-manufacturer datasheets from the comment then.

squintfox:
I'm going to somewhat disagree here.  Fortinet makes 3 physical boxes that they write "FortiExtender 40D" on the front.  Under the hood, they are 3 different physical things with 3 different cellular modems.  https://www.avfirewalls.com/datasheets/FortiExtender.pdf, page 4 illustrates this.

That said, I deal with Fortinet hardware very often and in large quantity.  In their software (regardless of marketing literature) they identify each of their devices by a 6 character string (no hyphens, no underscores).  There is no document that I know of that shows all of this information to the public.  The existing devicetype-library is a hodgepodge of different solutions:

model: FortiGate 81F
part_number: FGT81F

model: FortiGate 80E
part_number: FG-80E

model: FortiGate 40F
part_number: FGT_40F

In these cases FGT81F has some basis in reality.  That's how the device identifies itself and it's serial number follows this pattern.  The other two, as far as I know, are just made up.  I'm sort of fine with whatever, but there really isn't a correct standard here.  I think the 6 character code should be somewhere in some field so NetBox can match to the device's internal identification.  Where that goes, I'm not sure.  In my environment, I solved it with model (I guess I forgot that when I went back and uploaded these types), but that may not be appropriate for here.

danner26:
The issue that I am trying to tackle these past few weeks is standardizing this repository, since like you said there is a hodgepodge of different structures going on. That being said, it is extremely difficult to just believe "these numbers belong" without either having the device to see or having reference in the datasheet. Can you show me a terminal output where it lists that information as either a model or sku/product number? This issue is even further exasperated by the fact that on page 5 of the very document you linked states the Product name, for instance, is `FortiExtender 40D-INTL` with a SKU of `FEX-40D-INTL`

For instance in your original commit you had `FX04DA` which is nowhere to be found on this datasheet. What I did find was this [configuration guide](https://fortinetweb.s3.amazonaws.com/docs.fortinet.com/v2/attachments/3e5dfde3-8eec-11e9-81a4-00505692583a/FortiExtender-4.1.1-Admin-Guide.pdf) which mentions explicitly `ROM REV : FX04DA` (which points to my original statement). In this guide they also clearly prepend this value to the devices hostname/id and serial number. There is even a command which sets the local device ID `set localid FX04DA5918004527` during a VPN configuration. Neither hostname, serial number or device ID are explicitly model names or skus/product numbers but rather that value is a device specific one (i.e. a value you enter after creating the actual device in NetBox, not the template).

Unless you are saying that the datasheet provided by fortinet is lying, and that the product name and SKU are incorrect, then I dont think we are seeing eye to eye here. The goal of this repository is to have the exact data that is represented by the manufacturer. Again, if you need custom data like the `FX04DA` one, you can add those in NetBox as a custom field or an inventory item if that value would ever change.

jeremystretch:
Judging from the linked FortiExtender.pdf document, it seems to identify five discrete device types:

* FortiExtender 20D (FEX-20D)
* FortiExtender 40D-NAM (FEX-40D-NAM)
* FortiExtender 40D-INTL (FEX-40D-INTL)
* FortiExtender 40D-AMEU (FEX-40D-AMEU)
* FortiExtender 100B (FEX-100B)

These are the model names and part numbers that should be defined in the library, as each identifies a unique, atomic hardware configuration. Similarly, [this data sheet](https://www.fortinet.com/content/dam/fortinet/assets/data-sheets/FortiGate_80E_Series.pdf) shows the model names and numbers for the FortiGate 80 series:

* FortiGate 80E-PoE (FG-80E-POE)
* FortiGate 81E-PoE (FG-81E-POE)

This is the pattern we should follow for all Fortinet products, regardless of existing definitions that may not be accurate. As @kkthxbye-code suggests, where it becomes necessary to attach some supplementary identifier to accurately correlate live devices to their device type, this should be done using a custom field or some other mechanism.

squintfox:
@danner26 I think this should be good now.  I removed the non-vendor links and updated the modular PSUs.

danner26:
Thank you @squintfox for working through this with us!

PR #1112, opened 2023-02-08
Add APC PDUs, Smart-UPS and XLBP

squintfox:
Add Rack PDU models, Smart-UPS models and an XLBP extended battery pack.


danner26:
This is a lot to go through, give me some time and I will work on it. Please ensure you have all device information including `weight`, `airflow`, `u_height` (set to 0 if not rack mountable), and a `comments` to the datasheet if available. Thanks!

squintfox:
@danner26 So I built these before weight and airflow existed in NetBox (or at least they weren't commonplace in the devicetypes-library at that time).  We don't really care about those data points in our organization, so I'm probably not going to go all the way back through 30+ units at this time.  u_height should be good for all of them.  If they aren't useful as is, I'd say close this one out.  I would have kept links too, but at that time comments had a size limit and I kept running into it being too long.

danner26:
```
weight: 12.01
weight_unit: lb
```

PR #1113, opened 2023-02-08
Update APC NMC module ports

squintfox:
Updated missing ports on APC NMC modules

### DIFF
--- a/module-types/APC/AP9630.yaml
+++ b/module-types/APC/AP9630.yaml
@@ -4,9 +4,9 @@ model: AP9630
 part_number: AP9630
 comments: UPS Network Management Card 2
 console-ports:
-  - name: {module} Serial (3.5mm)
+  - name: Serial (3.5mm) [{module}]
     type: other
 interfaces:
-  - name: {module} Network
+  - name: Network [{module}]
     type: 100base-tx
     mgmt_only: true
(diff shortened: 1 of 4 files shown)


danner26:
Generally modules will use the module number in the name, to avoid collisions with on board interfaces. You should probably modify the interfaces and console names so that this doesnt occur

squintfox:
@danner26 I think I understood what you asked for?  Let me know if I didn't get it right.

danner26:
Not quite, sorry about that. If you take a look at [the C9300-NM-4G](https://github.com/netbox-community/devicetype-library/blob/master/module-types/Cisco/C9300-NM-4G.yaml) we have `{module}` in the name. Once this module is loaded into a devices module bay, lets call that module bay `1` in this case (but it could be anything such as `M1`, `MOD1`, etc), the interfaces associated with the module will get automatically added to the device called `GigabitEthernet1/1/1`, `GigabitEthernet1/1/2` and so forth

If the module bay is named `MOD1` and we loaded the C9300-NM-4G, you would see interfaces called `GigabitEthernet1/MOD1/1`, etc.

What this does is prevents clashing of non-unique interface/port names.

squintfox:
@danner26 OK, I think I understand.  Should be good now.

danner26:
That should work!

PR #1114, opened 2023-02-08
Update fortinet firewalls

squintfox:
Added missing power ports, updated incorrect power ports, added new PoE fields, updated comments to descriptions.


danner26:
For the FG-100E and the FG-200E lets name the RPS connector something better, that name is pretty long. Maybe something like `RPS` would do

Otherwise everything else looks good

squintfox:
@danner26 Updated as requested.  I went with a shortened version of what's physically written on the device.

PR #1116, opened 2023-02-08
Update meraki ports

squintfox:
Updated missing and incorrect ports for Meraki switches.

### DIFF
--- a/device-types/Cisco/Meraki-MS225-48LP.yaml
+++ b/device-types/Cisco/Meraki-MS225-48LP.yaml
@@ -11,7 +11,7 @@ power-ports:
     type: iec-60320-c14
     maximum_draw: 490
     allocated_draw: 53
-  - name: 22-Pin Redundant Power Supply Connector
+  - name: RPS Connector
     type: dc-terminal
     maximum_draw: 490
     allocated_draw: 53
--- a/device-types/Cisco/Meraki-MS225-48LP.yaml
+++ b/device-types/Cisco/Meraki-MS225-48LP.yaml
@@ -11,7 +11,7 @@ power-ports:
     type: iec-60320-c14
     maximum_draw: 490
     allocated_draw: 53
-  - name: Cisco RPS 2300 connector for PSU failover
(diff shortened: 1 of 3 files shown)


device-types/Cisco/Meraki-MS225-48LP.yaml
```diff
@@ -5,11 +5,16 @@ slug: ms225-48lp
 part_number: MS225-48LP-HW
 u_height: 1
 is_full_depth: false
+comments: '[MS225 Overview and Specifications](https://meraki.cisco.com/lib/pdf/meraki_datasheet_ms225.pdf)'
 power-ports:
   - name: Slot 1
     type: iec-60320-c14
     maximum_draw: 490
     allocated_draw: 53
+  - name: Cisco RPS 2300 connector for PSU failover
```
danner26, line 14: The name on this should be either what appears in  the config or printed on the device.. this name is a little too long to make sense

danner26 (changes requested):

danner26:
Otherwise everything looks good, thanks!

device-types/Cisco/Meraki-MS225-48LP.yaml
squintfox: @danner26 I updated the name as to what's written in the documentation.  Unfortunately it's not a whole lot shorter.  It has no physical label on the device.

danner26: Why dont we call it something like `RPScon` then, its shorter and to the point that way

danner26:
In regards to the `MS425-32`, I see that they are actually modular power supplies. These should be modelled as module-bays and then each specific PSU should be a module-type

device-types/Cisco/Meraki-MS225-48LP.yaml
squintfox: Updated as "RPS Connector" which has some basis in technical documentation and is more brief that than the other options.

squintfox:
@danner26 Updated as modular PSUs.

danner26 (approved):

PR #1118, opened 2023-02-09
Add albis-elcon BIG 2862

m2martin:


danner26:
Doing a quick search I was not able to find the details we needed, so I will leave that up to you.
In regards to the big-2862, please:

- modify the slug to be `albis-elcon_big-2862`
- add `weight` if you have it
- add the `part_number` which appears to be `900489`
- and add a `comments` with a link to the modules pdf if possible (I wasnt able to find much public info on these devices though)

In regards to the WANX-4VDSL.yaml, I did find this [document](https://www.albis-elcon.com/wp-content/uploads/2017/06/Data-sheet-BIG-2862-en-ae.pdf). I will need you to confirm or deny this, but in that PDF it appears that the `WANX-4VDSL` is actually `WANX-4VDSL-Extender` with a part number of `900421`. 
If this is the case, please update the module type accordingly. **It might also be beneficial for you to add the rest of the products on that page as well.**
In direct regards to the WANX-4VDSL-Extender, please

- add `weight` if you have it
- Add a `comments` with a link to the modules pdf if possible (I wasnt able to find much public info on these devices though)

Thank you!

m2martin:
Sorry for the delay.
I can change the slug but I'm also unable to provide any specs because they are undisclosed by the vendor. I can also not confirm the part numbers - I could for the devices to which I'm having physical access but it is not obvious if the part number is distinct to these ones. So, IMO leaving the part number empty is more accurate than writing an uncertain one.

If this is not what you want then please close the PR as I connot provide the infos. Then I will stay with my templates.

danner26:
Hello @m2martin in that case, please do change the slugs. Please omit the PNs you do not have physical access to/cannot confirm and also omit the spec sheet since you cant provide them. Anything that you do have access to you can add

danner26:
@ me when you are ready for re-review

danner26 (approved):

module-types/albis-elcon/WANX-4VDSL.yaml
```diff
@@ -0,0 +1,8 @@
+---
+manufacturer: albis-elcon
+model: WANX-4VDSL
```
danner26, line 3: According to the PDF you provided, it appears that part number `900421` is not `model: WANX-4VDSL`, it is actually `model: WANX-4VDSL-Extender`. Please update the model and file name accordingly

module-types/albis-elcon/WANX-4SHDSL.yaml
```diff
@@ -0,0 +1,8 @@
+---
+manufacturer: albis-elcon
+model: WANX-4SHDSL
```
danner26, line 3: According to the PDF you provided, it appears that part number `900420` is not `model: WANX-4SHDSL`, it is actually `model: WANX-4SHDSL-Extender`. Please update the model and file name accordingly

danner26 (changes requested):

module-types/albis-elcon/WANX-4VDSL.yaml
m2martin: It's described as being an extender in the datasheet but the model is printed without `-Extender` on the label.

module-types/albis-elcon/WANX-4SHDSL.yaml
m2martin: It's described as being an extender in the datasheet but the model is printed without `-Extender` on the label.

module-types/albis-elcon/WANX-4VDSL.yaml
danner26: Could you upload a picture of the module that shows the model on it?

danner26: This has nothing to do with believing or not believing you, it is the fact that the datasheet explicitly lists the model as `WANX-4VDSL-Extender`. I also cannot read the language that PDF is in, so if you could tell me a page to look on or send me an English version I would be happy to validate. I simply asked you to upload a picture of the physical device with the model number to show the exact wording printed on it. 

While I completely understand the sentiment you are conveying in your last sentence, we **do not** want inaccurate information for the sake of being fast. Fast is not always efficient or correct. The entire point of NetBox and this Device/Module Library is to act as a source of truth.

m2martin:
Weight property change commited.

PR #1119, opened 2023-02-09
Add Adtran NetVanta 4660

m2martin:


m2martin:
Check completion depends on #1120

danner26:
Hello, #1128  was merged and I have updated your branch. Please also add the devices `part_number`, `airflow` if you can find it, `weight` if you can find it, and `comments` with the datasheet if you have it. Thank you!

m2martin:
Sorry for the dalay.
Do you really want to have the datasheet link in the template?
To be honest, I personally avoid having any external links in my templates because the links may change and become unreachable. If I would use the template from here, I would delete the comment.
It has never been a requirement.

danner26:
Hello, yes the datasheet should be linked in a comment field. For example, `comments: '[HPE FlexFabric 5945 2-slot Switch Data sheet](https://www.hpe.com/psnow/doc/PSN1011424706USEN.pdf)'`. This is if you have it/it is publicly accessible that is

While it isnt a hard requirement, if it is available it is good to have. Often the links will remain while the device is in support, and afterwards that becomes less of an issue. They are for reference

`part_number`, `airflow`, and `weight` are also good to have but not required

m2martin:
I've added the weight info and part no.
Unfortunately, I've only found datasheets on 3rd party sites but not from the vendor's site. I think the product is not being sold anymore.

danner26 (approved):

PR #1124, opened 2023-02-09
Add N77-SUP2E

ascension2020:


danner26:
Please add the datasheet as a comment, I tried to myself but you seem to have the "Allow edits from maintainers" disabled

danner26:
```
weight: 12.01
weight_unit: lb
```

Due to this change, **you will be required to update your branch/PR from the new master branch**. This might require you to update your PR, if you had weight currently defined, so that your PR is in-line with the new weight schema. If you have any issues, please let me know or refer to [this PR](https://github.com/netbox-community/devicetype-library/pull/1187) for more information.

module-types/Cisco/N77-SUP2E.yaml
```diff
@@ -0,0 +1,11 @@
+---
+manufacturer: Cisco
+model: N77-SUP2E
+part_number: N77-SUP2E
+interfaces:
+  - name: mgmt0
```
DanSheps, line 6: This will cause a problem when inserting it into a chassis where there is two mgmt0 interfaces.  I believe it actually throws an exception.

DanSheps (changes requested):

module-types/Cisco/N77-SUP2E.yaml
scoffman77: I added the following to my documents to address this issue.
interfaces:
  - name: Sup{module} mgmt0
    type: 1000base-t
    mgmt_only: true
console-ports:
  - name: Sup{module} Con0
    type: rj-45

danner26: @scoffman77 The change needs to be made on the module, not the upstream

PR #1127, opened 2023-02-10
added juniper acx7100

uthapa82:
Added Juniper ACX7100-48L Cloud Metro Router. Had to create it for work in order to add a new devices. Didn't find it in device-types. Just added a single file ACX7100-48L


danner26:
So according to the [hardware link](https://www.juniper.net/documentation/us/en/hardware/acx7100-48L/topics/topic-map/acx7100-48L-system-overview.html) you mentioned, there are 4 actual models of this device. `ACX7100-48L-AC-AI`, `ACX7100-48L-DC-AI`, `ACX7100-48L-AC-AO`, and `ACX7100-48L-DC-AO`. These also have modular power supplied and fans. All of the power supplies and fans should be created as module-types, and the base models should be created with `module-bay`s for the fan/psu. If you are able to take care of creating these great, otherwise let me know and I can do so for you

uthapa82:
Thank you for your review, fixed the revision suggestions

danner26:
I think you should remove the `ACX7100-48L.yml` since that isnt the actual device, the actual device specification is one of the 4 listed previously.
Do you want to add any of the modules as well before we merge this (i.e. the fan and psu modules)?

uthapa82:
Hi, already added fan and psu modules.

Thanks

PR #1131, opened 2023-02-14
Create S6730-H28Y4C.yaml

grbeneke:
Add Huawei CloudEngine S6730-H28Y4C


danner26:
@grbeneke by default, out of the box are these actually/configured `25gbase-x-sfpp` or `25gbase-x-sfp28`?

grbeneke:
@danner26 that was a typo. Second commit corrects the error.

danner26:
It looks like the power supplies are modular? If this is the case, please make the psu slots module bays instead, and then model the PSUs themselves as module types, if they havent already been. If you have any questions let me know!

grbeneke:
@danner26 template has been changed to module-bays for power

PR #1138, opened 2023-02-19
Create Huawei CE8800 series

grbeneke:


danner26:
**device-types/Huawei/CE8850-64CQ-EI.yaml**
Quick question in regards to the port names, are these how they are in the config file (i.e. `100GE1/0/64`)? If so, then perfect. Otherwise please try to make these match the full names in the configuration, just as we do for the Cisco ports (i.e. `GigabitEthernet1/0/1`).
Please add `airflow` and `weight` if you have it
Please add a comment with the devices datasheet in markdown format
The PSUs on this device appear to be modular, as do the fans. Please see [here](https://support.huawei.com/enterprise/en/doc/EDOC1000019246/56d573f4/ce8850-64cq-ei). These should be created as `module-bays` and the `power-ports` definition should be removed

**device-types/Huawei/CE8861-4C-EI.yaml**
Is this a parent device (i.e. other devices get put into this chassis)? If so, please add `subdevice_role: parent` since this appears to be a parent chassis device
Please add `airflow` and `weight` if you have it
Please add a comment with the devices datasheet in markdown format
The PSUs and Fans appear to be modular here as well. Please see [this doc](https://support.huawei.com/enterprise/en/doc/EDOC1000019246/56e787dc/ce8861-4c-ei). These should be created as `module-bays` and the `power-ports` definition should be removed

**module-types/Huawei/CE88-D24S2CQ.yaml**
Quick question in regards to the port names, are these how they are in the config file (i.e. `25GE 1/{module}/1` or if the module id is `1` then `25GE 1/1/1`)? If so, then perfect. Otherwise please try to make these match the full names in the configuration, just as we do for the Cisco ports (i.e. `GigabitEthernet1/0/1`).
Please add a comment with the devices datasheet in markdown format

**module-types/Huawei/CE88-D8CQ.yaml**
Quick question in regards to the port names, are these how they are in the config file (i.e. `100GE 1/{module}/1` or if the module id is `1` then `100GE 1/1/1`)? If so, then perfect. Otherwise please try to make these match the full names in the configuration, just as we do for the Cisco ports (i.e. `GigabitEthernet1/0/1`).
Please add a comment with the devices datasheet in markdown format

grbeneke:
Hi @danner26,
- Unnecessary spaces have been removed in interface names, otherwise naming is correct.
- Weight and Comments added.
- Airflow dictated by inserted modules and not by the chassis.
- PWR and FAN module-bays added.

device-types/Huawei/CE8861-4C-EI.yaml
```diff
@@ -0,0 +1,26 @@
+---
+manufacturer: Huawei
+model: CE8861-4C-EI
```
danner26, line 3: Is this a parent device (i.e. other devices get put into this chassis)? If so, please add `subdevice_role: parent` since this appears to be a parent chassis device

grbeneke: The device does not contain sub-devices, only modules. No device-bays so no requirement for `subdevice_role`

PR #1141, opened 2023-02-21
Add QNAP 453B

TDarkShadow:
Some customers still use this legacy model.

### DIFF
--- a/device-types/QNAP/TS-453B.yml
+++ b/device-types/QNAP/TS-453B.yml
@@ -1,16 +1,11 @@
 ---
 manufacturer: QNAP
 model: TS-453B
-slug: ts-453B
+slug: qnap-ts-453B
+u_height: 0
 part_number: TS-453B
 airflow: side-to-rear
 is_full_depth: false
-subdevice_role: parent
-device-bays:
-  - name: Disk 1
-  - name: Disk 2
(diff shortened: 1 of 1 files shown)


danner26:
Please add `qnap-` as a prefix to your slug
Device bays are not meant to be used for disks. We are in the process of figuring out storage management, but in the meantime please remove this.
If you want to track disks, we suggest you use inventory items. These do not belong in this repo though.
Please set `u_height: 0`
Please remove `subdevice_role`
Otherwise everything looks good

TDarkShadow:
Incorporated your feedback, thanks!

I do want to note I followed the template of [QNAP TS-453mini](https://github.com/netbox-community/devicetype-library/blob/50925b63512c79a3a1ee7635391664cf5cba0336/device-types/QNAP/TS-453mini.yml) so might be necessary to change all those models as well?

Should there also be noted somewhere this device is a NAS?

danner26:
Hey @TDarkShadow so that is going to be true. We have come a long way and added a bunch of features over time.. unfortunately already defined devices fall behind the curve when this happens. They definitely will need to be updated, and if you want to take that on I would really appreciate that.
That being said, everything else looks good. Please feel free to open another PR if you want to tackle the other changes (I might take a stab at it as well)

PR #1143, opened 2023-02-24
Add Ubiquiti UISP Router

nomaster:


danner26:
Hello @nomaster thanks for the contribution! I added airflow and weight definitions for you as well. Once the test succeed I will merge this/fix any issues I caused.

PR #1152, opened 2023-02-28
added Aruba-2930M-48G: JL321A and JL322A

francesco-noc:
Can I remove commented lines?


danner26 (approved):

danner26:
> Can I remove commented lines?

Absolutely. These shouldnt be in those files. Only the devices defaults should be listed, even if you can configure them otherwise.. basically we stride to achieve the exact config that comes when you first take the machine out of the box

PR #1161, opened 2023-03-01
Added FS S3150-8T2FP

jine:


device-types/FS/S3150-8T2FP.yaml
```diff
@@ -0,0 +1,40 @@
+---
+manufacturer: FS
+model: S3150-8T2FP
+slug: s3150-8t2fp
```
danner26, line 4: Please append `fs-` to the slug

device-types/FS/S3150-8T2FP.yaml
```diff
@@ -0,0 +1,40 @@
...
+part_number: 'S3150-8T2FP (#90130)'
```
danner26, line 5: The FS part number is `S3150-8T2FP`, please remove the `(#90130)` part

device-types/FS/S3150-8T2FP.yaml
```diff
@@ -0,0 +1,40 @@
...
+is_full_depth: false
+u_height: 1
+airflow: left-to-right
+comments: 'S3150-8T2FP, 8-Port Gigabit Ethernet L2+ PoE+ Switch, 8 x PoE+ Ports @ 130W, with 2 x 1Gb SFP, Fanless'
+console-ports:
```
danner26, line 10: Please add weight if you have it, according to the datasheet this would be
```
weight:
  - value: 3.7
    unit: lb
```

danner26, line 10: Please add `comments: '[FS S3150 Series Switches Data Sheet](https://www.fs.com/products_support/preview.html?categories_id=3256&files_id=4543&model_name=S3150-8T2FP&is_datasheet=1)'`

device-types/FS/S3150-8T2FP.yaml
```diff
@@ -0,0 +1,40 @@
...
+  - name: Console
+    type: rj-45
+    label: Console
+    description: Serial Console Port
+power-ports:
+  - name: Power Port 1
```
danner26, line 16: In the configuration, is the PSU referred to as `Power Port 1`? If so then this is fine. Otherwise please make this match what the configuration/device refers to it as. 

Just a thought, if the PSU is not defined anywhere in the config/device cli maybe use `PSU1` since that is a bit more standard for naming.

danner26 (changes requested):
Please review my requested changes/questions. Ping me once you are ready for re-review

device-types/FS/S3150-8T2FP.yaml
jine: It's not mentioned more than as "Power Supply (AC)" - so I'll rename it to PSU1 for standard purposes.

danner26 (approved):

danner26:
Thank you for your contribution! Please feel free to add more if you would like :)

PR #1163, opened 2023-03-01
Update N9K-C9372TX.yaml

bryanward-net:
This model has 10G Copper ports, not 10G SFP ports.


danner26:
Thanks for the correction!

PR #1177, opened 2023-03-06
Intel h2312xxlr2

getspaghett:
Created definitions for Intel H2312XXLR2 blade system.
Specifications taken from [Intel.com](https://www.intel.com/content/www/us/en/products/sku/87573/intel-server-chassis-h2312xxlr2/specifications.html)

### DIFF
--- a/device-types/Intel/H2312XXLR2.yaml
+++ b/device-types/Intel/H2312XXLR2.yaml
@@ -1,0 +1,32 @@
+---
+manufacturer: Intel
+model: H2312XXLR2
+slug: intel-h2312xxlr2
+u_height: 2
+is_full_depth: true
+comments: |
+  [H2312XXLR2 Specifications on intel.com](https://www.intel.com/content/www/us/en/products/sku/87573/intel-server-chassis-h2312xxlr2/specifications.html)
+subdevice_role: parent
+device-bays:
+  - name: Node 1
+    label: Blade 1
+  - name: Node 2
+    label: Blade 2
+  - name: Node 3
+    label: Blade 3
+  - name: Node 4
+    label: Blade 4
+power-ports:
+  - name: Power 1
+    label: 2130W common redundant power supply (Platinum Efficiency)
+    type: FXX2130PCRPS
+    maximum_draw: 2130
+  - name: Power 2
+    label: 2130W common redundant power supply (Platinum Efficiency)
+    type: FXX2130PCRPS
+    maximum_draw: 2130
+inventory-items:
+  - name: FHW12X35HS12G
+    label: 12Gb SAS backplane
+    manufacturer: Intel
+    part_id: FHW12X35HS12G
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 .DS_Store
 venv/
 .vscode
+venv*/


getspaghett:
Sorry, i think i have changed this from a draft to a pull request too soon.  :(

getspaghett:
found out how to change it back to draft

getspaghett:
i am sorry but i do not have enough git skills yet to remove my .gitignore from this branch ...

getspaghett:
ok i think i figured it out, the gitignore is no longer in this branch ... sorry for the messy commits

danner26:
Its great seeing you move through the issues and overcome them :) I will work on reviewing this soon. Thanks in advance for your contribution!

device-types/Intel/H2312XXLR2.yaml
```diff
@@ -0,0 +1,32 @@
...
+device-bays:
+  - name: Node 1
+    label: Blade 1
+  - name: Node 2
+    label: Blade 2
+  - name: Node 3
+    label: Blade 3
+  - name: Node 4
+    label: Blade 4
+power-ports:
```
danner26, line 19: It looks like the power and nic bays (dunno what theyre actually called in Intel land) are actually modular, so these should be removed and created under `module-bays`. Then you should add both the nic bays and the PSUs to the module-types folder

According to the datasheet, the PSUs are `FXX2130PCRPS`

device-types/Intel/H2312XXLR2.yaml
```diff
@@ -0,0 +1,32 @@
+---
+manufacturer: Intel
+model: H2312XXLR2
+slug: intel-h2312xxlr2
+u_height: 2
+is_full_depth: true
+comments: |
+  [H2312XXLR2 Specifications on intel.com](https://www.intel.com/content/www/us/en/products/sku/87573/intel-server-chassis-h2312xxlr2/specifications.html)
+subdevice_role: parent
```
danner26, line 9: Please add `airflow` if you have it

danner26, line 9: Please add `weight` for the empty chassis if you have it

danner26 (changes requested):

device-types/Intel/H2312XXLR2.yaml
danner26: Oh wait, looking at your module commit it appears that the modules have the nics on them, so do you put the devices in from the back?? Sorry this is weird compared to the VRTX's I am accustomed to. If those "nic bays" are actually the devices themselves, please let me know and mark this as resolved

danner26, line 19: It looks like the PSUs are actually modular, so these should be removed from this file and created under `module-bays`. Then you should add the PSUs themselves to the module-types folder

According to the datasheet, the PSUs are `FXX2130PCRPS`

getspaghett: Yes, the compute blades are put in from the back and they have the network interfaces on them. I had to dig a little deeper on the intel support site and found this neat [Chassis Quick Reference](https://www.intel.com/content/dam/support/us/en/documents/motherboards/server/chassis/h2000/sb/jflabel1.pdf). :)

device-types/Intel/H2312XXLR2.yaml
danner26: Thank you for adding this, please remove the comment you inserted in the file near this area. In the future just respond within the change request :) After removing the comment you can resolve this conversation as well

danner26 (changes requested):

danner26:
Do the PSU modules already exist? If so then just the one last correction, otherwise please add the applicable PSU modules to this PR :)

device-types/Intel/H2312XXLR2.yaml
getspaghett: I removed the comment and will add the psu modules to this PR during the next hour. thank you very much for the help :)

getspaghett: I have added the power supplies as module-types. I am not sure if i created them correctly but i will rework them asap if i know what to do :+1:

danner26:
```
weight: 12.01
weight_unit: lb
```

device-types/Intel/H2312XXLR2.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - name: Node 2
+    label: Compute Node 2
+  - name: Node 3
+    label: Compute Node 3
+  - name: Node 4
+    label: Compute Node 4
+module-bays:
+  - name: PSU Slot 1
+    label: Power Supply 1
+    position: lower center rear
```
danner26, line 27: Please change this to `1`, so that when you call `{module}` in the module-type it will be called `PSU1`

module-types/Intel/FXX1600PCRPS.yaml
```diff
@@ -0,0 +1,8 @@
+---
+manufacturer: Intel
+model: FXX1600PCRPS
+power_ports:
+  - name: PSU
```
danner26, line 5: Please change this to `- name: PSU{module}`
What happens here is that `{module}` in the name, when added to a module-bay will use the `position` definition in place of this. For instance, if the module-bay defined the position as `position: 1`, then when attached this will get called `PSU1`

module-types/Intel/FXX2130PCRPS.yaml
```diff
@@ -0,0 +1,8 @@
+---
+manufacturer: Intel
+model: FXX2130PCRPS
+power_ports:
+  - name: PSU
```
danner26, line 5: Please change this to `- name: PSU{module}`
What happens here is that `{module}` in the name, when added to a module-bay will use the `position` definition in place of this. For instance, if the module-bay defined the position as `position: 1`, then when attached this will get called `PSU1`

device-types/Intel/H2312XXLR2.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - name: PSU Slot 2
+    label: Power Supply 2
+    position: upper center rear
```
danner26, line 30: Please change this to `1`, so that when you call `{module}` in the module-type it will be called `PSU 1`

danner26 (changes requested):

danner26 (approved):

getspaghett:
Thank you for the fixes :)

PR #1178, opened 2023-03-07
Added Intel HNS2600TPR Blade

getspaghett:
I created the .yaml file for the Intel blade with the Specs from the [Intel Spec sheet](https://ark.intel.com/content/www/us/en/ark/products/88297/intel-compute-module-hns2600tpr.html). I thought about adding the PCIe slots as module bays but i was not sure if this is the intended way to model this kind of data inside NetBox. Please let me know how i can improve my contributions.

### DIFF
--- a/device-types/Intel/HNS2600TPR.yaml
+++ b/device-types/Intel/HNS2600TPR.yaml
@@ -1,0 +1,23 @@
+---
+manufacturer: Intel
+model: HNS2600TPR
+slug: intel-hns2600tpr
+u_height: 0.5
+comments: |
+  [HNS2600TPR Specifications on intel.com](https://ark.intel.com/content/www/us/en/ark/products/88297/intel-compute-module-hns2600tpr.html)
+subdevice_role: child
+inventory-items:
+  - name: S2600TPR
+    label: Intel Server Board S2600TPR
+    manufacturer: Intel
+interfaces:
+  - name: GigabitEthernet1
+    type: 1000base-t
+  - name: GigabitEthernet2
+    type: 1000base-t
+  - name: IPMI
+    type: 1000base-t
+    mgmt_only: true
+console-ports:
+  - name: Rear VGA
+    type: other


danner26:
So for the PCIe slots, can they be used for NIC cards? If so, then yes they are 100% valid as module-bays

That being said, NetBox does not _yet_ support storage management, but it is on the roadmap.

device-types/Intel/HNS2600TPR.yaml
```diff
@@ -0,0 +1,23 @@
+---
+manufacturer: Intel
+model: HNS2600TPR
+slug: intel-hns2600tpr
+u_height: 0.5
+comments: |
+  [HNS2600TPR Specifications on intel.com](https://ark.intel.com/content/www/us/en/ark/products/88297/intel-compute-module-hns2600tpr.html)
+subdevice_role: child
```
danner26, line 8: Please add `weight` for the device if you have it

danner26, line 8: Please add `airflow` for the device if you have it

device-types/Intel/HNS2600TPR.yaml
```diff
@@ -0,0 +1,23 @@
...
+interfaces:
+  - name: GigabitEthernet1
+    type: 1000base-t
+  - name: GigabitEthernet2
+    type: 1000base-t
+  - name: IPMI
+    type: 1000base-t
+    mgmt_only: true
+console-ports:
+  - name: Rear VGA
```
danner26, line 22: This is not what this is intended for.. This is intended for console ports only. Please remove the VGA port

danner26 (changes requested):

danner26:
```
weight: 12.01
weight_unit: lb
```

danner26:
Does this PR need anything else?

getspaghett:
> Does this PR need anything else?
No, i think i added everything that is attached to the chassis and it is ready to be merged :+1:

PR #1179, opened 2023-03-07
Add FS S5860-20SQ

baldoarturo:
Add FS S5860-20SQ to device library
[https://www.fs.com/products/108710.html](https://www.fs.com/products/108710.html)


device-types/FS/S5860-20SQ.yaml
```diff
@@ -0,0 +1,73 @@
+---
+manufacturer: FS
+model: S5860-20SQ
+slug: s5860-20sq
```
danner26, line 4: please append `fs-` to the slug

device-types/FS/S5860-20SQ.yaml
```diff
@@ -0,0 +1,73 @@
...
+part_number: s5860-20sq
+u_height: 1
+is_full_depth: true
+airflow: front-to-rear
+console-ports:
```
danner26, line 9: Please add a comment with a link to the datasheet in markdown format if you can find a publicly available one from the manufacturer
I believe [this is it](https://www.fs.com/products_support/preview.html?categories_id=3258&files_id=1263&model_name=S5860-48SC&is_datasheet=1)

danner26, line 9: Please add `weight` information if you have it, according to the datasheet above it would be:
```
weight:
  - value: 10.14
    unit: lb
```

device-types/FS/S5860-20SQ.yaml
```diff
@@ -0,0 +1,73 @@
...
+  - name: Management
+    type: rj-45
+power-ports:
```
danner26, line 12: It looks like the PSUs are modular. These should be modelled as `module-bays` and then you should add the applicable PSUs for the device as module-types if they are not already created

It looks like the smart fans are also modular, so please create them in the same fashion

device-types/FS/S5860-20SQ.yaml
```diff
@@ -0,0 +1,73 @@
...
+  - name: Slot 1
+    type: iec-60320-c14
+    maximum_draw: 85
+    allocated_draw: 85
+  - name: Slot 2
+    type: iec-60320-c14
+    maximum_draw: 85
+    allocated_draw: 85
+interfaces:
+  - name: Te1/1
```
danner26, line 22: I do not work with these devices so I am just confirming, in the configuration are these referenced as `Te1/1` or do they have a full name such as in Cisco IOS devices `GigabitEthernet1/0/1`

danner26: If they are actually `Te1/1` feel free let me know and resolve this

danner26 (changes requested):

device-types/FS/S5860-20SQ.yaml
baldoarturo: Thanks Daniel. From what I see this should apply to the other FS switches too, I would be happy to fix them.

I guess this should cover it.
image
image

```
module-bays:
  - name: PSU1
    label: PSU1
    position: PSU1
  - name: PSU2
    label: PSU2
    position: PSU2
  - name: FAN1
    label: FAN1
    position: FAN1
  - name: FAN2
    label: FAN2
    position: FAN2
```

device-types/FS/S5860-20SQ.yaml
baldoarturo: My bad, just checked on an unit and in fact they are _fullinterfacename_ _unit/slot/port_

baldoarturo: Fixed on #57fe0051cb2bad4312d590ce8a4d569e7e321d19

device-types/FS/S5860-20SQ.yaml
baldoarturo: Fixed on #cb670f548df0b69e618c51b3b2c9341d8c1d228c

baldoarturo: Fixed

device-types/FS/S5860-20SQ.yaml
baldoarturo: Fixed

device-types/FS/S5860-20SQ.yaml
danner26: Absolutely, these should apply to the other devices as well. A lot of these were created before this was available, or they were just mistakenly created that way, so we have to update them as we get PRs like these.

If you would be willing to update the others that would be a **HUGE** help! I would majorly appreciate it!

baldoarturo: Sure, it helps me too. I fixed this for this model and added the specific PSU on module-types at module-types/FS/SPSM-C150WAC.yaml
LMK if that should be the proper way and I'll cover the rest soon.

baldoarturo: Ah interesting, makes sense to me. I just fixed it.

danner26:
```
weight: 12.01
weight_unit: lb
```

danner26 (approved):
Looks good now! Thank you for your contribution

baldoarturo:
Thank you for your help, I'll take care for modules and PSU the rest of switches from this vendor

PR #1190, opened 2023-03-11
Add Ubiquiti INS-3AF-O-G PoE adapter

BarbarossaTM:


danner26:
Hello @BarbarossaTM I do not think this is the correct way to model this. Instead, you should have 1 interface with PoE Mode set to Power Delivery `PD`, and the second interface with PoE Mode set to Power Supply `PSE`

BarbarossaTM:
Hm, I can see why you would want to model it this way. For me there are two reasons for rather choosing Front/Rear Port:

a) In my understanding an Interface terminates an L2 connection, which doesn't happen here, L2+-wise this thing is totally transparent, arguebly even L1-wise.
b) I want to be able to trace from the switch to the connected WiFi backhaul gear / see it directly on the switches interfaces page / the API, which I loose if these are interfaces.

What do you think?

danner26:
Hmm I see all of your points, let me pose this to the maintainers and lets see what the collective thinks

BarbarossaTM:
Hey @danner26, did the collective come up with a verdict? :-D

danner26:
We are torn on it, pretty much 50/50. So as a last ditch effort, just to cover all of my bases, I am going to pose the question to the community.. If I dont reach back out in a few days please ping me again! I am sorry for the delay

danner26:
[This discussion](https://github.com/netbox-community/devicetype-library/discussions/1235) includes the poll

danner26:
Alright we are going with what you have! The majority has decided. Thanks for waiting while we figured that out

PR #1210, opened 2023-03-17
Add front/rear images for Nokia devices

jbemmel:
All images are 226x19 PNG, and can be imported using the enhanced tool (separate PR)


danner26:
Hey @jbemmel, rather than keeping the images in the same folder as the definitions lets create a new folder for all images. 

I already have an implementation ready. I will get that merged to master soon so you can utilize it

danner26:
Hey @jbemmel is this still needed/still being worked on?

PR #1217, opened 2023-03-22
Add Cisco ASR-9010 Device Type

eboxd:


asr-9010.yaml
```diff
@@ -0,0 +1,45 @@
...
+  maximum_draw: 2100
+- name: 0/PT1-PM1
+  type: DC Terminal
+  maximum_draw: 2100
+- name: 0/PT1-PM2
+  type: DC Terminal
+  maximum_draw: 2100
+- name: 0/PT1-PM3
+  type: DC Terminal
+  maximum_draw: 2100
```
DanSheps, line 33: These should be modelled as removalble modules with the modules included (Ex: A9K-DC-PEM-V2=) in the module definitions

asr-9010.yaml
```diff
@@ -0,0 +1,45 @@
...
+- name: Slot 1 RSP 1
+- name: Slot 2 Line Card 0
+- name: Slot 3 Line Card 1
+- name: Slot 4 Line Card 2
+- name: Slot 5 Line Card 3
+- name: Slot 6 Line Card 4
+- name: Slot 7 Line Card 5
+- name: Slot 8 Line Card 6
+- name: Slot 9 Line Card 7
+- name: Slot 10 Line Card 8
```
DanSheps, line 45: These should be modelled as module bays and not devices bays.

Please include a selection of line-cards and Sups as well if they don't already exist.

DanSheps (changes requested):

PR #1218, opened 2023-03-22
Added Extreme Network 5520-12MW-36W FabricEngine device and modules

WagoL:
Added 5520-12MW-36W FabricEngine device type, 5520-VIM-4YE and XN-ACPWR-1100W-FB.

### DIFF
--- a/device-types/Extreme Networks/5520-12MW-36W-FabricEngine.yaml
+++ b/device-types/Extreme Networks/5520-12MW-36W-FabricEngine.yaml
@@ -1,0 +1,265 @@
+---
+manufacturer: Extreme Networks
+model: 5520-12MW-36W-FabricEngine
+slug: 5520-12mw-36w-fabricengine
+part_number: 5520-12MW-36W-FabricEngine
+u_height: 1.0
+is_full_depth: true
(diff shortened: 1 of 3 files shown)


danner26:
Hey @WagoL for the `module-types/Extreme Networks/5520-VIM-4YE-FabricEngine.yaml` any chance that 2/1 could already be taken on the device this would be added to? Just like you did in the PSU, using {module} in the name allows for the interfaces to be added based on the position for the module bay

WagoL:
> Hey @WagoL for the `module-types/Extreme Networks/5520-VIM-4YE-FabricEngine.yaml` any chance that 2/1 could already be taken on the device this would be added to? Just like you did in the PSU, using {module} in the name allows for the interfaces to be added based on the position for the module bay

@danner26, it is not possible there is only one module bay in the devices it will go into. 
If needed I can change it.

danner26:
Good to know, would this module ever get added to a different devices that maybe has 2 bays, which could cause the error?

WagoL:
This module can only go into one specific device.

danner26 (approved):

PR #1219, opened 2023-03-23
Add module EX-UM-2x4SFP.yaml

powerburner-nl:
Module for Juniper switch


danner26:
Is there ever a chance for the device which this is being slotted into having a `ge-0/1/0` already?

Generally with modules you define the ports with a name something like `ge-{module}/1/0` or similar. This allows the module to be imported dynamically based on the parent devices module-bay `position`

If you have any questions just let me know

powerburner-nl:
I am 99% certain there is no chance of that happening. This module is used in the EX3200/EX4200 series. They only have only 1 expansion port for a module. Since the switch has 1 plan which starts with ge-0/0/x, the module will always be ge-0/1/x. 

You can find information about it here: [](https://www.juniper.net/documentation/us/en/software/junos/interfaces-ethernet-switches/topics/topic-map/switches-interface-understanding.html#id-understanding-interface-naming-conventions__d95184e1011)

PR #1221, opened 2023-03-23
lower case slug

francesco-noc:
If slug is not specified as field, Netbox ansible module require lowercase slug when only "type" is given


danner26:
Hey @francesco-noc, all slugs should be lower case.. Thank you for fixing these. I will add a test case to ensure that the slugs are lower case.

danner26 (approved):

PR #1232, opened 2023-03-27
Added module bays

MrPaulAR:


danner26:
@MrPaulAR Please ensure all tests pass

device-types/Juniper/MX960.yaml
```diff
@@ -1,7 +1,32 @@
 ---
 manufacturer: Juniper
-model: MX960
+model: MX960xx
```
DanSheps, line 3: This doesn't look correct.

What is the xx for?  What does the inventory show on the actual CLI for a model number?

MrPaulAR:
I had an old version of this file in my local repo so you are correct in your note regarding the model number.  

Just need to add module bays.

PR #1234, opened 2023-03-28
Add UPS APC SMC1500i

davidusken:
Add UPS APC SMC1500i


danner26:
@davidusken the images need to be in the elevation-images folder. Also it is best practice to make sure you start on the master branch, sync that master branch with the upstream, then create a new branch where you start making your changes

PR #1256, opened 2023-04-01
replace '/' by os.path.sep

aurelienbyi:
replace / by os.path.sep to be windows compliant


danner26:
Hey @aurelienbyi I agree this is a better implementation, I just need to do my due-diligence so give me some time on that. That being said, I need you to please change this PR to merge into the develop branch, not the master branch

PR #1260, opened 2023-04-02
Add/update Lenovo ThinkSystem

Madrov:
Added: ThinkSystem Power Supply

Update: Lenovo ThinkSystem SR server uses module-bays instead of power-ports, the slug has been corrected to include the manufacturer. I believe this is a best practice.


Madrov:
Hello, not sure why I got 'Error: Process completed with exit code 1.' during the validation steps?

PR #1261, opened 2023-04-03
modules JH180A and JH181A

francesco-noc:
Added two HPE modules:

- (JH180A) HPE 5930 24-port SFP+ and 2-port QSFP+ Module
- (JH181A) HPE 5930 24-port SFP+ and 2-port QSFP+ with MACsec Module

### DIFF
--- a/module-types/HPE/HPE-JH180A.yaml
+++ b/module-types/HPE/HPE-JH180A.yaml
@@ -56,6 +56,3 @@ interfaces:
     type: 40gbase-x-qsfpp
   - name: 'FortyGigE1/{module}/26'
     type: 40gbase-x-qsfpp
-
-
-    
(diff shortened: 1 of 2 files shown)


francesco-noc:
I'm not able to understand why it fails

arcticash:
Hi Francesco, 
I'm not associated with this project but I've just had a quick look at your commits and I think its the interface names being in quotes which is causing the tests to fail.
If you can get the pre-commit function to run locally this will check and try to correct any issues before committing, but I think if you remove the single quotes around the interface names then this should pass the automated checks.

Madrov:
I just fixed my PR by running the "pre-commit run --all" command on my PR #1260 and indeed that works, I tested on your files and the pre-commit removed the quote on the interface like @arcticash said. 

For example:
  - name: Ten-GigabitEthernet1/{module}/1 
instead of
  - name: 'Ten-GigabitEthernet1/{module}/1'

Somehow I could not run my venv on windows (Run PyTest case Checks step never end..). Everything is fine on ubuntu desktop... odd.

danner26:
Thanks for jumping in and assisting @arcticash 
@francesco-noc I was away last week but I am moving through the PRs again now

PR #1269, opened 2023-04-11
Add Switch Zyxel GS1900-8HP

davidusken:


danner26:
Hey @davidusken in previous additions you said the ports were referred to as `1`, `2`, etc. Here you have the ports as `Port 1`, `Port 2`, etc. Which is it?

Lets try to keep things exactly as they are in the devices configuration, and make sure we are consistent going forward. Thank you!

davidusken:
Thanks for noticing! Fixed now, logged into the switch and checked that the correct naming is indeed 1, 2 etc.

danner26 (approved):

PR #1285, opened 2023-04-13
PAN-PA-5400-DPC-A.yaml

kalazzawi:
Module types are available for this chassis. I will create pull requests for them as well.
PAN-PA-5400-DPC-A
PAN-PA-5400-MPC-A
PAN-PA-5400-NC-A-1
PAN-PA-5400-NC-A-2


danner26:
Does this module not have any interfaces or console ports or power ports or anything? I dont use these personally so I have no idea myself

kalazzawi:
This module has no interfaces.  It is a data processing card.  We can have up to 4 total in slots 3-6.  Have this card listed is good for inventory management.

danner26 (approved):

PR #1287, opened 2023-04-13
PAN-PA-5400-NC-A.yaml

kalazzawi:
first network card


danner26:
Generally when creating modules, we define the interface/port name using `{module}` in it, this utilizes the position attribute on module bays within device types. When adding the module to a device, this definition would prevent possible name collisions.

For instance, if a device has an interface named `1/0/1` on-board, and a module bay with a position attribute of `1/1`, then using a module that has an interface name `{module}/1`, after plugging it into the device the interface names would evaluate to `1/1/1` when connected in NetBox.
---
Is there any chance that the device this module gets plugged into already has an adapter named `ethernet1/1`?

kalazzawi:
I like the {module} variable idea.  We should then just create one network card (NC) named PAN-PA-5400-NC-A.

Networking Cards (NCs) provide network connectivity for a PA-5450 firewall. To scale performance and capacity, you can install up to two NCs in a PA-5450 firewall (in slots 1 and 2 only).  Slots 4-6 are data processing card slots and slot 7 is the management card slot.  In referring back to the network cards, the port numbering designation is ethernet, followed by slot/port such as ethernet<slot>/<port>, where slot is the physical slot the card is installed in and the port is the interface port number. For example, the first Ethernet port on an NC installed in slot 1 shows ethernet1/1 and port 2 shows ethernet1/2. The first port on an NC installed in slot 2 shows ethernet2/1 and port 2 shows ethernet2/2. For information on installing the NC, see Install a PA-5400 Series Firewall Networking Card (NC).

kalazzawi:
Please change the file name to PAN-PA-5400-NC-A.yaml.

kalazzawi:
OK.  I saw where I could also change the file name.  I made the changes to the file name.  We should be good at this point.

kalazzawi:
Please review at your earliest convenience.

danner26 (approved):

PR #1294, opened 2023-04-14
Add QNAP TS-EC1280U

Madrov:


device-types/QNAP/TS-EC1280U.yaml
```diff
@@ -0,0 +1,43 @@
...
+slug: qnap-ts-ec1280u
+part_number: TS-EC1280U
+u_height: 2
+airflow: front-to-rear
+is_full_depth: true
+weight: 14.18
+weight_unit: kg
+subdevice_role: parent
+comments: '[TS-EC1280U product page](https://www.qnap.com/en/product/ts-ec1280u)'
+device-bays:
```
danner26, line 13: Currently NetBox is not utilized for storage management. We have that feature in planning, but it is not currently ready. Please remove the disk definitions and `device-bays`

device-types/QNAP/TS-EC1280U.yaml
```diff
@@ -0,0 +1,43 @@
...
+manufacturer: QNAP
+model: TS-EC1280U
+slug: qnap-ts-ec1280u
+part_number: TS-EC1280U
+u_height: 2
+airflow: front-to-rear
+is_full_depth: true
+weight: 14.18
+weight_unit: kg
+subdevice_role: parent
```
danner26, line 11: This is generally used for chassis which house other devices, this is not meant for devices with modules. Is this actually needed, and if so why are there no `device-bays`?

danner26 (changes requested):

Madrov:
I made the adjustments. Thank you 🙂

danner26 (approved):

PR #1296, opened 2023-04-14
Add Aruba 3810M, Procurve 6108 and E2510-24G

Madrov:
Add module-type HPE X371 as well (PSU)

### DIFF
--- a/module-types/HPE/Aruba-JL085A.yaml
+++ b/module-types/HPE/Aruba-JL085A.yaml
@@ -1,0 +1,9 @@
+---
+manufacturer: HPE
+model: JL085A
+part_number: JL085A
+comments: Aruba X371 250W AC Power Supply
+power-ports:
+  - name: Power {module}
+    type: iec-60320-c14
+    maximum_draw: 250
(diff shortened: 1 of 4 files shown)


module-types/HPE/Aruba-JL085A.yaml
```diff
@@ -0,0 +1,9 @@
+---
+manufacturer: HPE
+model: JL085A
+part_number: JL085A
+comments: Aruba X371 250W AC Power Supply
+power-ports:
+  - name: Power {module}
```
danner26, line 7: In the config, does this device actually call it `Power 1`, `Power 2`, etc? If not, please adjust this to the name matches what it does in the device config/output

danner26 (changes requested):

danner26:
Please take a look at the question above

module-types/HPE/Aruba-JL085A.yaml
Madrov: You are right, the output is as followed:
I will adjust to `PS {module}` 🙂 Thank you!

 PS#   Model     Serial      State           AC/DC  + V        Wattage   Max
 ----- --------- ----------- --------------- ----------------- --------- ------
 1     JL085A    CN04******  Powered         AC 120V/240V        44       250
 2                           Not Present     -- ---------         0         0

danner26 (approved):

PR #1298, opened 2023-04-14
Extreme networks vsp 7254xsq

WagoL:
added Extreme networks vsp-7254xsq device type. 
added PSU module for in the vsp-7254xsq.


module-types/Extreme Networks/EC72051AF-E6.yaml
```diff
@@ -0,0 +1,11 @@
+---
+manufacturer: Extreme Networks
+model: EC72051AF-E6
```
danner26, line 3: I am having a hard time finding `EC72051AF-E6`, but I have found `EC7205A1F-E6`

Is the proper model name `EC7205A1F-E6`?

device-types/Extreme Networks/VSP-7254XSQ.yaml
```diff
@@ -0,0 +1,131 @@
...
+manufacturer: Extreme Networks
+model: VSP-7254XSQ
+slug: extreme-networks-vsp-7254xsq
+part_number: VSP-7254XSQ
+u_height: 1
+is_full_depth: true
+weight: 7250
+weight_unit: g
+airflow: front-to-rear
+subdevice_role: parent
```
danner26, line 11: This is generally used for chassis which house other devices, this is not meant for devices with modules. Is this actually needed, and if so why are there no `device-bays`?

danner26 (changes requested):

device-types/Extreme Networks/VSP-7254XSQ.yaml
WagoL: removed the parent role.

module-types/Extreme Networks/EC72051AF-E6.yaml
WagoL: This is indeed a typo. Made the change.

PR #1304, opened 2023-04-19
Add server Dell PowerEdge R240

davidusken:
https://www.dell.com/en-us/shop/enterprise-products/r240-1ru-server-intel/spd/poweredge-r240


device-types/Dell/PowerEdge_R240.yaml
```diff
@@ -0,0 +1,26 @@
...
+is_full_depth: true
+airflow: front-to-rear
+weight: 12.2
+weight_unit: kg
+console-ports:
+  - name: Serial
+    type: de-9
+  - name: VGA
+    type: other
+power-ports:
```
danner26, line 15: Is this PSU modular? If so, this should be modelled in a `module-bays`, and then the applicable PSU should be added to the module-types folder

device-types/Dell/PowerEdge_R240.yaml
```diff
@@ -0,0 +1,26 @@
...
+slug: dell_poweredge_r240
+u_height: 1
+is_full_depth: true
+airflow: front-to-rear
+weight: 12.2
+weight_unit: kg
+console-ports:
+  - name: Serial
+    type: de-9
+  - name: VGA
```
danner26, line 13: VGA is not a console port, please remove this

danner26 (changes requested):

davidusken:
Ok thanks, Dell EMC PowerEdge R240 comes with 2 different options for PSUs. It is changeable but requires opening the cover, removing a screw, and disconnecting the cables. Here is a screenshot of the PSU specifications:

image

Would you prefer this to be added as modules still or have 2 different model versions of the server?

danner26:
Do the devices have different part numbers based on which PSU is installed?

PR #1305, opened 2023-04-19
Add server HPE Proliant DL 160 Gen9

davidusken:


device-types/HPE/ProLiant-DL160-Gen9.yaml
```diff
@@ -0,0 +1,27 @@
...
+slug: hpe-proliant-dl160-g9
+u_height: 1
+is_full_depth: true
+airflow: front-to-rear
+console-ports:
+  - name: Serial
+    type: usb-a
+  - name: VGA
+    type: other
+power-ports:
```
danner26, line 13: It looks like the power on these are modular. Thus, these should be in `module-bays` and the applicable PSUs should be added to the module-types folder if they dont already exist

device-types/HPE/ProLiant-DL160-Gen9.yaml
```diff
@@ -0,0 +1,27 @@
...
+manufacturer: HPE
+model: ProLiant DL160 Gen9
+slug: hpe-proliant-dl160-g9
+u_height: 1
+is_full_depth: true
+airflow: front-to-rear
+console-ports:
+  - name: Serial
+    type: usb-a
+  - name: VGA
```
danner26, line 11: This is not what this is meant for, please remove VGA

danner26 (changes requested):

PR #1307, opened 2023-04-19
Add server Supermicro SYS-5018D-MTF

davidusken:


device-types/Supermicro/SYS-5018D-MTF.yaml
```diff
@@ -0,0 +1,26 @@
+---
+manufacturer: Supermicro
+model: SYS-5018D-MTF
```
danner26, line 3: The model appears to actually be `SuperServer 5018D-MTF` according to their website

device-types/Supermicro/SYS-5018D-MTF.yaml
```diff
@@ -0,0 +1,26 @@
...
+slug: sys-5018d-mtf
```
danner26, line 4: slug should have the manufacturer appended like so `supermicro-`

danner26 (changes requested):

danner26 (approved):

PR #1311, opened 2023-04-20
Create ProCurve-2510G-48.yaml

Wlanfr3ak:
Added the HPE ProCurve-2510G-48


danner26:
Hey @Wlanfr3ak for interfaces 45-48, which type do they default to? This file should be written with all defaults, meaning config out of the box, in mind. I just want to clarify that before we merge it, Thanks!

Wlanfr3ak:
I didnt have access to the switch configuration but in the manual is written: "Each of the four mini-GBIC slots is shared with the associated 10/100/1000Base-T RJ-45 port. If a mini-GBIC is installed in a slot, the associated RJ45 port is disabled." So i think standard is both but when both connected the mini-GBIC is only the active. Source: https://gzhls.at/blob/ldb/6/b/7/0/ac880a560404223aa672906fb58461f78da4.pdf Page: 33

danner26 (approved):

PR #1314, opened 2023-04-26
Move FS-248E-FPOE.yaml file inside Fortinet dir

dainok:
I think the FS-248E-FPOE.yaml file should be placed inside the Fortinet dir, not on devicetype-library root.


danner26 (approved):
Absolutely, thanks for catching this

PR #1316, opened 2023-04-26
Add SR850.yaml

muratbulat:
Reference: https://lenovopress.lenovo.com/lp1052.pdf

### DIFF
--- a/device-types/Lenovo/SR850.yaml
+++ b/device-types/Lenovo/SR850.yaml
@@ -1,7 +1,7 @@
 ---
 manufacturer: Lenovo
 model: ThinkSystem SR850
-slug: SR850
+slug: sr850
 part_number: 7B27A03953
 u_height: 2
 is_full_depth: true
--- a/device-types/Lenovo/SR850.yaml
+++ b/device-types/Lenovo/SR850.yaml
@@ -1,0 +1,18 @@
+---
+manufacturer: Lenovo
+model: ThinkSystem SR850
+slug: SR850
+part_number: 7B27A03953
+u_height: 2
+is_full_depth: true
+comments: '[Lenovo System SR850 product guide](https://lenovopress.lenovo.com/lp1052.pdf)'
+subdevice_role: parent
+interfaces:
+  - name: XCC
+    type: 1000base-t
+    mgmt_only: true
+power-ports:
+  - name: PSU0
+    type: iec-60320-c14
+  - name: PSU1
+    type: iec-60320-c14


muratbulat (commented):
ThinkSystem SR850

muratbulat (commented):
lower

device-types/Lenovo/SR850.yaml
```diff
@@ -0,0 +1,18 @@
...
+part_number: 7B27A03953
+u_height: 2
+is_full_depth: true
+comments: '[Lenovo System SR850 product guide](https://lenovopress.lenovo.com/lp1052.pdf)'
+subdevice_role: parent
+interfaces:
+  - name: XCC
+    type: 1000base-t
+    mgmt_only: true
+power-ports:
```
danner26, line 14: According to the datasheet, the PSUs are hot-swappable. So the power slots should be created in `module-bays`, and then the associated PSU's themselves should be added to the module-types folder if they do not already exist

device-types/Lenovo/SR850.yaml
```diff
@@ -0,0 +1,18 @@
...
+manufacturer: Lenovo
+model: ThinkSystem SR850
+slug: sr850
+part_number: 7B27A03953
+u_height: 2
+is_full_depth: true
+comments: '[Lenovo System SR850 product guide](https://lenovopress.lenovo.com/lp1052.pdf)'
+subdevice_role: parent
+interfaces:
+  - name: XCC
```
danner26, line 11: Is there only one interface on this device?

device-types/Lenovo/SR850.yaml
```diff
@@ -0,0 +1,18 @@
+---
+manufacturer: Lenovo
+model: ThinkSystem SR850
+slug: sr850
+part_number: 7B27A03953
+u_height: 2
+is_full_depth: true
+comments: '[Lenovo System SR850 product guide](https://lenovopress.lenovo.com/lp1052.pdf)'
+subdevice_role: parent
```
danner26, line 9: `subdevice_role: parent` should only be used when there are children devices, which would be housed under `device-bays`

An example of this is the Dell VRTX chassis. The chassis itself is a device, which is a parent to blades such as the M620.

Unless this device houses other devices, which I cannot find information saying it does, then please remove this definition.

device-types/Lenovo/SR850.yaml
```diff
@@ -0,0 +1,18 @@
+---
+manufacturer: Lenovo
+model: ThinkSystem SR850
+slug: sr850
```
danner26, line 4: The slug for this should be `lenovo-thinksystem-sr850`

device-types/Lenovo/SR850.yaml
```diff
@@ -0,0 +1,18 @@
...
+part_number: 7B27A03953
+u_height: 2
+is_full_depth: true
+comments: '[Lenovo System SR850 product guide](https://lenovopress.lenovo.com/lp1052.pdf)'
```
danner26, line 8: Please add `weight` and `airflow` definitions

device-types/Lenovo/SR850.yaml
```diff
@@ -0,0 +1,18 @@
+---
+manufacturer: Lenovo
+model: ThinkSystem SR850
+slug: sr850
+part_number: 7B27A03953
```
danner26, line 5: This part number appears to only be for the processor and memory expansion tray. This is not the part number for the device itself.

danner26 (changes requested):

PR #1336, opened 2023-05-09
added 2650 and updated 2626

francesco-noc:


francesco-noc:
I found that both procurve 2626 and 2650 have three part numbers[0]:
J4900A-J4900B-J4900C ProCurve Switch 2626
J4899A-J4899B-J4899C ProCurve Switch 2650

How can we deal with this?
Can I add commented "part_number" line?

[0] https://support.hpe.com/hpesc/public/docDisplay?docId=emr_na-c02593383

francesco-noc:
Or can i create three files for each model?
es.
ProCurve-2626-A.yaml, ProCurve-2626-B.yaml, ProCurve-2626-C.yaml
or 
ProCurve-2626-J4900A.yaml, ProCurve-2626-J4900B.yaml, ProCurve-2626-J4900C.yaml

danner26:
I am not at my computer yet to check, but generally with these cases one of the SKUs will be just the device without add-ons. We suggest that you model this device as the single part itself, and then model the rest of the components as module-types

We don't currently have a better way to track this natively, and thus the base sku with the defaults is the preferred way to model the data in this repository

francesco-noc:
I'm not sure of having done what you suggest but where i work we have maintainance contract associated to part number even if the model name is the same, so we need to specify in our NetBox instance the part number reported in the device.

danner26:
Hello, so please correct me if I am wrong. There is a base model with one part number that does not include add-on modules. Then there are part numbers that include added modules as well but they all slot into the same base device

Is this the case?

danner26:
Hello, today we finished migrating all of the files and slugs to their new proper naming convention as described in the README. It looks like this PR has a merge conflict. I do not have access to modify this branch, so please resolve it by using the slug/model name defined in the master branch and then push it to this PR.

francesco-noc:
> Hello, so please correct me if I am wrong. There is a base model with one part number that does not include add-on modules. Then there are part numbers that include added modules as well but they all slot into the same base device
> 
> Is this the case?

Hello, that's not what you wrote. 

Both devices (2626 and 2650) have no modules, neither added nor included.
J4900A, J4900B, J4900C (as for J4899A, J4899B, J4899C) are one device but produced in different years.
Therefore, due to the maintenance contract bundled with devices, it is necessary to keep track of the correct part number.

I have seen how the slug/model name has been defined in the master branch but to correctly report the differences, I think, either I can add the part number to the slug, as I did, or the last letter as suffix to model.

es: 
slug: hpe-procurve-2626-j4900a 
or
model: Procurve 2626 A
slug: hpe-procurve-2626-a 

What do you think? Other ideas?

danner26:
I understand, your comment just confused me a bit. Unfortunately none of these work with the current standard we have been following. HPE has proven to be the most difficul to standardize because they do things like this.. Anyway, is the port count the same between all 3 models? Is the only difference in them the year they were manufactured?

francesco-noc:
> Anyway, is the port count the same between all 3 models?
Yes

> Is the only difference in them the year they were manufactured?
Yes

danner26:
Well that is annoying.. Do you happen to have a link to the HPE page for one of these devices?

francesco-noc:
https://support.hpe.com/hpesc/public/docDisplay?docLocale=en_US&docId=c01804526  with last part number (JxxxxC)
https://support.hpe.com/hpesc/public/docDisplay?docId=emr_na-c01815266   with technical specification

danner26:
Please either modify this PR to allow maintainers to make edits, or fix the conflict please.

danner26:
Very weird, im not sure. It wasnt letting me make resolutions to the conflict for some reason. It looks like you took care of it though

danner26 (approved):

PR #1340, opened 2023-05-12
added hp 4200 series

francesco-noc:


danner26:
@francesco-noc is this another case of same model/same hardware different part number based on release date?

francesco-noc:
Yes.

https://techlibrary.hpe.com/docs/products/eos/4200vlseriesabrollannouncement_nov_08_ww_eng.pdf

PR #1364, opened 2023-05-17
Update Cisco ASA 5500 series

williamhatcher:
I noticed several Cisco ASA 5500 Series boxes had the wrong `is_full_depth` value. I verified with [Cisco's website](https://www.cisco.com/c/en/us/products/collateral/security/asa-firepower-services/datasheet-c78-742475.html) and updated the values


danner26 (approved):

danner26:
Thank you for this!

PR #1367, opened 2023-05-19
fix juniper mic-3d-16xge-sfpp

gpipito:
fixed model value in the yaml file


danner26:
Hey @gpipito it looks like there is a conflict. Please resolve that and let me know when this is ready to be reviewed. Thanks!

device-types/Juniper/PTX10001-36MR.yaml
```diff
@@ -0,0 +1,88 @@
+---
+manufacturer: Juniper
+model: PTX10001-36MR
+slug: ptx10001-36mr
```
danner26, line 4: Please append the manufacturer name to the device like so: `juniper-`

device-types/Juniper/PTX10001-36MR.yaml
```diff
@@ -0,0 +1,88 @@
...
+interfaces:
```
danner26, line 5: You are missing a ton of information to make this a complete and valid PR. Please add the following data points: 

- `part_number`
- `u_height`
- `is_full_depth`
- `airflow`
- `weight` and `weight_unit`
- `comments` with a makrdown formatted link to the devices datasheet (on the manufacturer website, if available)

device-types/Juniper/PTX10001-36MR.yaml
```diff
@@ -0,0 +1,88 @@
...
+  - name: re0:mgmt-0
+    type: 1000base-t
+    mgmt_only: true
+  - name: et-0/0/0
```
danner26, line 9: Please confirm for me these are the full interface names. For instance, on a Cisco 2960 you will see `GigabitEthernet1/0/1` but you can shorthand it to `gi1/0/1`. We opt for the most complete information so we expect the interface to be named exactly as it is shown.

Can you please confirm that the devices configuration shows `et-0/0/0`?

device-types/Juniper/PTX10001-36MR.yaml
```diff
@@ -0,0 +1,88 @@
...
+    type: 100gbase-x-qsfp28
+  - name: et-0/2/11
+    type: 100gbase-x-qsfp28
+power-ports:
+  - name: PSM0
+    type: iec-60320-c20
+    maximum_draw: 3000
+  - name: PSM1
+    type: iec-60320-c20
+    maximum_draw: 3000
```
danner26, line 85: It appears this devices PSUs are modular. Thus, you should have these defined under `module-bays` and the applicable PSU modules should be added to the module-types folder if they do not already exist
It also appears this device has 6 modular fans. Those should be added to `module-bays` as well

danner26 (changes requested):
Please review and fix requested changes

device-types/Juniper/PTX10001-36MR.yaml
gpipito: sorry, I don't get it. Where am I supposed to write "juniper-"?
The other juniper in the repo are reported like this:

---
manufacturer: Juniper
model: QFX5110-48S-4C
slug: qfx5110-48s-4c
interfaces:
...

---
manufacturer: Juniper
model: MX480
slug: mx480
u_height: 8
is_full_depth: true
...

Consider that I used the QFX file as a template for my ptx.

device-types/Juniper/PTX10001-36MR.yaml
gpipito: ok...but again a lot of device in the repo are missing that information too. I'm just copy/pasting.

device-types/Juniper/PTX10001-36MR.yaml
gpipito: that is Juniper naming convention:

image

device-types/Juniper/PTX10001-36MR.yaml
gpipito: I can remove the power-ports information. Not a problem.
Again however, a lot of juniper device do not report that level of detail.

device-types/Juniper/PTX10001-36MR.yaml
danner26: That means they havent been migrated yet, one moment ill take care of that now

device-types/Juniper/PTX10001-36MR.yaml
gpipito: if you look at MX104-BASE or MX104-PREMIUM for example, you'll see they report PEM modules as built-in power-ports.

device-types/Juniper/PTX10001-36MR.yaml
danner26: I understand, new PRs require the full set of information to be added. A lot of these files are years old before these data points existed in NetBox. All new PRs require as much information as possible

device-types/Juniper/PTX10001-36MR.yaml
danner26: Perfect! Thank you

device-types/Juniper/PTX10001-36MR.yaml
danner26: @gpipito please run a git pull and take a look at the other Juniper devices now. You should see the new slug naming convention

device-types/Juniper/PTX10001-36MR.yaml
gpipito: ok. I'm going to find that information. However the chassis does not have a part number:

image

gpipito:
I deleted PTX and PSM modules. Only the correction on mic model is left.

danner26 (approved):

danner26:
Please feel free to open another PR for any other devices if you would like

PR #1376, opened 2023-05-22
Add QNAP NAS TS-1685

detvan:


device-types/QNAP/TS-1685.yml
```diff
@@ -0,0 +1,38 @@
+---
+manufacturer: QNAP
+model: TS-1685-D1521-16G-550W
```
danner26, line 3: The model name of this looks to be `TS-1658`. The part number is `TS-1685-D1521-16G-550W`

Please adjust accordingly

danner26 (changes requested):

device-types/QNAP/TS-1685.yml
danner26: @detvan besides the file name, also please modify the model definition to `TS-1658`

danner26 (approved):

PR #1381, opened 2023-05-23
Added yaml for APIC-M1

dryhen:


device-types/Cisco/APIC-M1.yaml
```diff
@@ -0,0 +1,31 @@
...
+console-ports:
+  - name: Con
+    type: rj-45
+power-ports:
+  - name: PS-0
+    type: iec-60320-c14
+    maximum_draw: 770
+  - name: PS-1
+    type: iec-60320-c14
+    maximum_draw: 770
```
danner26, line 17: Are the PSU's modular? If so, then they should be defined under `module-bays` and the associated PSUs should be created in the module-types folder if they do not already exist

mhendry: They are modular but I am unsure of the exact information to used to create the module types. I created this based off of the existing APIC-M2 and M3 files which had the power supplies the same.

device-types/Cisco/APIC-M1.yaml
```diff
@@ -0,0 +1,31 @@
+---
+manufacturer: Cisco
+model: APIC-M1
+slug: cisco-apic-m1
+part_number: APIC-M1
+u_height: 1
+is_full_depth: true
```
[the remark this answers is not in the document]
mhendry: Cannot find this datapoint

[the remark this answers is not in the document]
mhendry: Cannot find these datapoints

device-types/Cisco/APIC-M1.yaml
danner26: Hello @mhendry I understand, all new commits need to follow the proper modelling based on current NetBox data. Thus this will need to be modified. If you can search around you might be able to find the info on Google or something

PR #1396, opened 2023-05-27
adding HP t740 thin client (used as router)

SimonFischer04:
Adding device-type for the HP t740 thin client (used as a router by me).

Inspired by: https://github.com/netbox-community/devicetype-library/blob/master/device-types/Intel/NUC11TNHi5.yaml

### DIFF
--- a/device-types/HP/t740.yaml
+++ b/device-types/HP/t740.yaml
@@ -1,0 +1,32 @@
+---
+manufacturer: HP
+model: t740
+slug: hp-t740
+u_height: 0.0
+is_full_depth: false
+airflow: right-to-left
+comments: '[Datasheet](https://www8.hp.com/h20195/v2/GetDocument.aspx?docname=4aa7-5975enuc)'
+weight: 2.93
+weight_unit: lb
+power-ports:
+  - name: PSU1
+    type: dc-terminal
+    maximum_draw: 90
+    allocated_draw: 65
+interfaces:
+  - name: eth0
+    type: 1000base-t
+    description: Realtek RTL8111HSH GbE
+  - name: wl0
+    type: ieee802.11ac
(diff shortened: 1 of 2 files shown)


danner26:
Hello, the manufacturer should be HPE. Please follow standards already defined in this repository. You can refer to the HPE folder if you want to see some examples but be wary because some of the files might be slighly outdated from current available data points.

SimonFischer04:
I put them into HP intentionally, infact because
I actually looked at the hp example before and did some research.

Sorry if I don't understand this correctly / overlooked something, but:
Looking at this again for example (just a random recently updated one): 
https://github.com/netbox-community/devicetype-library/blob/master/device-types/HPE/HP-3100-24-v2-EI.yaml 
while the file is called "hp-..." the linked support page spec clearly states it as an **HPE** device.
**HP** and **HPE** are two different companies and this **t740 is a HP** device, so I thought I had to create another folder.

But as requested, changed it now anyways.

danner26:
Hello, so a lot of those devices are pre-exisitng before we migrated them to a new schema. 

HPE (Hewlett Packard Enterprise) is the manufacturer, if the device is branded HP instead of HPE that is fine. In that case the device can be called `HP-whatever` but the manufacturer remains HPE as they are the parent comanpy that manufactures and owns the device. The idea is to reduce the number of manufacturers, like Cisco and Meraki. That is the thought process at least.

Let me know if this PR is ready to be reviewed

SimonFischer04:
> The idea is to reduce the number of manufacturers, like Cisco and Meraki. That is the thought process at least.
Oh, not having tons of manufacturers makes sense, thanks for explaining

> Let me know if this PR is ready to be reviewed
Yes, then it's ready from my side.

PR #1398, opened 2023-05-28
Update TL-SG3428MP

MatMc:
Previous entry was missing POE info, had 4 incorrect interface speeds, and maximum power usage was specific for the v3 hardware revision. Also added info about differences between hardware revisions and renamed file for consistency.


danner26:
Any reason this was closed @MatMc ?

PR #1401, opened 2023-05-30
Add Cisco N9K-C9332D-GX2B

AuthenticAMD:
Adding Cisco N9K-C9332D-GX2B device type.


danner26:
Hello, we have released a new set of tests and features to assist with properly creating definition files for the DeviceType-Library. These tests will ensure that all files conform to a standardized format and contain valid data. Over the coming weeks we will be adding more tests to handle other currently undefined cases. If you notice anything that you believe is a bug, please @ me in a comment and details what you think is incorrect.

Along with additional test cases we have added a feature that will annotate why tests fail. To access these annotations after a test has completed, please navigate to the _Files changed_ tab within your PR. At the very bottom of the page you will see a list of errors along with the filename in question and the error which was found.

With all this being said, if your PR is validated then great! Please @ me me so I can review it further! If your PR is not valid, please review the issues and correct them; in the meantime I will mark your PR as a draft. After correcting the issues and recieving a valid status, please mark the PR as ready to be reviewed. Thank you!

PR #1405, opened 2023-05-31
Add Dell N2248X-ON

tnagafuchi:


device-types/Dell/PowerSwitch-N2248X-ON.yaml
```diff
@@ -0,0 +1,131 @@
+---
+manufacturer: Dell
+model: PowerSwitch N2248X-ON
+slug: dell-powerswitch-n2248x-on
+part_number: N2248X-ON
+u_height: 1
+is_full_depth: true
```
danner26, line 7: Please add the `airflow` data point if you have it or can find it. Often these are listed in the datasheet, but sometimes they are not

danner26, line 7: Please add the `weight` and associated `weight_unit` data point if you have it or can find it. Often these are listed in the datasheet, but sometimes they are not

device-types/Dell/PowerSwitch-N2248X-ON.yaml
```diff
@@ -0,0 +1,131 @@
...
+    label: RJ-45 Console
+    type: rj-45
+  - name: USB
+    label: USB Console
+    type: usb-micro-b
+power-ports:
+  - name: PSU1
+    type: iec-60320-c14
+  - name: PSU2
+    type: iec-60320-c14
```
danner26, line 19: It appears that these PSUs are modular, is this true?

If so, you should remove these definitions and create `module-bays` for them to slot into. Then, if the modules do not already exist in this repo you should add them to the associated module-types folder

danner26 (changes requested):

device-types/Dell/PowerSwitch-N2248X-ON.yaml
tnagafuchi: Airflow is not listed because there are two types, IO to PSU or PSU to IO.

danner26 (approved):

PR #1411, opened 2023-06-02
Create MikroTik CSS610-8G-2S+

zndrr:
Add device type CSS610-8G-2S+

### DIFF
--- a/device-types/MikroTik/CSS610-8G-2S+
+++ b/device-types/MikroTik/CSS610-8G-2S+
@@ -1,0 +1,28 @@
+---
+manufacturer: MikroTik
+model: CSS610-8G-2S+
+slug: mikrotik-css610-8g-2s-plus
+is_full_depth: false
+airflow: passive
+u_height: 1
+interfaces:
+  - name: Port1
+    type: 1000base-t
+  - name: Port2
+    type: 1000base-t
+  - name: Port3
+    type: 1000base-t
+  - name: Port4
+    type: 1000base-t
+  - name: Port5
+    type: 1000base-t
+  - name: Port6
+    type: 1000base-t
+  - name: Port7
+    type: 1000base-t
+  - name: Port8
+    type: 1000base-t
+  - name: SFP1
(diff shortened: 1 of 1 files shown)


danner26:
It appears you are missing a lot of data from this device. Please make sure the following data points exist if they are not already defined:

- manufacturer
- model
- slug
- part_number
- u_height
- is_full_depth
- airflow
- weight
- weight_unit
- power-ports

zndrr:
@danner26 thanks for reviewing. Added some missing detail in edited pull request.
For the weight detail, unfortunately product pages do not specify, so am unable to add.

Sorry, but I can't figure out what the validation is not satisfied with (Exit Code 1)
Have used CRS326-24G-2S+RM as a point-of-reference for interface label yaml formatting.
Note the labels are visually accurate.

device-types/MikroTik/CSS610-8G-2S+IN.yaml
```diff
@@ -0,0 +1,47 @@
+---
+manufacturer: MikroTik
+model: CSS610-8G-2S+IN
+part_number: CSS610-8G-2S+IN
+slug: mikrotik-css610-8g-2s-in
+is_full_depth: false
+airflow: passive
+u_height: 1
```
danner26, line 8: Please add the `weight` and associated `weight_unit` data point if you have it or can find it. Often these are listed in the datasheet, but sometimes they are not

danner26 (changes requested):

danner26:
Please review the comments above and fix the failed checks. Then mark this as ready for review. Thanks for your contribution so far!

device-types/MikroTik/CSS610-8G-2S+IN.yaml
zndrr: Hi @danner26, thanks for checking again. I previously stated that I was unable to acquire the weight information from MikroTik specifications and any documentation like datasheets. Unfortunately this still holds true.

danner26 (approved):

PR #1430, opened 2023-06-15
Add FireBrick devices

dar72:


device-types/FireBrick/FB2900.yaml
```diff
@@ -0,0 +1,22 @@
+---
+manufacturer: FireBrick
+model: FB2900
+slug: fb2900
```
danner26, line 4: Slugs should have the manufacturer name appended to them, in this case please append `firebrick-` to the slug

device-types/FireBrick/FB2900.yaml
```diff
@@ -0,0 +1,22 @@
...
+part_number: FB2900
+u_height: 1
+is_full_depth: false
```
danner26, line 7: Please add the `airflow` data point if you have it or can find it. Often these are listed in the datasheet, but sometimes they are not

danner26, line 7: Please add the `weight` and associated `weight_unit` data point if you have it or can find it. Often these are listed in the datasheet, but sometimes they are not

danner26 (changes requested):

device-types/FireBrick/FB6202.yaml
```diff
@@ -0,0 +1,22 @@
+---
+manufacturer: FireBrick
+model: FB6202
+slug: fb6202
+part_number: FB6202
+u_height: 1
+is_full_depth: false
```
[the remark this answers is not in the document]
dar72: Cannot find in datasheet

device-types/FireBrick/FB2900.yaml
dar72: Cannot find in datasheet

PR #1432, opened 2023-06-19
Add Cisco UCS-C220-M6S

btriller:
Add Cisco UCS-C220-M6S device and various Cisco PSU module types.


device-types/Cisco/UCSC-C220-M6S.yaml
```diff
@@ -0,0 +1,24 @@
...
+slug: cisco-ucs-c220-m6s
+part_number: UCSC-C220-M6S
+u_height: 1
+airflow: front-to-rear
+is_full_depth: true
+console-ports:
+  - name: Console
+    type: rj-45
+interfaces:
+  - name: ens1f0
```
DanSheps, line 13: These are how the names present in Linux.

IMO, I would name them how they would present in a managed (fabric interconnect connected) environment instead or a generic "Ethernet 1, Ethernet 2"

device-types/Cisco/UCSC-C220-M6S.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - name: ens1f1
+    type: 10gbase-t
+  - name: CIMC
+    type: 1000base-t
+    mgmt_only: true
+module-bays:
+  - name: PSU1
+    position: '1'
+  - name: PSU2
+    position: '2'
```
DanSheps, line 24: > 3 PCIe 4.0 slots plus 1 dedicated 12-Gbps RAID controller slot and 1 dedicated mLOM slot

I would include these slots, as they could house an additional network adapter.

DanSheps (changes requested):
Please try and include weights on all items if at all possible.

danner26:
Hello, we have released a new set of tests and features to assist with properly creating definition files for the DeviceType-Library. These tests will ensure that all files conform to a standardized format and contain valid data. Over the coming weeks we will be adding more tests to handle other currently undefined cases. If you notice anything that you believe is a bug, please @ me in a comment and details what you think is incorrect.

Along with additional test cases we have added a feature that will annotate why tests fail. To access these annotations after a test has completed, please navigate to the _Files changed_ tab within your PR. At the very bottom of the page you will see a list of errors along with the filename in question and the error which was found.

With all this being said, if your PR is validated then great! Please @ me me so I can review it further! If your PR is not valid, please review the issues and correct them; in the meantime I will mark your PR as a draft. After correcting the issues and recieving a valid status, please mark the PR as ready to be reviewed. Thank you!

PR #1434, opened 2023-06-20
add Cisco C9500-12Q

christian-boesch:


device-types/Cisco/C9500-12Q.yaml
```diff
@@ -0,0 +1,48 @@
+---
+manufacturer: Cisco
+model: Catalyst 9500-12Q
+slug: cisco-c9500-12q
+part_number: C9500-12q
+u_height: 1
+is_full_depth: true
```
danner26, line 7: Please add the `airflow` data point if you have it or can find it. Often these are listed in the datasheet, but sometimes they are not

danner26, line 7: Please add the `weight` and associated `weight_unit` data point if you have it or can find it. Often these are listed in the datasheet, but sometimes they are not

device-types/Cisco/C9500-12Q.yaml
```diff
@@ -0,0 +1,48 @@
...
+    type: rj-45
+  - name: usb
+    type: usb-mini-b
+power-ports:
+  - name: PS-0
+    type: iec-60320-c14
+    maximum_draw: 950
+  - name: PS-1
+    type: iec-60320-c14
+    maximum_draw: 950
```
danner26, line 20: Are the PSUs in this unit modular? If so, you should create `module-bays` for the PSUs and then the associated modular PSUs should be added to the module-type folder if they do not already exist

danner26 (changes requested):

device-types/Cisco/C9500-12Q.yaml
christian-boesch: done

christian-boesch: done

device-types/Cisco/C9500-12Q.yaml
christian-boesch: done

danner26 (approved):

PR #1449, opened 2023-07-06
Update Cisco Catalyst 4507R+E Switch

aice09:
Update Cisco Catalyst 4507R+E Switch, add power, line card, and supervisor engine modules.


aice09:
Hi @jsenecal,  I think its not good to add default module for the device because by default Cisco Catalyst 4507R+E Switch don't have module installed.
Its better to let the user of decide what module they want to add.

jsenecal:
> Hi @jsenecal, I think its not good to add default module for the device because by default Cisco Catalyst 4507R+E Switch don't have module installed. Its better to let the user of decide what module they want to add.

Then at least document some of them, otherwise this device template is impossible to use or adapt from.

jsenecal (changes requested):
Remove the file AP7552.yaml from this PR as it does not belong here

PR #1464, opened 2023-07-18
added HPE modules and psu

francesco-noc:
I'm not sure if I can use {vc_position} for interface name. Maybe it can be removed?


danner26:
Can you manually import this file and see if `vc_position` works? if so then this is ready to be merged.

francesco-noc:
It doesn't work. I think because vc_position belongs to vchassis object.

danner26:
That is what I was thinking, but I wasn't sure. I appreciate you testing it for me. Thank you for the contributions as always!

danner26 (approved):

PR #1482, opened 2023-07-24
Revisions and additions to Aruba 2540 series

pv2b:
Add new model


jsenecal (changes requested):

1. Please add the `airflow `data point if you have it or can find it. Often these are listed in the datasheet, but sometimes they are not
2. Please add the `weight `and associated `weight_unit` data point if you have it or can find it. Often these are listed in the datasheet, but sometimes they are not

pv2b:
Hello! I'm working on this, and I figured I'd also complete the 2540 line by adding the final missing model since I'm here and fixing up some other mistakes.

One challenge I'm running into though is that the airflow is listed as "Side to side". As far as I can tell that's not a valid option in the netbox airflow model. I think therefore it makes sense not to populate this since the Netbox data model can't model this type of airflow accurately?

pv2b:
Hello again, I took another look at the option and it seems I misspoke. There are indeed options for "left to right" and "right to left" airflow, however the datasheet doesn't specify which of these directions (or, in fact, both) that these switches run. It could even be a combination of both (which I guess would be mixed?)

Anyway, since Netbox doesn't have the more vague "side to side" option, I don't want to add information that is more specific and possiby incorrect. Even if I had a speciment of each switch (which I don't), since it's not specified by HPE which side to which side the switch airflow goes, there's no guarantee different hardware revisions might do this differently.

danner26:
Hello, I agree if there is no specific left to right or right to left, then it is best to leave it blank for the time being.

danner26:
We have decided that `side-to-side` is too vague. If you are unable to verify the airflow pattern yourself and there is no documentation, then in my opinion I would ask you to please omit it entierly. If you are able to verify the pattern yourself then that is fine enough for me.

pv2b:
> We have decided that `side-to-side` is too vague. If you are unable to verify the airflow pattern yourself and there is no documentation, then in my opinion I would ask you to please omit it entierly. If you are able to verify the pattern yourself then that is fine enough for me.

I'm unable to verify the airflow pattern myself, the devices I'm documented are already racked and it's not practical for me to take them down to check. Also I don't have access to every device in this particular series, so even if I could, I couldn't get consistent information.

Anyway, In that case, I think the PR is currently in the desired state already right now. Let me know if there's anything else you need me to improve.

danner26 (approved):

PR #1485, opened 2023-07-24
Fix height of WS-C3750G-24TS-S

pv2b:
Looks like height was accidentally set to 1.5U in commit 6f1fa04a7401e7ac321a9b54e2fb6c93f573fceb that was intended to implement half-height units.

My best guess is this device type was edited for test purposes and accidentally committed.

I've got this device in front of me and it's definitely 1U.


danner26:
Great catch, thanks!

PR #1495, opened 2023-07-27
Add support for stackwise

deku-m:
According to Platform details 
https://www.cisco.com/c/en/us/products/collateral/switches/catalyst-9300-series-switches/nb-06-cat9300-ser-data-sheet-cte-en.html#Platformdetails

All 9300 have stackwise support


danner26:
I have to review the rest, but the `C9300-48T` does not have dedicated stacking ports. It should actually be under `module-bays` and then the `C9300L-STACK-KIT`

DanSheps:
> C9300-48T

I think you mean the `C9300L-48T`

C9300-48T very much do have dedicated stack ports.

danner26:
Thanks for pointing that out @DanSheps my mistake

danner26 (approved):

PR #1499, opened 2023-07-27
Improvements to PA-220 model

pv2b:
Hello!

I have identified some issues with the rack-mounting modelling of the Palo Alto PA-220 firewall.

According to the Netbox Device Type Library, this is a 1U rackmount device. That is not true. In fact, this is a small firewall that's not rack-mountable.

To rack-mount it, you can use the one of two available rack mount kits, either the [PAN-PA-220-RACKTRAY](https://docs.paloaltonetworks.com/content/techdocs/en_US/hardware/pa-220-hardware-reference/install-the-pa-220-firewall/install-the-pa-220-firewall-in-an-equipment-rack/install-the-pa-220-firewall-using-the-pan-pa-220-racktray.html#id07f39d28-6ee8-4370-936b-a55704c0fc82) that allows you to install two PA-220s and their power supplies into a 1U form factor, or the [PAN-PA-220-RACK-SINGLE](https://docs.paloaltonetworks.com/content/techdocs/en_US/hardware/pa-220-hardware-reference/install-the-pa-220-firewall/install-the-pa-220-firewall-in-an-equipment-rack/install-the-pa-220-firewall-using-the-pan-pa-220-rack-single.html#id8c951c90-edb5-445b-8085-a11c66c2ba0d) that allows for one PA-220 and its PSU to be rack-mounted.

Additionally, there's an issue with how the power input is specified. Currently the Netbox Device Type Library lists this as having two nema-1-15p type power inputs. This is not true. In fact, the PA-220 uses a "power brick" or desktop power supply, that takes in AC power on one end through a C14 inlet (and a bundled power cord of course, depending on country, but that would reasonably not be part of the device itself) and a cable with a fixed DC connector on the other side. The device is bundled with 1 PSU, and running a second is an option.

To solve this I have:

1. Created new device types for the two possible rack mount kits, with device bays for power supplies and firewalls. This is a parent device.
2. Created a new device type for the PA-220's power supply. (It made more sense as a seperate device as opposed to a module, because it doesn't go inside the PA-220 itself, it's a seperate device. This is a child device, because it can go inside a rack mount kit.
3. Amended the PA-220's specification to specify that it's a child device (that optionally can go into a rack-mount kit, or be installed as a non-racked device), and that it takes a 2 x "other" input for the power.

I have used "other" to represent the DC barrel jacks, since Netbox doesn't have a port type for a DC barrel jack (nor do I expect Netbox to have a type for every possible DC barrel jack out there since there are way too many of them). I thought about using "DC Terminal" but a googling tells me that these are probably meant to represent screw-down DC terminals as opposed to barrel jacks.

There's an argument to be made to not model the power supplies seperately here for the sake of simplicity, but I figure if someone wants to omit them from their own model, they can just import the PA-220 and tweak the model to their own simplification, which is easier than adding the abstraction that's missing.


danner26:
Hello, there is a lot here so I am going to try to go over as much as I can. We rely on community contributions like yours, and unfortunately we can only vet the information as much as we have time/the ability to. Sometimes incorrect information like this slips through but we always welcome and appreciate the corrections.

Alright so lets talk about the rack mount adapter. There is already a defined way to do this. What you need to do is create a device-type for the rack mount and ensure it has `subdevice_role: parent`. Then, in the device in question (the PA-220 in this case), set `subdevice_role: child` and ensure it has `u_height: 0`

That is all it takes to handle this case

The correct way to define the power is based on the interface on the device itself. In this case it would be a `dc-terminal` because it is recieving DC current to that Jack, and ultimately it is a DC terminal. It doesnt matter if it is a barrel, molex, or screw down.

The power supply bricks are not suppose to be modelled in NetBox (or at least NB hasnt been designed for that).

danner26:
Power supplies are also **not** devices

pv2b:
> Alright so lets talk about the rack mount adapter. There is already a defined way to do this. What you need to do is create a device-type for the rack mount and ensure it has `subdevice_role: parent`. Then, in the device in question (the PA-220 in this case), set `subdevice_role: child` and ensure it has `u_height: 0`
> 
> That is all it takes to handle this case

Great, this is what I'm already doing in this PR.

> The correct way to define the power is based on the interface on the device itself. In this case it would be a `dc-terminal` because it is recieving DC current to that Jack, and ultimately it is a DC terminal. It doesnt matter if it is a barrel, molex, or screw down.
> 
> The power supply bricks are not suppose to be modelled in NetBox (or at least NB hasnt been designed for that).

> Power supplies are also not devices

I have amended the model to omit modelling the power supply as per your feedback, instead using the dc-terminal on the power port of the device itself. I guess it makes sense in a way, you can think of the AC adapter as a type of a "cable".

pv2b:
Sorry for the noisy commits on this PR, I'll try to install the required linter/tests for my next PR

pv2b:
>  Error: test_definitions[device-types/Palo Alto/PA-220-RACK-SINGLE.yaml-schema0]
> 
> Failed: device-types/Palo Alto/PA-220-RACK-SINGLE.yaml has does not appear to have a valid power source. Ensure either "power-ports" or "interfaces" with "poe_mode" is defined.
> 
> Error: test_definitions[device-types/Palo Alto/PA-220-RACK-TRAY.yaml-schema1]
> 
> Failed: device-types/Palo Alto/PA-220-RACK-TRAY.yaml has does not appear to have a valid power source. Ensure either "power-ports" or "interfaces" with "poe_mode" is defined.

Not sure how to fix the errors though, this perfectly correct. The rack trays have no power source, they're just dumb pieces of metal. :-)

danner26:
Its all good

Please make sure the model and part_numbers match the devices exactly. I do not think the `PA-220 Rack Single` is a legitament name

pv2b:
> Its all good
> 
> Please make sure the model and part_numbers match the devices exactly. I do not think the `PA-220 Rack Single` is a legitament name

The part has no "official" model name as far as Palo Alto goes, as far as I can tell. Palo Alto's documentation only names them by their "part numbers", i.e. PAN-PA-220-RACKTRAY and PAN-PA-220-RACK-SINGLE. (Yes, it infuriates me too how one of them has a dash, and the other one doesn't, but I don't make the part numbers...)

The part numbers are correct and match exactly.

(Reference: https://docs.paloaltonetworks.com/hardware/pa-220-hardware-reference/install-the-pa-220-firewall/install-the-pa-220-firewall-in-an-equipment-rack)

Also, the linter is a little picky when it comes to filenames and slugs. It either wants the filename and slug to be based on the model name or on the part number. I can't base the slug on the part number because that means I'd have to label it starting with PAN-, no other files in the current repo does that. So the only option is to name the file based on the device name.

So since the kits have no "official" name, I just ended up setting their device names so make the filenames of the YML files pretty (which I thought to do by making them mostly match the part numbers...)

Anyway, I hope you understand the constraints the linter is putting me under and why the names were set the way they were. I've committed a different suggestion that you might like better? The device names are still fictional, but they have the advantage of turning into decent slugs and filenames.

PR #1505, opened 2023-07-28
Adding equipment from the manufacturer DATACOM

rogeriomendesp:


danner26:
Hello, you need to remove the pickle file from your PR. The easiest way to do this on your current commit is to download the pickle file from the upstream master, add it in and the commit it. Lastly push that new commit up and you no longer should see it under the Files changed tab.

This should bring it up to the same baseline. 

In the future do not commit the pickle file. You should not be modifying this/commit it.

PR #1509, opened 2023-07-31
Juniper EX4550 Family (Update)

suom1:


danner26:
Is there any chance these modules need to use `{module}` in the name? Or will they only ever have that name?

suom1:
> Is there any chance these modules need to use `{module}` in the name? Or will they only ever have that name?

I don't see a situation when they would have another name!

PR #1533, opened 2023-08-03
Arista 7050S Family (Update)

suom1:
PSUs are found in #1531


danner26 (approved):
As always thank you for taking the time to update the definition files! This repo couldnt survive without dedicated contributors like you!

PR #1540, opened 2023-08-04
Fix Part Number for Device Type Aruba 6100-48G-4SFP+

desnoe:
Part number was duplicated on the POE and non-POE models which have different part numbers


danner26:
Hello, please do not make modification to the pickle files and commit those. Please remove them from your commit so that the tests can run fully.

desnoe:
GitHub is adding these 2 files, and I can't figure out why. On my side, there is only a single file that is modified

PR #1541, opened 2023-08-05
add Mikrotik CRS318 device

a-willem:
Add this device to the base. Also known as netPower 15FR.


device-types/MikroTik/CRS318-1Fi-15Fr-2S-OUT.yaml
```diff
@@ -0,0 +1,46 @@
+---
+manufacturer: MikroTik
+model: CRS318-1Fi-15Fr-2S-OUT
+slug: mikrotik-crs318-1fi-15fr-2s-out
+is_full_depth: false
+u_height: 1
```
danner26, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

danner26, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

danner26 (changes requested):

PR #1545, opened 2023-08-10
Improvements to HPE Synergy 12000 and related devices/models

pv2b:
There's a HPE Synergy 12000 Frame model in this repository, but there are a lot of ways to improve the existing model, that I decided to contribute to this repo.

Issues addressed:

- The frame is incorrectly modelled with fixed power supplies. In reality, fewer than 6 PSU:s are a valid configuration, so I made those modular. Also there are more PSU types out there, but I just modelled what we use.
- The frame incorrectly had a bunch of interfaces added that belonged to interconnect modules, and in one very specific configuration of interconnect modules that didn't match my configuration (or most other people's configuration probably). To address this, I've created module slots for interconnect modules and frame link modules to add the correct interfaces from there, as well as created module types for the modules that occur in our environment specifically.
- There was no modelling of the console ports for the frame itself. That has been added.
- There was a model for the SY 480 Gen10, but it had the partnumber for the SY 480 Gen10 Plus. I split this up into two different models.
- There was no modelling of the composer or appliance bays, that has been added.
- Weight and airflow information has been added where available, as well as links to QuickSpecs.

Some design justifications:

- There is no Netbox Console Port Type for DisplayPort cables. I'm not sure if the intent for console connections is to model video consoles, but for now I just put type "other" on these.
- There is no Netbox Interface Type for CXP28 used for the ILM's. I'm actually not even sure if it's Ethernet to be honest, so I just left it on "other".
- I have modelled everything that is modular as modular, with the exception of fan modules. The reasoning is that the 10 fan modules always have to be populated under normal operating circumstances. They are therefore not "optional" in any way, and modelling these doesn't add any useful functionality to Netbox users. To compensate for this, the weight of 10 fan modules has been added to the weight of the Synergy Frame itself. If someone desires to use Netbox to manage serial numbers of fans, for example, they could easilly do that themselves by adding inventory items for the fans as desired.
- I have only modelled devices and models that we personally use (with the exception for the SY 480 Gen10 Plus which I fixed), there are a lot of modules and devices out there for the Synergy ecosystem, but that'll be someone else's problem.
- There has been no attempt to model power distribution between Synergy chassis and blade devices, since there's no way to model a backplane or midplane in Netbox, except through cables, which, again, needlessly increases complexity for most use cases. If someone needs to track this, they're welcome to add cables and ports for their use case.
- There has likewise been no attempt to model the chassis networking through the Synergy midplane. Again, this is because Netbox doesn't let you model a midplane. I suspect most people will want to add virtual interfaces on the servers themselves to match the hardware and operating system configuration of the software on the servers. If someone really wants to model it, it's possible, you'd have to add interfaces to the interconnect modules to represent the data lanes on the midplane, and then manually run "cables" to represent these lanes. But I don't think most people will want to do that.


danner26:
Hey, thanks for all of this! I appreciate you explaining your reasoning behind this. Let me clarify a few things that make these changes invalid.

- `console-ports` are designed for ports which give the user console access. This does not apply the KVM USB ports, VGA ports, DisplayPorts, etc. Those are not valid types in NetBox at the moment, and thus should be omitted.
- `other` is a type we generally try to not use.. we try to be as specific as possible. If you could reference the datasheet and find the exact interface type, that would be best. If the interface type does not already exist, link the documentation and I can add it to the upstream NetBox repository.
- I understand and in this case agree with how you approached the fan situation. If someone wants to break this out they can submit a seperate PR
- As long as the frame has the frames input power, that is okay.


With this all being said, are these "modules" blade devices? For reference, the Dell M620 is a blade which technically would fit under `device-bays` rather than `module-bays`. Please just clarify if the current modelling in your PR is correct or if those should be in `device-bays`

pv2b:
> Hey, thanks for all of this! I appreciate you explaining your reasoning behind this. Let me clarify a few things that make these changes invalid.
> 
>     * `console-ports` are designed for ports which give the user console access. This does not apply the KVM USB ports, VGA ports, DisplayPorts, etc. Those are not valid types in NetBox at the moment, and thus should be omitted.

When you say "console access", I'm assuming you specifically mean *serial* console. (For me, the word "console" can also mean keyboard/video/mouse, so I just want to clarify.)

If the intent is to only model serial console ports, then of course, I will be happy remove any such KVM console ports.

>     * `other` is a type we generally try to not use.. we try to be as specific as possible. If you could reference the datasheet and find the exact interface type, that would be best. If the interface type does not already exist, link the documentation and I can add it to the upstream NetBox repository.

Other than the KVM console ports, the only use of the `other` device type remaining is on the CXP28 ports on the interconnect and virtual connect modules.

For example, as per the [QuickSpecs for the HPE Virtual Connect SE 100Gb F32 Module for Synergy](https://www.hpe.com/psnow/doc/a00067745enw) the ports are listed as "CXP28" at a data rate of 300 Gb/s.

According to the [[HPE Synergy Cabling Guide](https://techlibrary.hpe.com/docs/synergy/shared/cable_config/index.html)](https://techlibrary.hpe.com/docs/synergy/shared/cable_config/GUID-156A007D-0863-441F-A91A-3640380E301A.html), "These CXP28 interconnect link cables connect only the HPE Synergy 100Gb F32 Module (master) and the HPE Synergy 25/50Gb Interconnect Link Module (satellite)."

I therefore don't really have anything to go on more than the fact that they are CXP28 connectors, but that only tells us about the connector, it doesn't tell us about the protocol in use on this. There's hint that the data rate is 300 Gb/s, which could be a way to exclude a specific standard, but couldn't be used to determine one.

The reason the protocol is important, not just the connector type, is that in Netbox, CXP (as well as SFP and similar) are all listed under the "Ethernet (modular)" category. For example the existing CXP is listed as "CXP (100 GBE)". I've not been able to find any Ethernet standard that uses CXP28, except this [one article](https://www.connectortips.com/how-have-cxp-connectors-and-cables-evolved-in-recent-years/) from 2016, but I've been unable to confirm the data rate of an Ethernet standard using CXP28.

But anyway, even if we could find that, that doesn't confirm that it's Ethernet. For all intents and purposes, it's HPE Proprietary, but using a CXP28 connector.

That leaves the Netbox project with a choice to either:

1. Re-factor the SFP/CXP etc connector types to remove any references to "Ethernet", so that these types can also be used for non-Ethernet applications (such as fiber channel), and then add CXP28 as a supported standard.
2. Find an appropriate Ethernet standard that uses the CXP28 connector, add it there, and then set the CXP28 connector on these modules to use this standard (even if it might be a lie).
3. Add a HPE-specific CXP28 type to Netbox.
4. Do nothing, and just use "other" for this case.

In my opinion, option 4 is probably the right one, because for all intents and purposes, this is a proprietary vendor-specific connector/protocol combination, that only ever is supposed to be connected to other devices in the same ecosystem, and it doesn't make sense to pretend it's Ethernet when it might not be, and it makes even less sense to add vendor-specific protocol/connector combos to Netbox.

Option 1 also makes sense, but it would probably be subject to a lot of opinions, and larger changes across the board that would impact the whole Netbox ecosystem and the entire device type library, so it probably makes sense not to do that. :-) (Maybe as a compromise option, maybe add new connector types for SFP, SFP28, CXP, CXP28, QSFP, QSFP28 that are not Ethernet-specific and allow those to be used, in addition to the existing Ethernet-specific ones?)

> With this all being said, are these "modules" blade devices? For reference, the Dell M620 is a blade which technically would fit under `device-bays` rather than `module-bays`. Please just clarify if the current modelling in your PR is correct or if those should be in `device-bays`

That's a question of definition. The following types of bays are available on a HPE Synergy Frame:

Device Bay 1-12: This is where the actual "blade servers" go. These are devices IMO, because they have their own IP addresses and network identities. It makes sense to model them as devices.

Appliance Bay 1-2: This is where "appliances" for management of the Synergy frame ecosystem such as the Synergy Composer or Synergy Image Streamer goes. As far as I understand, they're small servers used for management of the Synergy Ecosystem, and thus, on second thought, it probably makes more sense to give these "device" rather than "module" status which they have right now. I'll have to confirm with someone who's more familiar with these systems.

Interconnect Bay 1-6: This is where "interconnect modules" goes. As far as I understand them, the best way I can explain them is that they're like switches that slide into the Synergy Chassis, which provide connectivity to the external network on the outside (these ports have been modelled) and to the individual blade servers on the other side (these connectors are "internal" to the frame and have not been modelled, but there is an argument to be made to model these so you can connect them with "cables" to the servers...). There's an argument to be made that these should be "devices" as well, since they can be [individually managed through SSH by going through the Synergy Composer](https://support.hpe.com/hpesc/public/docDisplay?docId=a00118216en_us&docLocale=en_US&page=s_accss-intrcnts-usg-ssh.html) even they don't have any IP addresses.

Frame Link Module 1-2: These are similar to Interconnect Modules, except they allow connectivity to appliances rather than devices. They could be considered devices for a similar reason. They have their own serial consoles so I assume you can configure them with SSH somehow as well.

Power Supplies 1-6: These are obviously modules.

In summary, on second thought, I'm leaning towards changing Appliances, Interconnects and Frame Link Modules into devices rather than modules. I'll have to confirm with someone with more experiences with these systems though.

pv2b:
I've had a discussion with our Synergy admin and he agrees. Appliances, Frame Link Modules and Interconnect Modules should be modelled as devices, not modules. I'll refactor the code accordingly.

pv2b:
OK, that's better, however, I'm not entirely convinced I'm done. There's no network interface in the model that belongs to the Synergy Composer for example, and that for sure has an IP address. That probably needs to be added.

I'm converting this back to a draft for now until that's been addressed.

pv2b:
Okay, I've added the missing Management and iLO interfaces. Further complicating things, typically the Synergy Composer will be deployed in a cluster and have no "real" IP address of itself other than the iLO one, instead it'll be floating. I guess this can be modelled as an FHRP group. 

Again, interface type is virtual, although in reality there are backplane links going back to the frame link modules, which IMO is unneccessary to model inside Netbox.

I found [this forum post](https://community.hpe.com/t5/hpe-synergy/hpe-synergy-12000-initial-acess-issues/m-p/7144643/highlight/true#M793) indicating that FLM:s do not receive their own interfaces, and I've not been able to find any IP addresses for the IC modules, they seem to be managed through the composer. And therefore should have no interfaces associated to them. I guess from the point of view of Netbox they'll be managed through "magic".

I think this is ready for a re-review now.

danner26:
`console-ports` are defined in the [NetBox documentation](https://docs.netbox.dev/en/stable/models/dcim/consoleport/) as a port that:
> provides connectivity to the physical console of a device
Console Ports are not for monitor/keyboard interfaces that do not provide access to the console via that one cable. As a collective the community has decided that VGA/keyboards should not be modelled within NetBox at this time. That functionality might be introduced in the future, but `console-ports` would likely not be the area those would be defined in.

In regards tothe CXP28 connector, I agree. Based on your logic and detailed response the type of `other` will be accepted in this condition

danner26 (approved):

PR #1546, opened 2023-08-10
Adding Fortinet FG-800D & FD3700D

arcticash:
Added the following two Firewalls, based off the other Fortinet templates and the data sheets

- Fortinet FortiGate 800D
- Fortinet FortiGate 3700D


device-types/Fortinet/FG-3700D.yaml
```diff
@@ -0,0 +1,97 @@
...
+    type: usb-mini-b
+power-ports:
+  - name: PS1
+    type: iec-60320-c14
+    maximum_draw: 840
+    allocated_draw: 300
+  - name: PS2
+    type: iec-60320-c14
+    maximum_draw: 840
+    allocated_draw: 300
```
danner26, line 26: It appears that these PSUs are hot swappable. These should thus be modelled under `module-bays`, and then the associated PSU modules should be modelled in the module-types folder if they do not already exist there

device-types/Fortinet/FG-800D.yaml
```diff
@@ -0,0 +1,96 @@
...
+    type: rj-45
+power-ports:
+  - name: PWR1
+    type: iec-60320-c14
+    maximum_draw: 187
+    allocated_draw: 128
+  - name: PWR2
+    type: iec-60320-c14
+    maximum_draw: 187
+    allocated_draw: 128
```
danner26, line 25: It appears that these PSUs are hot swappable. These should thus be modelled under `module-bays`, and then the associated PSU modules should be modelled in the module-types folder if they do not already exist there

danner26 (changes requested):

danner26 (approved):

PR #1561, opened 2023-08-18
Add Dell PowerConnect 3548p switch to device types

Aaron-von-Awesome:
Add Dell PowerConnect 3548p switch to device types:

- https://dl.dell.com/manuals/all-products/esuprt_ser_stor_net/esuprt_powerconnect/powerconnect-3548p_user's%20guide_en-us.pdf
- https://downloads.dell.com/manuals/all-products/esuprt_ser_stor_net/esuprt_networking/esuprt_net_fxd_prt_swtchs/powerconnect-3548p_reference%20guide_en-us.pdf


device-types/Dell/PowerConnect-3548p.yaml
```diff
@@ -0,0 +1,219 @@
+---
+manufacturer: Dell
+model: PowerConnect 3548p
+slug: dell-powerconnect-3548p
+u_height: 1
+is_full_depth: false
```
danner26, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

1. Please add the `part_number` if it differs from the model name. In most cases there is a difference, but sometimes there is not. 
2. Please add the `airflow` data point if you have it or can find it. Often these are listed in the datasheet, but sometimes they are not
3. Please add the `weight` and associated `weight_unit` data point if you have it or can find it. Often these are listed in the datasheet, but sometimes they are not

device-types/Dell/PowerConnect-3548p.yaml
```diff
@@ -0,0 +1,219 @@
...
+console-ports:
+  - name: rs-232
+    type: de-9
+power-ports:
+  - name: Power
```
danner26, line 11: Does the device actually report the power is named `Power`? We want to make sure that the interface names match what the configuration/device reports them as

danner26 (changes requested):

device-types/Dell/PowerConnect-3548p.yaml
Aaron-von-Awesome: I updated this - according to the docs the name is "PS1", and it has a redundant power supply / extended power supply port, so I went ahead and added that as well.

danner26: Great, thanks!

danner26:
Thanks for the changes. If this PR is ready for review please mark it as such

danner26 (approved):

PR #951, opened 2022-11-03
Opengear ACM7008-2-L

cerealcable:
This is essentially exactly the same as the ACM7008-2-M but doesn't have SFP ports, only ethernet.

PR #1571, opened 2023-08-22
Add Opengear ACM7008-2-L

shepherdjay:
Similar to the existing opengear ACM7008-2-M but includes cellular modem and does not include sfp ports


shepherdjay:
I split the difference currently but would be open to feedback... IE the console ports that users like me are most likely want to create netbox custom links for direct ssh to have been renamed to match exactly as needed for config and direct ssh serial.

whereas the connectivity ports are labeled as on device and referenced in documentation.

For the power I saw the previous comment on PR #951 and I'm not certain how power adapters should be modeled. Would that be modeled as a separate module and then configure the opengear with a module bay for power? Or would it be left empty and would be filled in with component?

The device ships with a wall wart style connector with several configurable swapouts depending on country region

Additionally there is an adapter that allows for a c13/c14 plug

danner26:
Hello, interface naming should match what the configuration uses. For instance, if you are automating a configuration deplyment you want to ensure the naming in netbox is the same as the configuration on the device is expecting. Hopefully that clears up the confusion

If the PSUs are modular, then you should create `module-bays` on the device for the PSUs, and the associated PSUs that this device uses should be modelled within the respective module-types folder in this repo

danner26:
I am going to mark this PR as a draft while you review my comments. Once you feel this is ready to be reviewed, please mark the PR as ready to review and let me know!

shepherdjay:
> Hello, interface naming should match what the configuration uses. For instance, if you are automating a configuration deplyment you want to ensure the naming in netbox is the same as the configuration on the device is expecting. Hopefully that clears up the confusion
> 
> If the PSUs are modular, then you should create `module-bays` on the device for the PSUs, and the associated PSUs that this device uses should be modelled within the respective module-types folder in this repo

shepherdjay:
Re-opened -- was moving too fast this morning without coffee and accidently closed instead of commenting.

@danner26 -- I could definitely see both sides of the argument where its valuable for it to follow the config or valuable for it to present as labeled especially when in config `lan` is not a very good correlation with net2 

Regarding the PSUs I will make a crack at the modular implementation.

danner26:
Sounds good, if you need any help feel free to ping me!

shepherdjay:
@danner26 -- Actually as I started implementing this I don't think doing the power as a module is the right direction forward for a couple of reasons. I will explain below and mark as ready then if still desired to move forward with a module output we can.

1. Modeling - Unlike a slide in power module on say a cisco switch the power input on the opengear is not modular. The input is a 4 pin 9-30v dc connector. I'm not well versed enough to recognize exact type so I set other. However, this connection is not changeable. You can get various adapters but I'm not aware of adapters being modeled as modules in other devices, for example returning to cisco on most of their units you can buy C14 or 5-15P plugs but it doesn't appear we model the cables in that manner.

2. Practicality - Since netbox does not appear to force a connector type to match between power port and power outlet creating this a module adds overhead to both the maintenance of the repo and user overhead where deploying a device requires filling in the power module.

danner26:
Hey @shepherdjay. I have no idea what I was looking at earlier, I must have confused this PR with another. I say that because this device clearly does NOT have modular power. The power for this device would be `dc-terminal` because on the device it is a DC Terminal jack. We model power based upon the interface on the physical device, not the adapters which are used with it. Please change the power type to be `dc-terminal`

That all being said, the idea of this repository is to be a source of truth. As such, even if modelling the data correctly is a PITA, it is nessecary. This is also a living repository, meaning that it has been around since the early days of NetBox. You might notice older device definitions which are lacking information we ask for now.. this is just one of the struggles we have to deal with. Since the repo is community driven, NetBox relies on people like you and me to create and update definition files as time goes on

One last thing, this device uses passive cooling so please add `airflow: passive` to the device definition

Thanks and again sorry for the confusion!

shepherdjay:
@danner26 -- Awesome, thanks for pointing me to the right power type. I have updated it and the air-cooling as requested.

danner26 (approved):

PR #1577, opened 2023-08-24
Add Extreme Networks Switches and modules

mag-marc-schulz:
Hello,
I've added some Extreme Networks Switches and Modules.

Thanks und kind regards,
Marc


danner26:
I believe it has been proposed and it is still on the list of things to look at. For the time being we attempt to model the defaults the device comes with from the manufacturer

Issue #1584, opened 2023-08-27
Duplicate positions for Power Supply for several HP ProLiant models


pv2b:
> Now that I'm thinking about it...it's almost like it's just ignoring my Module bay choice. Because it seems like its trying to make PowerPort - PSU-PSU1 when really it should be making PowerPort - PSU-PSU2.

Well howdy ho this is what it was. In the yaml file the position of PSU2 is PSU1...so that's what my issue was.

_Originally posted by @Mailstorm-ctrl in https://github.com/netbox-community/netbox/issues/13574#issuecomment-1694516730_

To flesh out the above a little, if we look for example at the [DL560 Gen10 model](https://github.com/netbox-community/devicetype-library/blob/aed11b3f8a3a42a26b9dbbfa35b1aa54c77b6333/device-types/HPE/ProLiant-DL560-Gen10.yaml), we can see that there are duplicate "position"s on the power supply, causing the problem @mailstorm-ctrl was having and misattributed as a Netbox bug. I looked through the repo and I found the problem on a few other servers. I'll get a PR going to try to find any duplicate module bay positions, but it's probably a great idea to also set up a test to detect this condition. Maybe as a pre-commit hook?

Not sure how that would work exactly though, having duplicate positions is in itself not invalid (they could be different module *types*)

danner26:
Thanks for figuring this out. I will close this issue as we will be implementing the fixes.

PR #1586, opened 2023-08-27
Fix mis-numbered module bays on HP Proliant servers

pv2b:
Fixes some mix-numbered module bays on HP Proliant Servers

See issue #1584 and also pull request #1583 that fixes a subset of these problems.

(Probably #1583 should be merged first, this shouldn't conflict with that)


danner26:
I believe we have resolved this in PR#1608

PR #1585, opened 2023-08-27
Add script to find duplicate module bay positions

pv2b:
This adds a script that looks for mistakes in device type definitions, specifically where multiple module bays have the same position, and also have similar names.

This will find mistakes such as:

```yaml
module-bays:
  - name: PSU1
    position: PSU1
  - name: PSU2
    position: PSU1
```

But will ignore situations like:

```yaml
module-bays:
  - name: Power Supply 1
    position: '1'
  - name: Power Supply 2
    position: '2'
  - name: Expansion Slot 1
    position: '1'
  - name: Expansion Slot 2
    position: '2'
```

Because "Power Supply 1" and "Expansion Slot 1" are dissimilar names ([Levensthein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) > 2).

This script was created to see if the same/similar errors as in #1584 occur in more than one model. This is the output at time of writing:

```
In HPE\ProLiant-DL325-Gen10-Plus-v2.yaml, module bays PSU1 and PSU2 both have position PSU1.
In HPE\ProLiant-DL325-Gen10-Plus.yaml, module bays PSU1 and PSU2 both have position PSU1.
In HPE\ProLiant-DL325-Gen10.yaml, module bays PSU1 and PSU2 both have position PSU1.
In HPE\ProLiant-DL360-Gen10-Plus.yaml, module bays PSU1 and PSU2 both have position PSU1.
In HPE\ProLiant-DL360-Gen10.yaml, module bays PSU1 and PSU2 both have position PSU1.
In HPE\ProLiant-DL365-Gen11.yaml, module bays PSU1 and PSU2 both have position PSU1.
In HPE\ProLiant-DL380-Gen10-Plus.yaml, module bays PSU1 and PSU2 both have position PSU1.
In HPE\ProLiant-DL380-Gen10.yaml, module bays PSU1 and PSU2 both have position PSU1.
In HPE\ProLiant-DL560-Gen10.yaml, module bays PSU1 and PSU2 both have position PSU1.
In HPE\ProLiant-ML350p-Gen8.yaml, module bays PCIe8 and PCIe9 both have position PCIe9.
In QNAP\TS-873.yml, module bays Slot 1 and Slot 2 both have position rear.
```

I'm keeping the PR to introduce this script seperate to fixing the models above for clarity. A second PR will be created to fix the mistakes above.

PR #1587, opened 2023-08-27
Improve documentation for module bay positions (and remove a spurious position)

pv2b:
The documentation for the "position" attribute in README.md was misleading. This is what the documentation says:

> The module bay's position within the parent device

This isn't very clear, and it can easilly be interpreted as a physical description of where you can find the port. For example, this module bay definition that I found for the QNAP TS-873:

```yaml
module-bays:
  - name: Slot 1
    position: rear
  - name: Slot 2
    position: rear
```

This PR clears up the documentation for module bay positions to clarify what the position attribute is for, removes this spurious position definition, and attempts to teach users how the Automatic Component Renaming feature works, hopefully this will lead to better models.

(I've not done an exhaustive search for mis-use of the position attribute on module bays, this just popped up when doing an unrelated search for duplicate module bay positions, see #1585)

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,7 +168,7 @@ The available fields for each type of component are listed below.
 
 - `name`: Name
 - `label`: Label
-- `position`: The module bay's position within the parent device
+- `position`: The alphanumeric position in which this module bay is situated within the parent device. When creating module components, the string `{position}` in the component name will be replaced with the module bay's `position`. See the [NetBox Documentation](https://docs.netbox.dev/en/stable/models/dcim/moduletype/#automatic-component-renaming) for more details.
 
 #### Device Bays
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -168,7 +168,7 @@ The available fields for each type of component are listed below.
 
 - `name`: Name
 - `label`: Label
-- `position`: The module bay's position within the parent device
+- `position`: The alphanumeric position in which this module bay is situated within the parent device. When creating module components, the string `{position}` in the component name will be replaced with the module bay's `position`. See the [NetBox Documentation](https://docs.netbox.dev/en/stable/models/dcim/moduletype/#automatic-component-renaming) for more details.
```
abhi1693, line 171: I believe it's `{module}` which gets replaced

abhi1693 (changes requested):

README.md
pv2b: My bad, fixed

danner26 (approved):

danner26:
I appreciate the update to the documentation!

PR #1589, opened 2023-08-28
Add Cisco ESA-C390

mag-marc-schulz:
Hello,
I've added the Cisco ESA-C390 and a power supply module.

Thanks in Advance and Kind regards.
Marc

### DIFF
--- a/device-types/Cisco/ESA-C390.yaml
+++ b/device-types/Cisco/ESA-C390.yaml
@@ -9,19 +9,21 @@ airflow: front-to-rear
 console-ports:
   - name: Console
     type: rj-45
+  - name: RPC
+    type: rj-45
 interfaces:
   - name: Management
     type: 1000base-t
     mgmt_only: true
-  - name: GigabitEthernet0/0
+  - name: Data 1
     type: 1000base-t
-  - name: GigabitEthernet0/1
+  - name: Data 2
     type: 1000base-t
-  - name: GigabitEthernet0/2
+  - name: Data 3
     type: 1000base-t
-  - name: GigabitEthernet0/3
+  - name: Data 4
     type: 1000base-t
-  - name: GigabitEthernet0/4
+  - name: Data 5
     type: 1000base-t
 module-bays:
   - name: PSU1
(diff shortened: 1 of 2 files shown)


device-types/Cisco/ESA-C390.yaml
```diff
@@ -0,0 +1,30 @@
...
+slug: cisco-esa-c390
+part_number: ESA-C390-K9
+u_height: 1
+is_full_depth: true
+airflow: front-to-rear
+console-ports:
+  - name: Console
+    type: rj-45
+interfaces:
+  - name: Management
```
DanSheps, line 13: Typically it would not be named management, it would typicially be named "GigabitEthernet0" if it is a device running an IOS like system.

If it is a IMC type interface, I would recommend using "CIMC" for the name.

device-types/Cisco/ESA-C390.yaml
```diff
@@ -0,0 +1,30 @@
...
+manufacturer: Cisco
+model: ESA C390
+slug: cisco-esa-c390
+part_number: ESA-C390-K9
+u_height: 1
+is_full_depth: true
+airflow: front-to-rear
+console-ports:
+  - name: Console
+    type: rj-45
```
DanSheps, line 11: You are missing the RPC port.

Not sure if it is best to classify it as a Console or an interface though.

device-types/Cisco/ESA-C390.yaml
mag-marc-schulz: The Device is running an on a AsyncOS. 
I had the output of the interfaces sent to me:
```
Currently configured interfaces:
1. Data1_Inside (5.5.5.5/24 on Data 1: something.example.com)
2. Data2_Outside (5.5.5.5/24 on Data 2: something.example.com)
3. Mgmt (5.5.5.5/24 on Management: something.example.com)
```
After that I renamed the interfaces but the management interface is called "Management".
Please feel free to let me know if you see it otherwise.

device-types/Cisco/ESA-C390.yaml
mag-marc-schulz: I've added the PRC Port to then Console Section. I think this fits best in this section.

danner26 (approved):

PR #1593, opened 2023-08-29
Add and Update C9800-L Series

shepherdjay:
Added the fiber version of the C9800-L designated with the suffix F

Modified the C9800-L-C with the correct names, interface types, power types, and other physical attributes.

Modified both C9800-L-C and C9800-L-F units to represent that they are child devices intended to be mounted either outside the rack on a wall, on a rack mounted shelf, or using a dedicated cisco provided rack mount which mounts two units side-by-side.

Add the C9800-L 2U Rack Mount as a parent device.


device-types/Cisco/C9800-L-C.yaml
```diff
@@ -3,8 +3,12 @@ manufacturer: Cisco
 model: Catalyst 9800-L-C
 slug: cisco-c9800-l-c
 part_number: C9800-L-C
-u_height: 1
+u_height: 0
+subdevice_role: child
```
DanSheps, line 7: Before you ask @danner26 this I think is perfectly acceptable.  The device can be slotted into a 1.5 (or 2, can't remember) RU shelf (2 WLC's in 1 device).

device-types/Cisco/C9800-L-C.yaml
```diff
@@ -17,19 +21,19 @@ interfaces:
...
     mgmt_only: true
-  - name: TenGigabitEthernet1
-    type: 10gbase-t
-  - name: TenGigabitEthernet2
-    type: 10gbase-t
-  - name: GigabitEthernet1
+  - name: TwoGigabitEthernet0/0/0
     type: 2.5gbase-t
-  - name: GigabitEthernet2
+  - name: TwoGigabitEthernet0/0/1
```
DanSheps, line 26: Is this actually the name of the interface from "show interface status"?

device-types/Cisco/C9800-L-C.yaml
shepherdjay: The Cisco provided one is a 2u shelf.. we actually mount this on a 1u shelf ourselves. But I've added the official 2U shelf as part of this PR

device-types/Cisco/C9800-L-C.yaml
shepherdjay: In our deployment it is. This is due to it being a 2.5 Gigabit Interface which is referenced to cisco as TwoGigabitEthernet... the `/` placement I'm not 100% on as we run ours in cluster.. but their cluster is active/passive synced and it doesn't address the interfaces separately like you might see in a switchstack so I do think its supposed to be the full `0/x/x` 

And yes oddly they number these units starting at 0... possibly hold over from aireos days I think.

danner26: Are you able to show us an output from the cli? If so that would be great

Also, is this by default or does it default to the removed name? We want to keep this repo with the defaults as much as possible

shepherdjay: Hey @danner26 - I was looking to see if easy way to verify whether interface name changes after joining a cluster or not. Was able to reference [this youtube video](https://youtu.be/UhZedbZ1pIA?si=oBn2d__QrMSeH-JG&t=770) where they setup a stand alone non-ha 9800-L to confirm yes the names are still 0/x/x regardless of cluster status -- this also shows the interface names as `Tw...` as discussed

If a cli output is still needed I can get one will just need to redact a bunch from it --

danner26 (approved):

danner26:
I will just wait for @DanSheps to approve the changes, then we can merge this. Thanks!

DanSheps (approved):

PR #1594, opened 2023-08-30
Create MES5324.yml

GitRost:
Fix Mgmt interfice


device-types/Eltex/MES5324.yml
```diff
@@ -0,0 +1,79 @@
+---
+manufacturer: Eltex
+model: MES5324
+slug: eltex-mes5324
+part_number: MES5324
+is_full_depth: false
```
danner26, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

device-types/Eltex/MES5324.yml
```diff
@@ -0,0 +1,79 @@
...
+console-ports:
+  - name: con0
+    type: rj-45
+power-ports:
+  - name: PS0
+    type: iec-60320-c14
+    maximum_draw: 107
+  - name: PS1
+    type: iec-60320-c14
+    maximum_draw: 107
```
danner26, line 79: It appears that the PSUs on this device are modular. If that is the case, then you should be modelling these under `module-bays` and the applicable modular PSUs should be added under the module-types folder

danner26 (changes requested):

danner26:
Did you just ignore both of my comments and request a re-review?

GitRost:
> Did you just ignore both of my comments and request a re-review?

Sorry if I inserted something in the wrong place and clicked the wrong thing.
I don’t know how to use github, and nothing is intuitive here.

PR #1590, opened 2023-08-29
Add Pulse Secure PSA Series

mag-marc-schulz:
Hello,
I've added the Pulse Secure PSA Series and a PWR Module.

Please check and merge into master.

Thanks and kind regards,
Marc

PR #1595, opened 2023-08-31
Add Ivanti Pulse Secure PSA Series

mag-marc-schulz:
Hello,
this is a follow up pull request for #1590 .
I changed the Pulse Secure files and moved it to the "Ivanti" folder.

I hope everything is fine now.
Thanks and Kind regards,
Marc.


device-types/Ivanti/PulseSecure-PSA300.yaml
```diff
@@ -0,0 +1,23 @@
+---
+manufacturer: Ivanti
+model: PulseSecure PSA300
+slug: ivanti-pulse-secure-psa300
+part_number: PSA300
+u_height: 1
```
danner26, line 6: The PSA300 looks to not be a rack mounted device. If this is the case please mark this as `u_height: 0`

device-types/Ivanti/PulseSecure-PSA300.yaml
```diff
@@ -0,0 +1,23 @@
...
+weight: 1.13
+weight_unit: kg
+airflow: front-to-rear
+comments: '[Pulse Secure Appliance Serie Data Sheet](https://secureaccessworks.com/datasheets/PulseSecure_OneSheet_PSAHardware.pdf)'
+console-ports:
+  - name: Console
+    type: rj-45
+power-ports:
+  - name: PS1
+    type: iec-60320-c14
```
danner26, line 17: This is not a C14 on board. It appears to be a `dc-terminal`. Please adjust accordingly

device-types/Ivanti/PulseSecure-PSA300.yaml
```diff
@@ -0,0 +1,23 @@
...
+    maximum_draw: 37
+interfaces:
+  - name: port1
```
danner26, line 20: Are these actually called `port1` in the devices configuration? If so then you can mark this as resolved, otherwise please make sure this matches

danner26 (changes requested):

device-types/Ivanti/PulseSecure-PSA300.yaml
mag-marc-schulz: Correct, I fixed it.
I had seen in other Device Types, which are also not rack mounted, that there is also u_height 1. I assumed that is a minimum height. But then I know for the future. Thanks.

device-types/Ivanti/PulseSecure-PSA300.yaml
mag-marc-schulz: You are right. I fixed it.

device-types/Ivanti/PulseSecure-PSA300.yaml
mag-marc-schulz: Your also right here. In GUI and XML Config the Interfaces are called "internal" and "external" on all appliances. 
Also the PSA7000 has 2 internal and external ports.
I fixed all enties.

device-types/Ivanti/PulseSecure-PSA300.yaml
danner26: Unfortunately not, this is largely a community maintained repo so if you do notice any that need to be fixed please submit a PR! It would be much appreciated!

danner26 (approved):

PR #1598, opened 2023-09-01
Added 9200(L) Cisco stack kit

deku-m:


shepherdjay:
Was working on exploring the inventory for a similar PR when I saw this one. However, this PR needs to be fixed... the `STACK-KITs` are not themselves modules. They are kits containing two stack adapters and an accompanying cable. It is the stack adapters themselves that should be defined as the module providing a single stack port.

deku-m:
> Was working on exploring the inventory for a similar PR when I saw this one. However, this PR needs to be fixed... the `STACK-KITs` are not themselves modules. They are kits containing two stack adapters and an accompanying cable. It is the stack adapters themselves that should be defined as the module providing a single stack port.

Well i am gonna close them and re-add maybe

PR #1607, opened 2023-09-06
Add NetApp C250, NSM100, NS224 devices

pv2b:
Hi!

Some more NetApp devices for the library!

Some notes. These parent devices here are disk shelves or head units, and each of these shelves will have two "controllers" sliding into them.

The power supplies do actually (surprisingly) belong to the controllers, not the chassis, and each controller has a single PSU each. Technically, the PSU is modular insofar that it can be easilly swapped out in the field without any tools, but it doesn't make sense to model it as "modular", since the power supply is required for operation, and there's only one model.


danner26:
Please correct the formatting issues and then let me know, thanks!

pv2b:
> So the power does look like it is on the device, that is fine. That being said it also looks like the PSUs are modular. Modular components should be created in `module-bays` and then the associated PSUs should be created in the module-type folder

I'm not so sure. Physically the power supplies sit inside the "carrier" the controller sits in, but I believe the power supplies power both controllers as well as the shelf itself?

It's kind of an edge case.

danner26:
It is hard to tell from the images without the devices in front of me.. Lets pose this as the question

Do the PSUs themselves slot into the child device, or do the PSUs slot into the parent chassis? They most likely power the chassis as well if it is the former.

That being asked, my point still stands. Modular components should be created in `module-bays` and then the associated PSUs should be created in the module-type folder

pv2b:
> It is hard to tell from the images without the devices in front of me.. Lets pose this as the question
> 
> Do the PSUs themselves slot into the child device, or do the PSUs slot into the parent chassis? They most likely power the chassis as well if it is the former.
> 
> That being asked, my point still stands. Modular components should be created in `module-bays` and then the associated PSUs should be created in the module-type folder

Yeah, I agree, they should be modular. I'll have to see if I can dig up the part number for the AC power supply at least, I don't have a DC one.

I've never replaced one of these power supplies with my own hands ,but looking at the documentation links and the associated videos, it seems that the power supply can be removed without removing the controller (child device), but when you remove the controller, that takes the power supply with it because the power supply "holder" is part of the controller. However, I then believe that the backplane allows the child devices and the parent device to share those two redundant power feeds.

PR #1608, opened 2023-09-06
Module Test Improvements

danner26:
In reference to issue #1584 we have discovered a missing test case, it is now implemented and the affected files are fixed

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,24 +1,24 @@
 ---
 files: ^(device-types|module-types)/
 repos:
-  # - repo: https://github.com/pre-commit/pre-commit-hooks
-  #   rev: v4.4.0
-  #   hooks:
-  #     - id: trailing-whitespace
-  #     - id: end-of-file-fixer
-  #     - id: check-yaml
-  #       name: Check YAML files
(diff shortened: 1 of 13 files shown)


danner26:
Closes #1585 
Closes #1584

DanSheps:
@danner26 Was busy yesterday and missed this.

I don't think this is a valid test, as it is perfectly reasonable (I do it often) to have same positions as long as they generate different component names.

For example:

Mod Bay 1:
  Name: PSU1
  Position: 1
Mod Bay 2:
  Name: PSU2
  Position: 2
Mod Bay 3:
  Name: FRU Uplink
  Position: 1

What we want is to be able to evaluate the final interface/power port/etc name and not the position and whether it is duplicated (which admittedly is not possible).  I think our tests for something like this should follow the upstream repo unique keys.

danner26:
The test is to check for the case where PSU1 and PSU2 both share the same position

An `FRU Uplink` and `PSU1` with the same `position: 1` is valid and will pass the test case. Let me know if I am misunderstanding you

DanSheps:
I didn't fully understand the code is I think the problem.  If it passes then all good IMO.

PR #1613, opened 2023-09-07
adding LC 96 fiber patch panel

dejantep:


tests/known-modules.pickle
DanSheps, line 1: Pretty sure we don't want the pickle files included

tests/known-slugs.pickle
DanSheps, line 1: Same as above

DanSheps (changes requested):
Is this simplex 96 or duplex 96?  Seems like a lot of ports.

Also, is it actually generic or is there a specific manufacturer and you are just genericising it?

tests/known-modules.pickle
dejantep: true but i'm not sure what i'm doing wrong

dejantep: its generic and every port has its own number 1-96 so simplex is possible
its known as LC quad something....

dejantep: image

dejantep (commented):
ok. remove the files

DanSheps (approved):
Looks good from my end, but I only took a quick look.  I will let @danner26 do a more in-depth dive.  😄

dejantep:
is there anything i can do to help to merge this?

danner26:
Please remove the pickle files from your PR. These should not be included at any time

PR #1615, opened 2023-09-10
Add Alcatel-Lucent OS6560-P48Z16 and OS6860-P48

ststereno401:
added devices for manufacturer Alcatel-Lucent


danner26:
Do you still want this to be added?

PR #1618, opened 2023-09-12
Create Infoblox IB-1425.yaml

Lon1:
add new IB-1425 appliance


device-types/Infoblox/IB-1425.yaml
```diff
@@ -0,0 +1,27 @@
+---
+manufacturer: Infoblox
+model: IB-1425
+slug: infoblox-ib-1425
+is_full_depth: false
+u_height: 1
```
danner26, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

1. Please add the `part_number` if you have it
2. Please add the `airflow` data point if you have it or can find it. Often these are listed in the datasheet, but sometimes they are not
3. Please add the `weight` and associated `weight_unit` data point if you have it or can find it. Often these are listed in the datasheet, but sometimes they are not

danner26 (changes requested):

PR #1622, opened 2023-09-14
Create Infoblox IB-4010.yaml and add description to schema

Lon1:
Create new Infoblox IB-4010 device type and also added description to schema which was fixed in 2.11, see https://github.com/netbox-community/netbox/issues/6438


schema/devicetype.json
```diff
@@ -14,6 +14,9 @@
         "part_number": {
             "type": "string"
         },
+        "description": {
+            "type": "string"
+        },
```
danner26, line 19: These files should not be modified. Please remove this

device-types/Infoblox/IB-4010.yaml
```diff
@@ -0,0 +1,31 @@
+---
+manufacturer: Infoblox
+model: IB-4010
+slug: infoblox-ib-4010
+is_full_depth: false
+u_height: 2
+weight: 61
```
danner26, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

device-types/Infoblox/IB-4010.yaml
```diff
@@ -0,0 +1,31 @@
...
+    type: 1000base-t
+  - name: HA
+    type: 1000base-t
+  - name: LOM
+    type: 1000base-t
+console-ports:
+  - name: Serial
+    type: de-9
+  - name: VGA
+    type: other
```
danner26, line 24: VGA is not a console port

danner26 (changes requested):

PR #1627, opened 2023-09-15
Create Infoblox IB-825

Lon1:


device-types/Infoblox/IB-825.yaml
```diff
@@ -0,0 +1,24 @@
+---
+manufacturer: Infoblox
+model: IB-825
+slug: infoblox-ib-825
+is_full_depth: false
```
danner26, line 5: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

1. Please add the `airflow` data point if you have it or can find it. Often these are listed in the datasheet, but sometimes they are not. I believe the airflow is `front-to-rear`
2. Please add the `weight` and associated `weight_unit` data point if you have it or can find it. Often these are listed in the datasheet, but sometimes they are not. I believe the weight is `7.71 kg`
3. Please add the manufacturer `part_number` which I believe is `TE-825`

device-types/Infoblox/IB-825.yaml
```diff
@@ -0,0 +1,24 @@
+---
+manufacturer: Infoblox
+model: IB-825
```
danner26, line 3: Is the device actually called `Trinzic 825`?

danner26 (changes requested):

device-types/Infoblox/IB-825.yaml
Lon1: Trinzic would be the original name of the company from long ago before it was Infoblox.  I go by what our NetMRI tool discovers it as (shown below).  NetMRI is a Network Management platform from Infoblox.
image

Lon1: Daniel - I took a look at what our actual product sheet said for these devices and it uses TE-805-HW-AC.  That is the base hardware.  The '2' (in the 825) is addition software options.  Splitting hairs here, but wanted to be fair on what direction you wanted to go with this.

image

danner26: So the hardware is what we are modelling here. I think it is best to make the part number equal what the hardware is and then have the model number reflect IB-825

Lon1: I would agree to that.  I'll try and update the others with the part number as well next week.

device-types/Infoblox/IB-825.yaml
Lon1: added requested fields.

Lon1 (commented):
added requested fields

danner26 (approved):

PR #1628, opened 2023-09-18
Add Cisco C9200L-24PXG-2Y

okoelbl:
Add Cisco C9200L-24PXG-2Y switch

### DIFF
--- a/device-types/Cisco/C9200L-24PXG-2Y.yaml
+++ b/device-types/Cisco/C9200L-24PXG-2Y.yaml
@@ -1,0 +1,128 @@
+---
+manufacturer: Cisco
+model: C9200L-24PXG-2Y
+part_number: C9200L-24PXG-2Y
+slug: cisco-c9200l-24pxg-2y
+u_height: 1
+is_full_depth: false
+comments: '[Cisco Catalyst 9200 Series Switches Data Sheet](https://www.cisco.com/c/en/us/products/collateral/switches/catalyst-9200-series-switches/nb-06-cat9200-ser-data-sheet-cte-en.html)'
+console-ports:
+  - name: Console
+    type: rj-45
+  - name: usb
+    type: usb-mini-b
+power-ports:
+  - name: PS-1
+    type: iec-60320-c14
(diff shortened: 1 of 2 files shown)


device-types/Cisco/C9200L-24PXG-2Y.yaml
```diff
@@ -0,0 +1,128 @@
...
+    type: rj-45
+  - name: usb
+    type: usb-mini-b
+power-ports:
+  - name: PS-1
+    type: iec-60320-c14
+    maximum_draw: 600
+  - name: PS-2
+    type: iec-60320-c14
+    maximum_draw: 600
```
danner26, line 20: These PSU's are modular and thus you should define `module-bays` with 2 slots. Then, the applicable PSU modules should be added to the module-types folder if they do not already exist

device-types/Cisco/C9200L-24PXG-2Y.yaml
```diff
@@ -0,0 +1,128 @@
+---
+manufacturer: Cisco
+model: C9200L-24PXG-2Y
+part_number: C9200L-24PXG-2Y
+slug: cisco-c9200l-24pxg-2y
+u_height: 1
+is_full_depth: false
```
danner26, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

danner26 (changes requested):

device-types/Cisco/C9200L-24PXG-2Y.yaml
okoelbl: Hello and thank you for reviewing! I've added the requested data points and module bays, as well as the supported (AC) PSU.

device-types/Cisco/C9200L-24PXG-2Y.yaml
okoelbl: Thanks! Added module bays (see other reply)

danner26 (approved):

PR #1639, opened 2023-09-27
Synology RS2416RP+

sulik76:


device-types/Synology/RS2416RP+.yaml
```diff
@@ -0,0 +1,37 @@
...
+    mgmt_only: true
+power-ports:
+  - name: PSU 1
+    type: iec-60320-c14
+    maximum_draw: 120
+    allocated_draw: 56
+  - name: PSU 2
+    type: iec-60320-c14
+    maximum_draw: 120
+    allocated_draw: 56
```
danner26, line 26: These PSUs are modular, so they should be modelled under `module-bays` and the associated PSUs themselves should be modelled in the module-types folder if they do not already exist

danner26 (changes requested):

danner26 (approved):

PR #1640, opened 2023-09-28
QNAP TS-251+

madecto:


madecto:
Goodday @danner26, I'm struggling to get this PR work: it appears new enforcements are in place for device `slug` and filenames and not on `model` and `part_number`: as this Model and Part Number both include the _forbidden_ character "+" the verification fails.
I could swap the "plus" string to the "+" sign, but when I look up a device type I would personally much prefer to find the vendor grammar rather than an alternative version.

abhi1693 (changes requested):
Hello, it appears that you have modified the `tests/known-slugs.pickle` file. This file is automatically generated via a GitHub Action, which contains all currently available slugs. This file should **never** be commited by a contributor in a PR.
  
In order to fix this, you will need to remove the modification on the pickle file. There are a handful of ways to do so but often it is easiest to revert the commit and force push it again without the pickle file included. If you need assistance you are welcome to ping a maintainer via this PR, or via our NetDev Slack.

device-types/QNAP/TS-251-plus.yml
```diff
@@ -0,0 +1,26 @@
+---
+manufacturer: QNAP
+model: TS-251+
+slug: qnap-ts-251-plus
+part_number: TS-251+
+is_full_depth: false
+weight: 1.28
```
danner26, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

danner26 (changes requested):

device-types/QNAP/TS-251-plus.yml
madecto: Hello @danner26,
I've managed to grab this from a real-life example.

madecto: Hello @danner26,
I've managed to grab this from a real-life example.

PR #1648, opened 2023-10-02
Update dell

Stetsed:
I updated the R420 and the R530 templates to add there PCIe slots. The R530 only has the slots it has by default, not the riser as the R530 can function without a riser while the R420 will just not have PCIe slots.


Stetsed:
I do not see why it's getting failed, can't pull it out of the error logs.

danner26:
`pre-commit` is a powerful command line tool that assists with finding these errors. It also is the same tool we use for testing and itll help you correct errors automatically sometimes as well.

That being said, the issue is that you have quotes around string text that is not needed for both the R420 and the R530

danner26 (approved):

PR #1664, opened 2023-10-06
HPE-5710-and-modules

francesco-noc:


danner26:
HPE drives me insane. I cannot believe it is actually called `FlexFabric 5710 24SFP+ 6QSFP+ or 2QSFP28`

danner26 (approved):

PR #1667, opened 2023-10-06
Rename FortiSwitch-224E-POE.yaml to FS-224E-POE.yaml

mihailim:
We currently have a stray `FortiSwitch-224E-POE.yaml` file in `device-types/Fortinet` which doesn't match the file naming convention. Should be renamed to `FS-224E-POE.yaml`.

It's also partially incorrect due to missing PoE stanzas on ports 1-12, will submit a separate PR for that (couldn't figure out how to do both simultaneously while keeping the pre-commit hooks active.)

Evidently, this fails the pre-commit checks since they don't seem to handle file renames and complain about a duplicate slug.

By the way, since the PR is beyond trivial, I'd also be perfectly happy if nobody bothered with the PR and just renamed the offending file :)

Thanks!


danner26:
Thanks again for this! It was a great test case that was missed originally

PR #1690, opened 2023-10-18
Added APC AP7841 and AP8841 PDUs

bberberov:
Added two single phase PDUs

### DIFF
--- a/device-types/APC/AP7841.yaml
+++ b/device-types/APC/AP7841.yaml
@@ -19,7 +19,7 @@ console-ports:
   - name: Serial
     label: Serial Port
     type: rj-12
-interface:
+interfaces:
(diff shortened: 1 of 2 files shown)


bberberov:
Is `description` not supposed to be in model definitions?  It's supported in V3.6.1.

ArmanEsf:
> Is `description` not supposed to be in model definitions? It's supported in V3.6.1.

No, it's not supposed to be there, the description field in the UI is the comments in the YAML file, but you have both.
Could you try to put that info in the comments field or delete the description value?

danner26 (approved):

danner26:
Hello @bberberov please resolve the conflict and then mark this PR as ready for review

PR #1691, opened 2023-10-18
Added APC AP8881 and AP8888 PDUs

bberberov:
Added two 3-phase PDUs


bberberov:
Is `description` not supposed to be in model definitions?  It's supported in V3.6.1.

danner26 (approved):

PR #1696, opened 2023-10-20
Add Cisco C9300X-12Y

haakon-ranum-nrk:


device-types/Cisco/C9300X-12Y.yaml
```diff
@@ -0,0 +1,55 @@
+---
+manufacturer: Cisco
+model: Catalyst 9300X-12Y
+part_number: C9300X-12Y
+slug: cisco-c9300x-12y
+u_height: 1
+is_full_depth: true
```
danner26, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

danner26 (changes requested):

PR #1700, opened 2023-10-23
jl086a

francesco-noc:


danner26:
Please check how the rest of the Aruba line is modelled and follow that. The file name is the first problem to be fixed (should be Aruba-MODELNAME). Thank you

PR #1701, opened 2023-10-23
Hpe aruba (JL806A, JL807A, JL686A)

vittore:
Added:
- JL806A
- JL807A
- JL686A


danner26:
Please check how the rest of the Aruba line is modelled, and follow that. Your file names are incorrect which is the first problem we need fixed. Thank you

vittore:
Please don't add: switches are already present with the correct names.

PR #1713, opened 2023-10-25
Huawei

jronnblom:
New hardware. CE16804, S6730-H24X4Y4C and CE6865-48S8CQ-EI


danner26:
Before I review these please make sure you have `weight` and `weight_unit` as well as `airflow` defined where you can find the data. Most of the time you can find this info in the datasheet

Please mark this PR as ready to review once you get around to doing that

jronnblom:
@danner26 The MPU cards have both RJ45 console and Meth console (same name). Not sure how to model this. I tried adding {module} to the RJ45 console. Which gives med Console5. That kinda works since its in slot 5. Not sure about Meth 0/0/0 though

jronnblom:
> Will there only ever be one `Meth0/0/0`? If so, then this wont be an issue.. But if there could be another then we need to figure out how it assigns the numbers, and use the same funcationality to ensure the number is assigned correctly to the power

There will be only one in the configuration. However there are 1 nic on every MPU card so in reality there are 2 physical nics which both are Meth 0/0/0 depending on which one is active. How do we model this?

danner26:
You cannot have more than one interface with the same name on the same device. Can you show an output from the console of this being the case?

jronnblom:
> You cannot have more than one interface with the same name on the same device. Can you show an output from the console of this being the case?

On the CE16800 series its possible. They have redundant cpu cards where one is in slave in standby.

`
<coresw>display current-configuration interface MEth 
#
interface MEth0/0/0
 ip binding vpn-instance xxx
 ip address a.b.c.d 255.255.255.0
#
return
<coresw>
<coresw>display device slot 5                        
CE16804's Device status:
------------------------------------------------------------------------------------------
Slot  Card   Type                     Online   Power Register     Alarm     Primary        
------------------------------------------------------------------------------------------
5     -      CE-MPUD-HALF             Present  On    Registered   Normal    Master         
------------------------------------------------------------------------------------------
  Part  Number      : 03058876
  Board Type        : CE-MPUD-HALF
  Board Description : CE16800 Main Processing Unit D(half-width)
-------------------------------------------------------------------------------------------

-------------------------------------------------------------------------------
Interface          Port     Optic   MDI    Speed  Duplex Input   Output  Port  
                   Type     Status         (Mbps)        Flow-   Flow-   Status
                                                         control control       
-------------------------------------------------------------------------------
MEth0/0/0          Copper   Absent  --     1000   Full   Disable Disable Up    
Sip5/0/0           Fiber    Absent  --     10000  Full   Disable Disable Down  
Sip5/0/1           Fiber    Absent  --     10000  Full   Disable Disable Down  
-------------------------------------------------------------------------------
<coresw> 
<coresw>display device slot 6                        
CE16804's Device status:
------------------------------------------------------------------------------------------
Slot  Card   Type                     Online   Power Register     Alarm     Primary        
------------------------------------------------------------------------------------------
6     -      CE-MPUD-HALF             Present  On    Registered   Normal    Slave          
------------------------------------------------------------------------------------------
  Part  Number      : 03058876
  Board Type        : CE-MPUD-HALF
  Board Description : CE16800 Main Processing Unit D(half-width)
-------------------------------------------------------------------------------------------

-------------------------------------------------------------------------------
Interface          Port     Optic   MDI    Speed  Duplex Input   Output  Port  
                   Type     Status         (Mbps)        Flow-   Flow-   Status
                                                         control control       
-------------------------------------------------------------------------------
MEth0/0/0          Copper   Absent  --     1000   Full   Disable Disable Up    
Sip6/0/0           Fiber    Absent  --     10000  Full   Disable Disable Down  
Sip6/0/1           Fiber    Absent  --     10000  Full   Disable Disable Down  
-------------------------------------------------------------------------------
<coresw>
`

PR #1716, opened 2023-10-26
Updated Dell M1000e, and added M630 and two network modules

bberberov:


bberberov:
I have no idea what the error is?!

device-types/Dell/PowerEdge-M1000e.yaml
```diff
@@ -4,34 +4,74 @@ model: PowerEdge M1000e
...
     type: iec-60320-c20
-  - name: PSU4
+  - name: PS-4
     type: iec-60320-c20
-  - name: PSU5
+  - name: PS-5
     type: iec-60320-c20
-  - name: PSU6
+  - name: PS-6
     type: iec-60320-c20
```
danner26, line 46: Please correct me if I am wrong, but it looks like the PSUs are modular. If that is true, these should be defined as `module-bays` and the associated PSUs should be created in the module-types folder

device-types/Dell/PowerEdge-M1000e.yaml
```diff
@@ -4,34 +4,74 @@ model: PowerEdge M1000e
...
+  - name: IOM Slot A2 management backplane
+    type: 1000base-kx
+  - name: IOM Slot B1 management backplane
+    type: 1000base-kx
+  - name: IOM Slot B2 management backplane
+    type: 1000base-kx
+  - name: IOM Slot C1 management backplane
+    type: 1000base-kx
+  - name: IOM Slot C2 management backplane
+    type: 1000base-kx
```
danner26, line 30: Are these actually built into the chassis or are they modular as well?

danner26:
I also fixed your validation errors. `pre-commit` is a useful tool that you can install alongside your repo in a python venv. It will correct the definitions and tell you if there are errors even before you push them to GitHub

PR #1717, opened 2023-10-26
Updated Dell N2048 and N3048

bberberov:
- weight
- power supplies
- management port
- comment

### DIFF
--- a/device-types/Dell/PowerSwitch-N3048.yaml
+++ b/device-types/Dell/PowerSwitch-N3048.yaml
@@ -5,13 +5,23 @@ slug: dell-powerswitch-n3048
 part_number: N3048
 u_height: 1
 is_full_depth: true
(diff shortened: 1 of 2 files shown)


device-types/Dell/PowerSwitch-N3048.yaml
```diff
@@ -5,13 +5,23 @@ slug: dell-powerswitch-n3048
...
   - name: Console
     type: rj-45
 power-ports:
-  - name: PS1
+  - name: PS-1
     type: iec-60320-c14
+    maximum_draw: 75
+  - name: PS-2
+    type: iec-60320-c14
+    maximum_draw: 75
```
danner26, line 24: Please correct me if I am wrong, but it looks like the PSUs are modular. If that is true, these should be defined as `module-bays` and the associated PSUs should be created in the module-types folder

danner26 (changes requested):

PR #1727, opened 2023-10-30
Add D-Link DXS-3400-24TC

rickypid:


device-types/D-Link/DXS-3400-24TC.yaml
```diff
@@ -0,0 +1,88 @@
...
+    type: rj-45
+  - name: usb0
+    type: usb-mini-b
+power-ports:
+  - name: PS1
+    type: iec-60320-c14
+    maximum_draw: 160
+  - name: PS2
+    type: iec-60320-c14
+    maximum_draw: 160
```
danner26, line 20: Please correct me if I am wrong, but it looks like the PSUs are modular. If that is true, these should be defined as `module-bays` and the associated PSUs should be created in the module-types folder

device-types/D-Link/DXS-3400-24TC.yaml
```diff
@@ -0,0 +1,88 @@
+---
+manufacturer: D-Link
+model: DXS-3400-24TC
+slug: d-link-dxs-3400-24tc
+part_number: DXS-3400-24TC
+comments: '[D-Link DXS-3400-24TC](https://eu.dlink.com/it/it/products/dxs-3400-series-top-of-rack-10-gigabit-stackable-managed-switches)'
```
danner26, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

danner26 (changes requested):

PR #1729, opened 2023-10-31
Fixed Mellanox SX6036

bberberov:
- Default port mode is IB FDR
- PSUs are PS1 and PS2 in OS

Similar to SX6012, which we still have

### DIFF
--- a/device-types/Mellanox/SX6036.yml
+++ b/device-types/Mellanox/SX6036.yml
@@ -8,10 +8,10 @@ console-ports:
   - name: Console
     type: rj-45
 power-ports:
-  - name: psu1
+  - name: PS1
     type: iec-60320-c14
     maximum_draw: 500
-  - name: psu2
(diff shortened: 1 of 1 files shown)


device-types/Mellanox/SX6036.yml
```diff
@@ -8,10 +8,10 @@ console-ports:
...
     type: rj-45
 power-ports:
-  - name: psu1
+  - name: PS1
     type: iec-60320-c14
     maximum_draw: 500
-  - name: psu2
+  - name: PS2
     type: iec-60320-c14
     maximum_draw: 500
```
danner26, line 16: Please correct me if I am wrong, but it looks like the PSUs are modular. If that is true, these should be defined as `module-bays` and the associated PSUs should be created in the module-types folder

danner26 (changes requested):

PR #1730, opened 2023-10-31
Added Mellanox SX6012, all versions, and mounting rails

bberberov:


bberberov:
> Failed: device-types/Mellanox/SX60-DKIT.yaml has does not appear to have a valid power source. Ensure either "power-ports" or "interfaces" with "poe_mode" is defined.

`SX60-DKIT` is the rail kit that comes with the SX6012.  It cannot have "power-ports" or "interfaces", but it's a `device` because it has `device_bays`

danner26 (approved):

PR #1731, opened 2023-10-31
Expanded definition of Mellanox SB7800, added SB7890

bberberov:


device-types/Mellanox/SB7890.yml
```diff
@@ -0,0 +1,94 @@
...
+
+  - [Product Brief](https://network.nvidia.com/files/doc-2020/pb-sb7890.pdf) (PDF)
+  - [Specifications](https://docs.nvidia.com/networking/display/sb77x0edr/specifications)
+power-ports:
+  - name: PS1
+    type: iec-60320-c14
+    maximum_draw: 332 # Max power with optical cables (assuming 3.5W per port): 332W / PSU rating: 460W
+  - name: PS2
+    type: iec-60320-c14
+    maximum_draw: 332 # Max power with optical cables (assuming 3.5W per port): 332W / PSU rating: 460W
```
danner26, line 21: Please correct me if I am wrong, but it looks like the PSUs are modular. If that is true, these should be defined as `module-bays` and the associated PSUs should be created in the module-types folder

PR #1742, opened 2023-11-07
Addition of Fortinet FG-400F device and images

n00dl3st:
Submit new Fortinet device to libaray. [FG-400F](https://www.fortinet.com/content/dam/fortinet/assets/data-sheets/fortigate-400f-series.pdf)


device-types/Fortinet/FG-400F.yaml
```diff
@@ -0,0 +1,95 @@
...
+    type: rj-45
+power-ports:
+  - name: PS1
+    type: iec-60320-c14
+    maximum_draw: 214
+    allocated_draw: 109
+  - name: PS2
+    type: iec-60320-c14
+    maximum_draw: 214
+    allocated_draw: 109
```
danner26, line 25: These PSUs are modular. Thus you should create `module-bays` on this device, and then add the associated PSUs to the module-types folder if they do not already exist

danner26 (changes requested):

danner26 (approved):

PR #1755, opened 2023-11-09
Create RS822RP+.yaml

pierrickj:


pierrickj:
Add Synology RS822RP+ NAS Model

device-types/Synology/RS822RP+.yaml
```diff
@@ -0,0 +1,30 @@
...
+console-ports:
+  - name: Console
+    type: de-9
+power-ports:
+  - name: PSU 1
+    type: iec-60320-c14
+    label: PSU 1
+  - name: PSU 2
+    type: iec-60320-c14
+    label: PSU 2
```
danner26, line 17: The power for this chassis appears to be modular, is that correct?
If so, you should create `module-bays` on this device and then add the associated PSU definitions to the module-types folder

device-types/Synology/RS822RP+.yaml
```diff
@@ -0,0 +1,30 @@
+---
+manufacturer: Synology
+model: RS822RP+
+slug: synology-rs822rp-plus
+part_number: RS822RP+
+u_height: 1.0
+is_full_depth: true
```
danner26, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

danner26 (changes requested):

PR #1760, opened 2023-11-13
Synology rx1217rp

pierrickj:


device-types/Synology/RX1217RP.yaml
```diff
@@ -0,0 +1,18 @@
...
+u_height: 1.0
+is_full_depth: true
+weight: 13.4
+weight_unit: kg
+comments: Expansion Unit RX1217 with Redundant Power Supplies
+power-ports:
+  - name: PSU 1
+    type: iec-60320-c14
+  - name: PSU 2
+    type: iec-60320-c14
```
danner26, line 15: The power for this chassis appears to be modular, is that correct?
If so, you should create `module-bays` on this device and then add the associated PSU definitions to the module-types folder

danner26 (changes requested):

device-types/Synology/RX1217RP.yaml
pierrickj: ok, changes have been made

danner26:
I didnt realize this at first, but there is no networking for this device. Currently storage is out of scope from this repo since the upstream NetBox repo does not directly support storage management.. we will have to close this PR until storage management becomes a thing. Sorry about that

PR #1774, opened 2023-11-20
Added Aruba 3810M & Cisco A9K-24X10GE-TR 

chrisbish:
Added: Aruba 3810M 16SFP+ 2-slot Switch Cisco A9K-24X10GE-TR Cisco SG350X-08


module-types/Cisco/A9K-24X10GE-TR.yaml
danner26, line 1: Generally modules use the `{module}` definition in interface names so that they are assigned the proper name when slotted into a device

i.e. `TenGigabitEthernet1/{module}/1`

danner26 (changes requested):

PR #1775, opened 2023-11-21
Balckmagic devices

Ivarrrrr:


danner26:
I noticed that the majority of the interfaces on these devices are audio/video. We dont really model those at the moment. I am waiting for some other maintainers to chip in on if this is out of scope for this repo or not

danner26:
@Ivarrrrr it does not look like you reviewed/fixed my comments. Please do so before marking ready for review

PR #1776, opened 2023-11-21
More (cisco) models

steffenschumacher:
Added: C8300-1N1S-6T, C8300-2N2S-4T2X, ISR-C1111-4P, IE-3200-8P2S & C891F-K9,


steffenschumacher:
Added in some artefact - doing a fresh PR

PR #1779, opened 2023-11-22
Arista Elevation Set

colinmacgiolla:
Adding a fairly complete set of elevations for Arista devices.
Note - I've deleted the jpg files in the previous PR, and converted everything to PNG as the preferred format


danner26 (approved):
Holy moley that way a ton! Thanks for all of your work on this

PR #1786, opened 2023-11-28
Add Supermicro 1018R-WC0R

kennethdammyr:
Shuold probably add a ton of SuperMicro servers, here is one.


device-types/Supermicro/SuperServer-1018R-WC0R.yml
```diff
@@ -0,0 +1,28 @@
...
+  - name: BMC
+    type: 1000base-t
+    mgmt_only: true
+power-ports:
+  - name: Power 1
+    type: iec-60320-c14
+    maximum_draw: 700
+  - name: Power 2
+    type: iec-60320-c14
+    maximum_draw: 700
```
danner26, line 28: The power on these are modular. You should move the device definition to `module-bays` and then create the associated PSUs in the module-types folder if they do not already exist

danner26 (changes requested):

device-types/Supermicro/SuperServer-1018R-WC0R.yml
kennethdammyr: Hi @danner26!

As far as I can see, that is not common practice on any other servers or network gear. Should we start doing this now?

danner26: Hello, this repo has been around since the beginning of NetBox, before modular types existed. That being said, it is absolutely widely used in this repo (search the code base for `module-bays` and see for yourself if you disagree). That being said, since module-types have been implemented we do require it going forward for new and modified device PRs.

The point of this repo is to be as accurate to the real life device as possible. If the device has modular items then they should be modelled that way. With that said, there are still a lot of legacy files which were created before `modular-bays` existed which is why you do not see it in every device. Since this is a community driven repo we rely on the community to update the definitions if they are so inclined to.

kennethdammyr:
@danner26: Should be okay now :)

danner26 (approved):

PR #1790, opened 2023-11-30
Add Dell PowerSwitch S5212F-ON and shelf

elipsion:


danner26:
Can you link to this shelf? I cant find it anywhere. If it is just a 1U shelf you are trying to model then this is not the something we should be adding. The only device you need to model is the switch itself

elipsion:
A-ha! [They exist!](https://www.dell.com/support/manuals/en-us/force10-s3048-on/rail_guide_pub/trays?guid=guid-6ce6d1d1-aaed-42a9-9d93-4df581bfd323&lang=en-us)

danner26:
Oh awesome, you can add them to this PR and rename it

PR #1804, opened 2023-12-11
add Cisco IE 3000 series switches

abu-rehan:
Added three templates from Cisco IE 3000 series.


danner26 (changes requested):

danner26:
Device-bays are used when devices physically slot into another device. It does not appear this is the case. Can you please clarify this?

abu-rehan:
> Device-bays are used when devices physically slot into another device. It does not appear this is the case. Can you please clarify this?
The devices have to be physically combined. The child device cannot be a standalone device. Device Bay, we thought would be the most appropriate way of modelling it. If it doesn't look right, you can close this pull request. Also if you have encountered IE 3000 before in netbox, do you have tips on how to model them?

DanSheps:
These are modules, not devices and should be modelled using module bays.

Are you not using 3.6 for NetBox?

abu-rehan:
Yes we are using netbox 3.6 and above. We are closing this PR for now. We will try out these devices as modules and see if that works fine.

PR #1807, opened 2023-12-14
Reichle & De-Massari modular fiber panels

jhofmueller:
Add one 1U 19" empty panel and three modules from *Reichle & De-Massari*.


DanSheps:
I don't know if I agree with modelling individual strands going to a splice as a single bundle.  It depends on the use case of course but I think t is better that they are modelled as single rear ports for each strand.

jhofmueller:
Each strand has it's own rear port now for both fiber modules.

DanSheps (approved):

jhofmueller:
This and pr 1817 are ready to be merged AFAICS.  Please advise if there is something I neglected to do.

PR #1814, opened 2023-12-18
Update frontend images to higher resolution

jbemmel:


abhi1693 (changes requested):
Better resolution is a good feature, but I'll suggest to compress all images using lossless algorithm to save network bandwidth when these are viewed by the user. There is a potential savings of 65% for each image here.

danner26:
Hey @jbemmel hope all has been well. Glad to see you are still around.

Could you do me a big favor and remove the extra whitespace surrounding some of those images? It will appear much better that way

danner26 (approved):

PR #1823, opened 2023-12-21
Adds qnap device types

abhi1693:


device-types/QNAP/TS-531X.yml
danner26, line 1: It looks like there are 2 actual part numbers, which is this for?

`TS-531X-2G` and `TS-531X-8G`

device-types/QNAP/TS-563.yml
```diff
@@ -0,0 +1,22 @@
+---
+manufacturer: QNAP
+model: TS-563
+slug: qnap-ts-563
+part_number: TS-563
+u_height: 2
+is_full_depth: false
```
danner26, line 7: 1. Please add the `weight` and associated `weight_unit` data point if you have it or can find it. This appears to be listed in the datasheet (9.7 lbs or 4.4 kg)

device-types/QNAP/TS-832-PXU-RP.yml
```diff
@@ -0,0 +1,40 @@
...
+weight_unit: kg
+device-bays:
+  - name: Disk 1
+  - name: Disk 2
+  - name: Disk 3
+  - name: Disk 4
+  - name: Disk 5
+  - name: Disk 6
+  - name: Disk 7
+  - name: Disk 8
```
danner26, line 22: Right now these are not supported, please remove the `device-bays`

danner26 (changes requested):

PR #1826, opened 2023-12-21
Add WatchGuard Firebox M270

rickypid:


device-types/WatchGuard/M270.yaml
```diff
@@ -0,0 +1,37 @@
+---
+manufacturer: WatchGuard
+model: M270
+slug: watchguard-firebox-m270
+part_number: M270
+is_full_depth: false
+u_height: 1
+weight: 8.5
+weight_unit: kg
```
danner26, line 9: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

danner26 (changes requested):

danner26 (approved):

PR #1829, opened 2023-12-21
Added cisco module types

abhi1693:


DanSheps (commented):
I can't speak inteligently about the UCS stuff, but the Nexus 7K stuff looks spot on.

PR #1835, opened 2023-12-21
Added misc module types

abhi1693:


danner26:
While this is great, I don't think these can be added yet. Especially not the HDD/SSD ones. Jeremy suggested not to do these until we expand the upstream NetBox to model storage and these other components.

PR #1848, opened 2024-01-05
Adding Nokia 7750 SR-1 device type

geg347:
Hello,

First happy new year!

Based on alimsado (Thanks to him) other PR for SR1s I quickly did this Yaml for the device type Nokia 7750 SR-1.
I am not 100% sure of the ports as this device can have 2 linecards of 6 or 12 ports (here I put 1 of 12 ports as standard ports). Should I set 2 module_bays for that?

I wasn't able to find the Part Number on the data sheet I get from Nokia's website...

Thanks for reviewing this and feel free to comment for any changes / improvements that can be done.


tardoe:
@geg347 I would leave the two slots unpopulated and no interfaces defined. Or, I'd model the MDAs and IOMs as modules so that you can model the different MDAs that can go into the two slots and thus track their serials / PNs separately.

geg347:
Hello @tardoe,
OK I deleted the interfaces and set 2 new module types for a 6 and 12 QSFP28 ports MDA.
I don't know if there is something more to do with the IOM you mentionned?
I still couldn't find the part numbers. I can fill them later when I'll have access to hardware.

geg347:
I tried with the module variable and it worked for a single card in SR-1.

But I still have the problem with the card number for bigger devices...

geg347:
I may have found a work around but I don't know if it is the best solution...
If I put the variable as {module}/cX and I set the position on module bays linked to a device to 1/1 or 1/2 or 2/1, ...
It is working.

tardoe:
> I may have found a work around but I don't know if it is the best solution... If I put the variable as {module}/cX and I set the position on module bays linked to a device to 1/1 or 1/2 or 2/1, ... It is working.

So this is pretty much the approach.

So the port naming is `card/mda/connector` - when you create the module bay, you name the first with `1/1` and then `1/2` because the SR-1 is a fixed-slot box with 2x MDAs. When you create the module-type, you'll want to do exactly as you did and use the `{module}` variable to populate the first half of the port name into the module.

For multi-slot devices, I suspect the best approach would be to create a module_bay for each MDA.

tardoe:
Part numbers for the MDAs:
E6 = 3HE12333AA
E12 = 3HE12334AA

geg347:
Perfect thank you for your help and the part numbers, I am pushing my changes.
I just miss the part number for the device SR-1 and we are good to go I think.

geg347:
One last thing maybe. Should I add a comment somewhere to say that the position variable on the module bay needs to be in the form x/y with x = card number and y mda number?

tardoe:
@geg347 I made a PR against your fork. Merge that it should come in here. I can't push to a PR here as I'm not a maintainer.

geg347:
Perfect thanks for the MDA module bays with right position, part number and images! I merged and it passes the pipeline.
It is good to merge in master then.

elevation-images/Nokia/nokia-7750-sr-1.rear.png
danner26, line 1: Please try to find a straight on image, and remove extra white space

elevation-images/Nokia/nokia-7750-sr-1.front.png
danner26, line 1: Please remove the extra white space

device-types/Nokia/7750-SR-1.yaml
```diff
@@ -0,0 +1,45 @@
...
+front_image: true
+rear_image: true
+console-ports:
+  - name: Console
+    type: rj-45
+power-ports:
+  - name: PS1
+    type: iec-60320-c14
+  - name: PS2
+    type: iec-60320-c14
```
danner26, line 20: Are these PSU's modular? If so, they should be modelled as `module-bays` on this device, and the associated PSU modules should be created in the module-types folder if they do not already exist

If not, just let me know and mark this resolved

danner26 (changes requested):

elevation-images/Nokia/nokia-7750-sr-1.rear.png
geg347: Done but should I remove the top of the device? (That causes extra white spaces on top corners).

elevation-images/Nokia/nokia-7750-sr-1.front.png
geg347: Done

device-types/Nokia/7750-SR-1.yaml
geg347: Yes you are right. I will add a module-bay for this (based on official documentation). But again I don't know where to find the part numbers, sorry.

geg347:
Also I found out that for other Nokia devices the Fan module bays are also defined. Should I do the same?

PR #1856, opened 2024-01-13
Add BDCOM-S2900-48P6X-PoE.yaml

aleferrariuy:
BDCOM brand switch. Originally from China.
With L3 capabilities.
48 GE ports.
6 bays for optical modules up to 10Gbps.
PoE capability on all 48 ports.


aleferrariuy (commented):
I modified syntax errors

PR #1859, opened 2024-01-15
Add the X440-G2-48p/t-10GE4 and EPS-C2

tradiuz:
Adds the X440-G2-48p-10GE4, X440-G2-48t-10GE4, EPS-C2, and the Summit 750W PSU (used for the EPS-C2 and X460 Gen 1)


danner26:
Please resolve the test cases before marking this PR ready to review

PR #1862, opened 2024-01-17
Adds FS patch panel model 43552

hooligan-sa:
Adds 1U FS.com patch panel - https://www.fs.com/products/43552.html?attribute=29153&id=556469


device-types/FS/FHU-96FLC12MSMF.yaml
```diff
@@ -0,0 +1,35 @@
+---
+manufacturer: FS
+model: FHU-96FLC12MSMF
+slug: fs-fhu-96flc12msmf
+comments: '[FHU 1U Breakout Cassette, 96 Fibers OS2 Single Mode, Holds up to 12x MTP-8 to 24x LC Quad](https://www.fs.com/products/43552.html?attribute=29153&id=556469)'
+part_number: FHU-96FLC12MSMF (#43552)
```
danner26, line 6: The part number should just be `43552`

danner26 (changes requested):

PR #1863, opened 2024-01-17
Add DCS-7358 series

colinmacgiolla:
Small PR to add the DCS-7358 chassis, and a linecard


device-types/Arista/DCS-7358.yaml
```diff
@@ -0,0 +1,49 @@
...
+    maximum_draw: 2000
+  - name: PS2
+    type: iec-60320-c20
+    maximum_draw: 2000
+  - name: PS3
+    type: iec-60320-c20
+    maximum_draw: 2000
+  - name: PS4
+    type: iec-60320-c20
+    maximum_draw: 2000
```
danner26, line 21: Are these PSU's modular? If so, then they should be modelled here under `module-bays` and the associated PSU's should be created in the module-types folder if they do not already exist

device-types/Arista/DCS-7358.yaml
```diff
@@ -0,0 +1,49 @@
+---
+manufacturer: Arista
+model: DCS-7358
+slug: arista-dcs-7358
+part_number: DCS-7358
+u_height: 4
+is_full_depth: false
```
danner26, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

danner26 (changes requested):

device-types/Arista/DCS-7358.yaml
colinmacgiolla: Is it better to have the chassis only weight, or the fully loaded weight?

danner26: Correct, it should be the chassis only weight

colinmacgiolla:
Should be good now, but no idea how to mark the requested change as done 😉

PR #1865, opened 2024-01-21
Add ServerTech C1W08HC-0ABA2BAC

dorkmatt:


device-types/Server Technology/C1W08HC-0ABA2BAC.yaml
```diff
@@ -0,0 +1,67 @@
...
+  - name: th1
+    label: T/H1
+    type: rj-11
+  - name: th2
+    label: T/H2
+    type: rj-11
+console-server-ports:
+  - name: link
+    label: LINK
+    type: rj-12
```
danner26, line 25: Is this port actually used to provide remote access to the local console of a connected downstream device? If not then it should not be modelled under `console-server-ports`

device-types/Server Technology/C1W08HC-0ABA2BAC.yaml
```diff
@@ -0,0 +1,67 @@
...
+console-ports:
+  - name: ser
+    label: SER
+    type: rj-45
+  - name: th1
+    label: T/H1
+    type: rj-11
+  - name: th2
+    label: T/H2
+    type: rj-11
```
danner26, line 21: A `console-port` is defined as:

> A console port provides connectivity to the physical console of a device. These are typically used for temporary access by someone who is physically near the device, or for remote out-of-band access provided via a networked console server.

Do these actually fit that description? What are these ports actually used for?

dorkmatt: These are temperature and humidity sensor ports, which match a previously accepted commit - see `device-types/Server Technology/C2WG24SN-YCLN5D6.yaml`

device-types/Server Technology/C1W08HC-0ABA2BAC.yaml
dorkmatt: This is technically used to daisy chain PDU's together to share a common management  ethernet port. As mentioned with the other comment, this just follows a previously accepted merge - I think the intention was to cover all physical ports.

device-types/Server Technology/C1W08HC-0ABA2BAC.yaml
dorkmatt: If you'd like, I can remove these sensor ports from both my PR and the prior ServerTech units?

device-types/Server Technology/C1W08HC-0ABA2BAC.yaml
danner26: If that is the ports usage, then that is fine. Thanks for clarifying

PR #1876, opened 2024-01-23
Extreme x460 g2 24x 10ge4

hooligan-sa:
Adds model for the Extreme Networks X460-G2-24x-10GE4


danner26:
@DanSheps care to review the FS cassette "device"?

device-types/FS/FHU-96FLC12MSMF.yaml
DanSheps, line 1: Are the casettes already in the database as module types?

PR #1885, opened 2024-01-29
Cisco modules

francesco-noc:


module-types/Cisco/N7K-SUP2.yaml
```diff
@@ -0,0 +1,12 @@
...
+manufacturer: Cisco
+model: N7K-SUP2
+part_number: N7K-SUP2
+console-ports:
+  - name: console{module}
+    type: rj-45
+  - name: com1{module}
+    type: rj-45
+interfaces:
+  - name: mgmt0
```
DanSheps, line 11: This will result in a name collision, please change this to somehow utilize the {module} to prevent this.

DanSheps (changes requested):

DanSheps (approved):

DanSheps (changes requested):
Please remove the pickle files from your commit and fix the merge conflicts

danner26:
@francesco-noc one more just so you have it on your list

francesco-noc:
Why are there always pickle files committed? I removed them before commit and push.
Am I missing some configuration?

danner26:
> Why are there always pickle files committed? I removed them before commit and push. Am I missing some configuration?

You probably forked this too early and somehow got a different pickle file version. If you download the latest pickle files and add them in your commit, I think it'll fix the issue. That or start fresh from the latest master branch

module-types/Cisco/C9200-STACK.yaml
danner26, line 1: Why does this have nothing else on it? Is it the stacking kit that inserts into the switch?

DanSheps: it is a stack kit, I think it should have stack interfaces though

PR #1889, opened 2024-01-29
Add Datacom VS-1124-F

rtm516:


rtm516:
Not stale

device-types/Datacom/VS-1124-F.yaml
```diff
@@ -0,0 +1,70 @@
+---
+manufacturer: Datacom
+model: VS-1124-F
+slug: datacom-vs-1124-f
+u_height: 1
+is_full_depth: true
```
danner26, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

danner26 (changes requested):

danner26 (approved):

PR #1896, opened 2024-01-30
Add HPE StoreServ (3PAR) SANs

GreyChame1eon:


danner26:
@GreyChame1eon 
`subdevice_role: parent` is reserved for devices that have child devices which slot into them. Specifically a parent device should have `device-bays` defined. All you have is `module-bays`. Please convert these where appropriate, remove `subdevice_role` if the device has only modules, and add `device-bays` to the devices which have them

danner26:
Closing this since it has been incomplete for a long period of time. Feel free to work on it then reopen it again if you would like.

PR #1899, opened 2024-01-30
Added HPE SAN switches

GreyChame1eon:


device-types/HPE/1gb-ethernet-pass-thru-module-for-c-class-bladesystem.yml
danner26, line 1: Is this a module? Does this slot into another device? It has in the name module, this should probably be created as a module-type

danner26 (changes requested):

danner26:
Closing this since it has been incomplete for a long period of time. Feel free to work on it then reopen it again if you would like.

PR #1911, opened 2024-02-04
Added CCS-710P-16P device

siguroot:
Added Arista CCS-710P-16P, please let me know any amendments you'd like to see.  Thanks


siguroot:
Just snooping around on the other arista templates.  It appears arista themselves are updating these templates.  Can I assume this is why this is taking a while?

danner26 (approved):

PR #1916, opened 2024-02-08
Added Middle-Atlantic Storage Dawers and Tripp Lite Passthrough Patch Panel

kbrown900:
Added a small but desired feature of the Middle Atlantic storage drawers of the 2U and 4U variants. Also added the Tripp Lite pass through 24 port patch panel (pre Cat5e).


kbrown900:
Closed for further testing on why it is failing on yaml.

kbrown900:
I can't find where the issue is at, any help appreciated.

DanSheps (approved):

DanSheps:
Any chance these have weights you can assign?

kbrown900:
> Any chance these have weights you can assign?

Not that I have been able to find unfortunately. Best I could find was shipping weight for the patch panel which won't be entirely accurate.

PR #1921, opened 2024-02-10
Create SG-6100

ITS-Wied:


device-types/Netgate/SG-6100
danner26, line 1: Please remove all of the blank definitions

danner26 (changes requested):

PR #1922, opened 2024-02-10
Create SG-2100

ITS-Wied:


device-types/Netgate/SG-2100
```diff
@@ -0,0 +1,69 @@
+manufacturer: netgate
+model: sg-2100
+slug: sg-2100
```
danner26, line 3: The slug needs to be in the format of `{manufacturer}-{model}` all lower case

device-types/Netgate/SG-2100
```diff
@@ -0,0 +1,69 @@
...
+part_number: ''
```
danner26, line 4: Part number should replicate the slug unless it has a different part number

device-types/Netgate/SG-2100
```diff
@@ -0,0 +1,69 @@
...
+console-ports:
+- name: USB Console Port
+  type: usb-micro-b
+  label: USB Console Port
+  description: ''
+power-ports:
+- name: 12V DC in
+  type: dc-terminal
+  maximum_draw: null
+  allocated_draw: null
```
danner26, line 19: These should not be defined if you do not know them

danner26 (changes requested):

danner26:
@ITS-Wied 
It looks like this PR hasn't been worked on even though we reviewed it. We are going to close it for now, but feel free to re-open the PR if you get a chance to work on it. Thank you.

PR #1923, opened 2024-02-11
Added new MikroTik L009UiGS router model

daryll-swer:


device-types/MikroTik/L009UiGS-RM.yaml
```diff
@@ -0,0 +1,57 @@
+---
+manufacturer: MikroTik
+model: L009UiGS-RM
+slug: mikrotik-l009uigs-rm
+u_height: 0.0
+is_full_depth: false
+airflow: passive
```
danner26, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

danner26 (changes requested):

device-types/MikroTik/L009UiGS-RM.yaml
daryll-swer: Hello

As far as I know, MikroTik never publishes data on weight. I tried looking before and couldn't find the data.

Issue #1719, opened 2023-10-26
Add AVM Fritz!Box 7590 AX


tschmidtb51:
I would like to add AVM Fritz!Box 7590 AX to the device list. However, I came across the issue that there are (at least) 4 version of it (see different article numbers):
- 20002999 (international; https://en.avm.de/products/fritzbox/fritzbox-7590-ax/technical-specifications/)
- 20002998 (DE; without ISDN: https://avm.de/produkte/fritzbox/fritzbox-7590-ax/technische-daten/)
- 20002929 (DE; with ISDN: https://avm.de/produkte/fritzbox/fritzbox-7590-ax/technische-daten/)
- 20003022 (A/CH; https://at.avm.de/produkte/fritzbox/fritzbox-7590-ax/technische-daten/)

How should the `model` and `slug` be set?

danner26:
This is a community driven repo. If you would like to see a device added I suggest you take a look at existing files, and try to create your own for this device. After doing so feel free to submit a PR and we will review it/help you fix it up before merging!

PR #1927, opened 2024-02-15
Add Fritz!Box 7590 AX (#1719)

tschmidtb51:
as suggested in #1719

Still my question remains how to deal with the `model` name and the `slug` based on the fact that their are different hardware versions that have different article numbers (SKUs).


tschmidtb51:
The test fails as the `slug` does not end with the `model` name (which has parenthesis in it (as a suggestion). Is it intended to create slugs with parenthesis and semicolons given that they would result in not longer being a valid URL...

Please advice.

danner26:
This is in fact a bug in the tests, I can take a look at this soon hopefully. Sorry for the mess up

PR #1952, opened 2024-02-22
add NGXCA01Q && NGXPP01

antoine101009:


device-types/Nexconec/NGXPP01.yaml
danner26, line 1: This is missing a lot of information including:

- Part number
- airflow
- weight and weight_unit
- is_full_depth
- power definitions
- interface definitions if there are any management or non-modular ones

danner26 (changes requested):

PR #1962, opened 2024-02-26
Fortigate - Adjust weight, airflow and modular power supplies

lchabert:


danner26:
Hello, please fix the conflicts before we evaluate this further

PR #1980, opened 2024-03-05
Fix OpenGear OM2224-24E-L interface names and add OM1208-8E-L

pv2b:
Hello!

Here are some contributions for some OpenGear device.

As for OM2224-24E-L:

The interface names for the OpenGear devices were completely wrong. There is no "eth0" and "eth1", they're actually labeled "net1" and "net2" inside of the Linux OS also. (As seen in `ip addr`). This also matches the corresponding outside labelling.

Also, the switchports are not actually 25-48, they're 1-24, so I renamed them according to the correct interface name according to `ip addr` for the sake of consistency. I also renamed the cellular interface to `wwan0` to keep it consistent.

I removed the `management_only` field since that makes no sense for this device. This device has no "management only" ports, any port can be a routing port if so configured.

Finally, I added a `sw0` bridge interface. I would like to define in the YAML that all the `sw0p1` etc ports are part of this bridge, but I can't figure out how to define this in YAML? I've just solved it on my end by just changing the device templates all point-and-clicky on my end, but it would of course be better if there was a real way of doing this in the YAML for the community benefit.


DanSheps (approved):
Review the model, you are likely correct.

PR #1985, opened 2024-03-06
Nokia add 7750 1se 2se devices

jbemmel:


DanSheps:
@jbemmel I think you forgot to `git add` the renamed files.  It only shows a delete here.

DanSheps:
FYI, still not seeing any files in the commits

PR #1991, opened 2024-03-12
Added Fortigate FG-120G, and created front/rear images

Kullsyrevann:
Added device type and images for the new Fortinet Fortigate FG120G.


device-types/Fortinet/FG-120G.yaml
```diff
@@ -0,0 +1,82 @@
...
+    allocated_draw: 38
+  - name: PS2
+    type: iec-60320-c14
+    maximum_draw: 40
+    allocated_draw: 38
+interfaces:
+  - name: ha
+    type: 1000base-t
+  - name: mgmt
+    type: 1000base-t
```
danner26, line 26: Are either of these out of band management interfaces? If so, you should add `mgmt_only: true`

danner26 (commented):
Please mark OOB management interfaces with `mgmt_only: true`

danner26 (approved):

PR #2004, opened 2024-03-18
Eaton 5sx1750rau device type added

jscooksey:
Added Eaton 5SX1750RAU device type


jscooksey:
Passing ore-commit.  I'll have to find out what this is referring to
```
Run mshick/add-pr-comment@v2
Error: Resource not accessible by integration
```

PR #2005, opened 2024-03-18
Eaton 5sx1750rau (2nd)

jscooksey:
Added Eaton UPS 5SX1750RAU device type
pre-commit has run and Passed all locally


jscooksey:
IM not modifying that file......hmmm

PR #1955, opened 2024-02-22
Adding template for Juniper ACX7024 router and ZPE NSR-BASE-DAC with preloaded expansion modules

modonovan:
The purpose of this pull request is to add a new template, not yet created, for the [Juniper ACX7024 router](https://www.juniper.net/us/en/products/routers/acx-series/acx7000-family-cloud-metro-routers/acx7000-line/acx7024-acx7024x-cloud-metro-routers-datasheet.html) as well as the [ZPE Network Services Router with preloaded expansion modules](https://zpesystems.com/products/nodegrid-services-router/).

In the case of the Juniper ACX7024, I have modeled this to the best of my capability based on the Juniper product datasheet for this router and in conjunction with the ACX7100 templates that preexist. 

In the case of the ZPE NSR device, the template is modeled on the assume that;

* There is a [16 port serial console module](https://zpesystems.com/products/nodegrid-services-router/serial-expansion-card/) in Slot 1 of the device (hence the numbering ttyS1-1 etc)
* There is a [16 port ethernet switch module](https://zpesystems.com/products/nodegrid-services-router/ethernet-expansion-card/) in Slot 2 of the device (hence the numbering starting netS2-1 etc)

PR #2011, opened 2024-03-19
MOD - adding ACX7024 Juniper device type and ZPE NSR

modonovan:
The purpose of this PR is to replace legacy PR1955 which had binary files that were not required and should not have been committed.

### DIFF
--- a/device-types/Juniper/ACX7024-AC.yaml
+++ b/device-types/Juniper/ACX7024-AC.yaml
@@ -82,9 +82,3 @@ module-bays:
     maximum_draw: 150
     position: '1'
   - name: Fan Module 0
-  - name: Fan Module 1
-  - name: Fan Module 2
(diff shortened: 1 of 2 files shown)


netravnen:
#1955

netravnen:
@modonovan [Disclaimer: Dumb question] Would it not have been easier to `push --force-with-lease` to the old branch PR1055 was based on? Instead of a new PR? _(not that it really matters, just a throught)_

modonovan:
Yes, alas, it almost certainly would have been but I was having too much trouble trying to remove the binary files that shouldn't have been there and it felt like this was the cleanest way. This is my lack of GitHub PR knowledge that held me back this time. Next PR I'll be forewarned and a bit more clued in.

DanSheps (approved):

PR #2025, opened 2024-03-24
Master

darkcloud784:
Change ASR99k's to use module bays instead of device bays to bring all the 9k's into the same standard.


device-types/Cisco/ASR-9906.yaml
```diff
@@ -39,3 +40,4 @@ device-bays:
     label: Line Card 3
```
danner26, line 40: You are missing `position` for these bays (slot 1-4)

danner26 (changes requested):

darkcloud784 (commented):
Updated positions

danner26 (approved):

PR #2038, opened 2024-03-27
JSON Schema Migration

danner26:
JSON Schema has deprecated the RefResolver that we have been relying on for validation. This is becoming even more prevalent as we update other test suites and python packages. This PR resolved the deprecation and also does the following:

- Updates jsonschema to 4.21.1
- Adds $id and #schema to each of the applicable schema files
- Updates schema references from file paths to URN's
- Adds new `SCHEMAS_BASEPATH` variable which defined the local path to the schema files

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-jsonschema==4.19.0
+jsonschema==4.21.1
 jsondiff==2.0.0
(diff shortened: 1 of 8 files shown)


danner26:
Fixed #1908

PR #2050, opened 2024-03-29
Device Type: Update Raspberry Pi

netravnen:
~~Add~~
- ~~Pi 1 (256M, 512M)~~
- ~~Pi 2 (1G)~~
- ~~Pi 4 400 (4G)~~
- ~~Pi 5 (4G, 8G)~~

Collapse all PI models into generation and remove memory sizes from the files at @danner26 request.

### DIFF
--- a/device-types/Raspberry Pi/RPI-MODB.yaml
+++ b/device-types/Raspberry Pi/RPI-MODB.yaml
@@ -1,15 +1,13 @@
 ---
 manufacturer: Raspberry Pi
-model: Raspberry Pi 3 Model B 1GB
-slug: raspberry-pi-rpi3-modb-1gb
-part_number: RPI3-MODB-1GB
+model: Raspberry Pi Model B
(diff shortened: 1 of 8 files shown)


danner26:
Right now storage size is not a differentiating factor. They are otherwise the same hardware from a networking perspective. For the time being (until we introduce storage management), please condense them down into 1 model for each series, unless a specific series has multiple models with different interfaces.

netravnen:
> Right now storage size is not a differentiating factor. They are otherwise the same hardware from a networking perspective. For the time being (until we introduce storage management), please condense them down into 1 model for each series, unless a specific series has multiple models with different interfaces.

@danner26 Noted!

Should I condense the already existing RPI models in the library, too?

https://github.com/netbox-community/devicetype-library/tree/master/device-types/Raspberry%20Pi

danner26:
Yes that would be appreciated!

danner26 (approved):

PR #2053, opened 2024-04-01
RUT955 spec

TaLoN1x:
added Teltonika RUT955 spec


TaLoN1x:
will sumbit with another device

PR #2065, opened 2024-04-04
Test Cases: Added VGA denial test case

danner26:
VGA interfaces are not supported types as of now. Thus a test case was created to ensure there are none listed.


NeverBehave:
Why VGA is not a supported type? What's the context here?

NeverBehave:
Ah I think it is related to this one: https://github.com/netbox-community/netbox/issues/5298

Media related port are seems to be rejected and have to develop as plugin.

PR #2070, opened 2024-04-04
Add HPE Aruba CX 6400 Switches & Modules

robinreinhardt:


robinreinhardt:
Hey @danner26,

I'm quite new to the devicetype library so hopefully I didn't get it wrong.

The PoE Type `type6-ieee802.3bt` is not known in the `generated_schema.json` file.
As far as I understand, this file is generated out of the Netbox Repo (contrib/generated_schema.json).
How does the process work here?
I would suggest that we add the two values `type6-ieee802.3bt` and `type8-ieee802.3bt`

Best regards
Robin

robinreinhardt:
nvm, realized I was messing up PoE classes and types. The PoE types are now fixed.

danner26 (approved):

PR #2075, opened 2024-04-09
Cisco Devices: Replace 'C' with 'Catalyst' in model

netravnen:
Counting models using Catalyst (215'ish) vs C (55'ish) this change proposes changing 'C' with 'Catalyst' in the model name field for the Cisco devices.

Resolve a validation error in C7201.yaml

### DIFF
--- a/device-types/Cisco/C1000-16FP-2G-L.yaml
+++ b/device-types/Cisco/C1000-16FP-2G-L.yaml
@@ -1,6 +1,6 @@
 ---
 manufacturer: Cisco
-model: C1000-16FP-2G-L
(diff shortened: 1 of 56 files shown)


netravnen:
**Before**

```
$ grep -r --perl-regexp 'model: Catalyst \d+' --files-with-matches | wc -l
215
```

```
$ grep -r --perl-regexp 'model: C\d+' --files-with-matches | wc -l 
56

```
**After**

```
$ grep -r --perl-regexp 'model: Catalyst \d+' --files-with-matches | wc -l
271
```

danner26 (approved):

PR #2076, opened 2024-04-10
Cisco: Add front image cisco-n5k-c5548up

netravnen:


danner26:
Hey, can you try to get a more straight on image (or better yet one from the manufacturer)? You will also need to update the devices definition file to include `front_image: true`

harryajc:
PR has not had a response in over 6 months and image has been merged from a different PR

PR #2080, opened 2024-04-11
Add HPE Aruba AP 600 Series

robinreinhardt:


danner26 (approved):

moetezz:
I think this should be reviewed, As there is only one wifi interface, however, APs have a dedicated interface for each Radio Frequency so 5Ghz should have it is own as for 6Ghz and 2.4Ghz, Not all merged together as one interface, since you might wanna be granular on the way you connect your devices in NetBox, some use 2.4Ghz some uses 5Ghz, etc....

danner26:
@moetezz you are more than welcome to correct anything you find in this repo. This is largely a community driven project, so we rely on people who have these devices/have technical expertise with them to make and correct the definition files. Please feel free to submit a PR and mark it ready for review once the corrections have been made. Thank you!

PR #2085, opened 2024-04-12
Adding CyberPower CPS1215RM

amyers1616:


device-types/CyberPower/CPS1215RM.yaml
```diff
@@ -0,0 +1,43 @@
+---
+manufacturer: CyberPower
+model: CPS1215RM
+slug: cyberpower-cps1215rm
+part_number: cps1215rm
+u_height: 1
```
danner26, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

amyers1616 (commented):
Sure thing. Will get those in here.

danner26 (approved):

PR #2093, opened 2024-04-17
Cisco Device: Add C9500X-28C8D

netravnen:
Cisco Catalyst 9500X switch with 28x 40/100G QSFP28 ports
+ 8x 40/100/200/400G Gigabit Ethernet


device-types/Cisco/C9500X-28C8D.yaml
```diff
@@ -0,0 +1,247 @@
+---
+manufacturer: Cisco
+model: Catalyst 9500X-28C8D
+slug: cisco-c9500x-28c8d
+part_number: C9500X-28C8D
+u_height: 1
+is_full_depth: true
```
danner26, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

device-types/Cisco/C9500X-28C8D.yaml
```diff
@@ -0,0 +1,247 @@
...
+    type: rj-45
+  - name: usb
+    type: usb-c
+power-ports:
+  - name: PS-1
+    type: iec-60320-c14
+    maximum_draw: 1500
+  - name: PS-2
+    type: iec-60320-c14
+    maximum_draw: 1500
```
danner26, line 22: These appear to be modular PSU's, is this the case? If so, these should be modelled under `module-bays` and the associated PSU's should be added to the module-types folder if they do not already exist

device-types/Cisco/C9500X-28C8D.yaml
```diff
@@ -0,0 +1,247 @@
...
+    position: '6'
+  - name: ps1
+    label: PS-1
+    position: '1'
+  - name: ps2
+    label: PS-2
+    position: '2'
+  - name: disk0
+    label: DISK-0
+    position: '1'
```
danner26, line 247: disks should not be here, please remove this

danner26 (changes requested):

harryajc (approved):

PR #2094, opened 2024-04-18
Add Synology devices

rose-a:


device-types/Synology/RS1221RP-plus.yaml
```diff
@@ -0,0 +1,31 @@
...
+weight: 8.4
+weight_unit: kg
+console-ports:
+  - name: Serial
+    type: de-9
+power-ports:
+  - name: PSU 1
+    type: iec-60320-c14
+  - name: PSU 2
+    type: iec-60320-c14
```
danner26, line 18: Are these PSU's modular? If so, these should be defined under `module-bays`, and then the associated PSUs should be in module-types if they do not already exist

module-types/Synology/E10G18-T2.yaml
```diff
@@ -0,0 +1,10 @@
+---
+manufacturer: Synology
+model: E10G18-T2
+part_number: E10G18-T2
+comments: '[Synology E10G18-T2 - Network adapter - PCIe 3.0 x8 - 10 Gigabit 2x RJ45](https://www.synology.com/en-global/products/E10G18-T2#specs)'
+interfaces:
+  - name: Port 1
```
danner26, line 7: In these names, you should replace the numbers with `{module}`

module-types/Synology/E10G18-T2.yaml
```diff
@@ -0,0 +1,10 @@
...
+    type: 10gbase-t
+  - name: Port 2
```
danner26, line 9: In these names, you should replace the numbers with `{module}`

danner26 (changes requested):

danner26 (approved):

PR #2101, opened 2024-04-22
PowerEdge R760 added

TaLoN1x:
PowerEdge R760 added


TaLoN1x:
@danner26, do you have any specific issues? I went through the commits again and dint't detect an issue

harryajc:
the dell R760 has been added from a different pull request

PR #2112, opened 2024-04-23
Add Several FS LIUs

Nealtron:


Nealtron:
> I am not sure if `module-bays` are best or not here. @DanSheps what do you think?

I cloned the the existing FS LIUs and modified the configuration to account for the differences. If module-bays are incorrect, then the entire series of FS LIUs should be updated to meet the standard.

DanSheps:
Module bays would be good here.  They are enclosures for fiber cassettes.

Nealtron:
Is there anything I need to do in order to merge this commit?

harryajc (approved):

PR #2125, opened 2024-04-30
yamllint: Add empty-values

netravnen:
Enable
- forbid-in-block-mappings
- forbid-in-flow-mappings
- forbid-in-block-sequences

https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.empty_values

### DIFF
--- a/.yamllint.yaml
+++ b/.yamllint.yaml
@@ -27,3 +27,7 @@ rules:
   octal-values: disable
   quoted-strings: disable
   trailing-spaces: enable
+  empty-values:
+    forbid-in-block-mappings: true
+    forbid-in-flow-mappings: true
+    forbid-in-block-sequences: true


danner26:
Hello, thank you for your PR.

Per the [documentation](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.empty_values) you linked, we defined `empty-values` as enabled ([see line 17 in .yamllint.yaml](https://github.com/netbox-community/devicetype-library/blob/2e70c7c7728c0d4448b7e508f5402c047984e21b/.yamllint.yaml#L17)). Thus, according to that documentation the default dictations for `empty-values` is the following:
```
rules:
  empty-values:
    forbid-in-block-mappings: true
    forbid-in-flow-mappings: true
    forbid-in-block-sequences: true
```

Since `empty-values` is in fact enabled, the 3 rules you are looking to add are already implemented per the yamllint documentation. I went a step further and verified this by running `yamllint -c .yamllint.yaml --format parsable --strict device-types/` and `yamllint -c .yamllint.yaml --format parsable --strict --strict module-types/`. Both of these commands confirmed that there are no changes, and the current implementation is the same as your proposed changes.

With all of this in mind, we will not be merging this PR. Thank you for your time and contribution.

PR #2127, opened 2024-04-30
Gitignore: Ignore .pdn (Paint.Net) files

netravnen:


danner26:
This is not a required gitignore

PR #2131, opened 2024-05-01
added NIM-24A.yaml

Ido-Don:
Hello there :wave:
I noticed the NIM-24A module type was missing from the modules folder and i needed it for some experimenting in Netbox. 
I added it in my local Netbox instance and I believe it would be beneficial to have it in the main repo.

If there is anything i missed or that need changing let me know and i will address them.

Looking forward to your feedback.


module-types/Cisco/NIM-24A.yaml
```diff
@@ -0,0 +1,54 @@
+---
+manufacturer: Cisco
+model: NIM-24A
+part_number: NIM-24A
```
danner26, line 4: You are missing a lot of definitions. Please review other files and the main readme to see what should be defined

danner26 (changes requested):

module-types/Cisco/NIM-24A.yaml
Ido-Don: Thank you for reviewing my pull request! I've added the module's weight and weight_unit fields, but I want to ensure we're covering all relevant information. Since it's a module-type, I wasn't able to find any more definitions, but I might be missing something. Do you have any suggestions for additional details or improvements we could include? I'm open to any feedback you have. Thanks again for your help!

harryajc:
Hi Thank you for your contribution.

NIM-24A has already been added through a different PR

PR #2135, opened 2024-05-02
Dell C6400 Series updates

adamboutcher:
Four C6420's sit inside a C6400 Chassis, these changes reflect that. I've also added estimate weight of a fully laden system.
Updated the 6420 PDF link which was incorrectly linking to a 6520.


danner26:
Please resolve all test cases before marking this PR ready

harryajc (approved):

PR #2156, opened 2024-05-21
add dell powerscale H700

sjurtf:


harryajc (approved):
Adjusted files to update already existing H700 device

PR #2163, opened 2024-05-21
Add Cisco WS-C2960C-8TC-S

sebastianreifetshammer:
Add device type Cisco WS-C2960C-8TC-S


device-types/Cisco/WS-C2960C-8TC-S.yaml
```diff
@@ -0,0 +1,31 @@
+---
+manufacturer: Cisco
+model: Catalyst 2960-8TC-S
+slug: cisco-ws-c2960c-8tc-s
+part_number: WS-C2960C-8TC-S
+is_full_depth: false
+u_height: 1
```
danner26, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

danner26 (changes requested):

device-types/Cisco/WS-C2960C-8TC-S.yaml
sebastianreifetshammer: Hello,

I added weight & unit, switch has no airflow - fanless switch

Thanks!

harryajc (approved):

PR #2167, opened 2024-05-24
Add Juniper mx304

OliElli:
Add Juniper MX304 devicetype, along with RE and LMIC miodule types to match


Olen:
Sorry, I did not see this when I created #2298 or else I would have suggested modifications to this one.
However, I believe the port mapping and a few other details are more correct in my PR.

OliElli:
@Olen Agreed, closing mine!

PR #2185, opened 2024-06-06
Create one Raritan PDU PX3-5516.yml

FlorianHeigl:
Also added the USB Port, it likely should be on all PX3. It also has a Wireless connection option but I don't know how that would need to be modelled.


FlorianHeigl (commented):
fix a typo and write a comment because this is 2024.

PR #2207, opened 2024-06-23
Add Huawei CE6856-48T6Q-HI

AskskwBv8T2nrm4Qnj:
Add Huawei CE6856-48T6Q-HI (https://support.huawei.com/enterprise/en/doc/EDOC1000019246/5e0ec9b2/ce6856-48t6q-hi)


device-types/Huawei/CE6856-48T6Q-HI.yaml
```diff
@@ -0,0 +1,135 @@
...
+module-bays:
+  - name: FAN1
+  - name: FAN2
+power-ports:
+  - name: PWR1
+    type: iec-60320-c14
+    maximum_draw: 600
+  - name: PWR2
+    type: iec-60320-c14
+    maximum_draw: 600
```
danner26, line 23: Hello, thanks for the contribution. It appears that these PSU's are modular. These should be defined under `module-bays` rather than `power-ports`, and then the associated PSU modules should be created within the module-types folder if they do not already exist. Thank you.

danner26 (changes requested):

harryajc (approved):

PR #2245, opened 2024-07-08
Added Images for various Juniper EX platform

harryajc:
Added Images for various Juniper EX series products, including updating of the yaml files.


harryajc:
> Hey @harryajc could you please crop these images better so that it is only the devices rather than extra whitespace/transparent background. It will show better if there is no extra background and just the device face/rear. Thanks!

All done!

danner26 (approved):

PR #2269, opened 2024-07-25
Adicionado imagens do Aruba 1930-48P

renatobeletti:
Adicionado imagens do Aruba 1930-48P


ravench:
Did you set `front_image: true` and `back_image: true`?

harryajc (approved):

PR #2286, opened 2024-08-04
Add Cisco C2960G-8TC-L

philippderdiedas:
Add the model above

also GigabitEthernet0/8 is physically a shared port with an 1000base-t rj45 and and sfp (1G). how would one represent that?
in cli its just `GigabitEthernet8`


device-types/Cisco/WS-C2960G-8TC-L.yaml
```diff
@@ -0,0 +1,30 @@
+---
+manufacturer: Cisco
+model: Catalyst 2960G-8TC-L
+part_number: WS-C2960G-8TC-L
+slug: cisco-ws-c2960g-8tc-l
+u_height: 1
```
danner26, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

danner26 (changes requested):

device-types/Cisco/WS-C2960G-8TC-L.yaml
philippderdiedas: Sorry for late response, back from the holidays. Added missing datapoints as requested

harryajc (approved):

PR #2291, opened 2024-08-09
Add C9300X-24Y.yaml

rhasti:
Add C9300X-24Y device type


harryajc:
the C9300X-24Y has been already added from a different PR

PR #2292, opened 2024-08-09
Add U7-Pro.yaml

rhasti:
Requires wlan type ieee802.11be


harryajc:
Hi @rhasti
The Unifi U7-PRO has already been added by a previous PR

PR #2296, opened 2024-08-12
Add weight for some NK9 devices and fix device height.

ravench:
- Fix U height for `N9K-C93240YC-FX2`
- Add weight for `N9K-C93216TC-FX2` and `N9K-C93240YC-FX2`


device-types/Cisco/N9K-C93240YC-FX2.yaml
```diff
@@ -5,7 +5,9 @@ part_number: N9K-C93240YC-FX2
 slug: cisco-n9k-c93240yc-fx2
 front_image: true
 rear_image: true
-u_height: 2
+u_height: 1.5
```
DanSheps, line 8: This should be labelled as 2RU as that is Cisco's official stance on the height according to the datasheet

DanSheps (changes requested):

device-types/Cisco/N9K-C93240YC-FX2.yaml
Nefti-sama: Not correct, the 93240 has a height of 1.2 units (see datasheet https://www.cisco.com/c/en/us/products/collateral/switches/nexus-9000-series-switches/datasheet-c78-742282.html#Switchmodels or comparision https://www.cisco.com/c/en/us/products/switches/nexus-9000-series-switches/nexus-9300-10ge-fiber-switches-comparison.html)
The question is, how netbox will react to a 1.2RU device (drawing it into the rack). In practice it is quite unlikeley that you will use the 0.8RU for anything as it completely screws the positioning of all other devices, so 2RU will be "used up"

harryajc (approved):

device-types/Cisco/N9K-C93240YC-FX2.yaml
harryajc: As the spec sheet says 1.2 and netbox allows .5 U increments i think 1.5U is the best option

DanSheps: > Cisco Nexus 93240YC-FX2 Switch
>
> The Cisco Nexus 93360YC-FX2 Leaf Switch is an 2-Rack-Unit (2RU) Leaf switch that supports 7.2 Tbps of bandwidth and 2.4 bpps across 96 fixed 10/25G SFP+ ports and 12 fixed 40/100G QSFP28 ports (Figure 3). The 96 ports of downlinks support 1/10/25-Gbps. The 12 uplinks ports can be configured as 40- and 100-Gbps ports, offering flexible migration options. The switch has FC-FEC and RS-FEC enabled for 25Gbps support over longer distances. Please see feature table below for more information.

DanSheps (approved):

PR #2300, opened 2024-08-14
FS FHD Rackmount + DWDM Modules

suom1:


module-types/FS/FHD-D82936EM.yaml
grbeneke, line 1: The "1310" front-port appears to be missing from this model. Please update to match the data sheet for this part number.

module-types/FS/FHD-D85360EM.yaml
grbeneke, line 1: The "1310" front-port appears to be missing from this model. Please update to match the data sheet for this part number.

grbeneke (changes requested):
Please update the two modules with missing front-ports.

module-types/FS/FHD-D85360EM.yaml
suom1: This version does not have 1310 front port.

module-types/FS/FHD-D82936EM.yaml
suom1: This does not have 1310 frontport, but updated the url and partnumber

PR #2310, opened 2024-08-21
added device type Dell PowerEdge R350 with elevation images

ionoci:
I added device type Dell PowerEdge R350 with elevation images dell-poweredge-r350.front.jpg and dell-poweredge-r350.rear.jpg


device-types/Dell/PowerEdge-R350.yaml
```diff
@@ -0,0 +1,25 @@
...
+front_image: true
+rear_image: true
+console-ports:
+  - name: Serial
+    type: de-9
+power-ports:
+  - name: Power 1
+    type: iec-60320-c14
+  - name: Power 2
+    type: iec-60320-c14
```
danner26, line 17: Hello, thanks for the contribution. It appears that these PSU's are modular. These should be defined under `module-bays` rather than `power-ports`, and then the associated PSU modules should be created within the module-types folder if they do not already exist. Thank you.

danner26 (changes requested):

device-types/Dell/PowerEdge-R350.yaml
ionoci: Hi, your are right. PSUs are modular. I changed it in the template and added the corresponding PSU we got in use.
I'm not familiar with git(hub) workflows, so do not hesitate telling me any improvment I may do.

DanSheps (approved):

DanSheps (approved):

PR #2332, opened 2024-09-09
Fix arubaAPs

moetezz:
Fixed the WiFi interfaces for the Aruba APs since technically each WiFi radio, for example, 2.4GHz, 5GHz, and 6GHz, is an independent interface rather than one WiFi interface( in a use case a device might be connected using 2.4GHz signal to the AP thus it doesn't make it connected to the 5GHz), and this granularity makes it better.
Also fixed the console ports and power ports name to better fit the specs from the DataSheet.
Removed USB-A from console ports since they are not console ports.
FIxed Some Ethernet interfaces definition


robinreinhardt:
Hey moetezz,

I'd suggest that we should leave the product numbers (SKUs) as they were because they depend on the different areas in the world.
For example, the AP-515 has the following SKUs:

- US: Q9H63A
- RW: Q9H62A
- JP: Q9H61A
- IL: Q9H60A
- EG: Q9H59A

This is how it is done with most of the Aruba AP models.

moetezz:
Hey @robinreinhardt, 

Thank you for your feedback.
You are right about the different SKUs for different regions. In this case, I would suggest keeping them empty, it is not a required field anyway and I don't think having inaccurate information will be helpful.

Let me know what you think.

iamk3 (commented):
These updates are the correct way to go for Aruba APs. Hopefully this will be merged soon! Thank you for this work!

robinreinhardt:
@moetezz 
In my opinion I would not leave the SKU field empty, as I think it is often used for imports and automations.
If it is empty you would have to parse the name from the model name

moetezz:
Under your request, I reverted the part number to reflect the model name.

iamk3:
In my opinion, the SKU was the correct way to go. This would bring consistency across the portfolio. Then APs for other Regulatory Domains could be added.

harryajc (approved):

PR #2333, opened 2024-09-10
Add Palo Alto PA-1410

mathieumd:


mathieumd:
Not sure why `format` is failing. Its output is not helpful. Any idea?

harryajc:
Hi @mathieumd 

I would try removing all of the values that just have ' ' as these are not required.

harryajc (approved):

mathieumd:
Thanks @harryajc.

PR #2349, opened 2024-09-19
Synology rs3621xs+

iamk3:
add Synology RS3621xs+


iamk3:
I'm looking at the test and it is failing due to not having "usb-a" as a definition in the rear ports section. I can't find which code to update to add "usb-a" to the schema.

iamk3:
I went ahead and removed the rear usb ports definition. Where is the best place to define those for future reference?

jonkerj:
The validation is done according to the schema (`./schema/*`) and `rear-ports` have a couple of valid options (see `generated-schema.json`), but none of which is USB. That schema is automated from upstream NetBox, so it will be hard to add USB to it.

Most networking devices that include USB (for instance FortiGate firewalls) have these ports modeled as `console-ports`, which have USB options. Rear ports [appear to be something different](https://demo.netbox.dev/static/docs/core-functionality/devices/#rear-ports) than you seem to expect, I'm afraid

harryajc (approved):

PR #2350, opened 2024-09-19
Hetzner

arosenhagen:
add hetzner dedicated, cloud, managed and storage device-types

### DIFF
--- a/device-types/Hetzner/ax102.yaml
+++ b/device-types/Hetzner/ax102.yaml
@@ -1,0 +1,6 @@
+---
+manufacturer: Hetzner
+model: AX102
+slug: htz_ax102
+part_number: ax102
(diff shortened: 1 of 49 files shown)


DanSheps:
I am sorry, but Hetzner is not a manufacturer with their own device types.  They are a hosting provider and these are the "plans" you can buy from them.

The only thing we might be able to accept for addition is the actual Physical hosts and even then I am not 100% swayed.  The Cloud, Storage and Managed virtual is 100% something we won't accept.

Please review the "[Device Type](https://netboxlabs.com/docs/netbox/en/stable/models/dcim/devicetype/)" definition in the NetBox docs for more information.

arosenhagen:
Thanks for the clarification. As we want to use netbox in a broader scope which might be of interest to others, I'll maintain it in a branch of my fork.

PR #2351, opened 2024-09-19
Apple

arosenhagen:
add/optimize apple device-types
add iphone and ipad device-types


danner26:
Hello @arosenhagen iPads, iPhones, etc are not really network devices and moreso end user devices. These are not the intent of this repo. You are welcome to keep them in a forked copy but I do not think they belong here.

If you want to remove them from this PR and only have the updates to the network devices you are welcome to reopen this at that time. Thanks in advance for your understanding.

PR #2368, opened 2024-10-04
Add some devices

jonkerj:
I came across a couple of devices in our environment(s) that were not (yet) in the DTL, so I wrote this PR to address that. It is based upon #2364, assuming that will get merged.


jonkerj:
I don't know why the validator bails, there is no clear error message, and it runs fine locally:

```
devicetype-library:$ pre-commit run --config .pre-commit-yamlfmt-config.yaml --all-files
Format YAML files........................................................Passed
```

I'm just guessing here, but could it be that my additions make the yamlfmt process trip over some hidden limit (memory, cpu, time)? 

I could try and see if I can modify the validator to only validate changed YAML files, but the amount of open PRs without response worry me: I wouldn't like to put time into something that gets ignored.

harryajc:
Hi @jonkerj 
Just had a look over this, not quite sure why the test is failing.

However a few changes are required as there are a few devices that have been added as having "power ports" however if the device has removable PSU these should be modelled as module bays.

Harry

jonkerj:
Sure, I'll have a look today and modularize these removable PSUs. As I have based many of these switches on existing models (ie, copy/paste the fibre-model into a copper-model), I'll try and fix the orignals as well.

jonkerj:
I rebased my branch on current master, modularized PSUs of my added devices (adding some module-types on the way), and modularized the sources of devices I copy/pasted (eg ICX6610). Should be fine now, I think

danner26:
Hey all, ill start working on this now. I will update you once I have more info

danner26:
The issue is the following file: `device-types/HPE/ProCurve-2900-48G.yaml`
The `A1`, `A2, etc.. names have `'` around them, but since they start with a letter they do not need that. The linting task corrects this, but since there is a change it fails the validation steps. Running `pre-commit run -a` on your local machine would have corrected this which would have required a commit and push afterwards.

danner26 (dismissed):

harryajc (dismissed):

harryajc (approved):

PR #2372, opened 2024-10-07
Add deciso dec2700. Pre commit checks passed.

bk-lg:
Added  the dec2700 firewall appliances from deciso.

### DIFF
--- a/device-types/Deciso/DEC2700.yaml
+++ b/device-types/Deciso/DEC2700.yaml
@@ -2,7 +2,7 @@
 manufacturer: Deciso
 model: DEC2700
 part_number: DEC2700
-slug: deciso-dec3840
+slug: deciso-dec2700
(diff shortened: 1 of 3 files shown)


harryajc:
Hi @bk-lg 
Looking at the spec sheet is this PR for the 2750? as there doesn't appear to be a 2700 model.

bk-lg:
@harryajc you are completly right. I was confused by the 2700 series label on the hardware itself. I fixed the naming with my latest commit.

harryajc (approved):

bk-lg:
Fyi: I documented an older revision of this hardware with 1GbE Ports. But I'm fine with the change to the newer revisions specs. 

@harryajc thank you vermy much for looking into it and removing the accidently added second file. This should have been only for our internal device types and not in this MR.

PR #2378, opened 2024-10-10
New Device type: Cudy WR3000 v1

Floppe:
Adds device type for Cudy WR3000 v1 Wifi access point.

### DIFF
--- a/device-types/Cuda/WR3000-v1.yaml
+++ b/device-types/Cuda/WR3000-v1.yaml
@@ -1,0 +1,27 @@
+---
+manufacturer: Cudy
+model: WR3000 v1
+slug: cudy-wr3000-v1
+u_height: 0
+is_full_depth: false
+airflow: passive
+weight: 295
+weight_unit: g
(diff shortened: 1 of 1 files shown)


device-types/Cudy/WR3000-v1.yaml
```diff
@@ -0,0 +1,24 @@
+---
+manufacturer: Cudy
+model: WR3000 v1
+slug: cudy-wr3000-v1
+u_height: 0
+is_full_depth: false
```
danner26, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

danner26 (dismissed):

device-types/Cudy/WR3000-v1.yaml
Floppe: Hello,

Weight and airflow now added.

harryajc (approved):

PR #2398, opened 2024-10-23
add extreme-networks-X250e-Series

Kaioguilherme1:
add extreme-networks-X250e-Series

*  X250e-24p
*  X250e-24t
*  X250e-24tDC
*  X250e-24x
*  X250e-24xDC
*  X250e-48p
*  X250e-48t
*  X250e-48tDC

Reference: https://documentation.extremenetworks.com/summit_16/GUID-CDAF102B-B591-4011-A892-2AA495D4E1DB.shtml


harryajc:
HI @Kaioguilherme1 Thank you for your contribution.
In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

3. You seem to have a duplicate file for each model with "-switch" on the end these appear to be the same as the regular and the datasheets are the same so these are not required.
4. For POE interfaces please add the following to each interface:
`poe_mode: pse`
`poe_type: type1-ieee802.3af`

PR #2400, opened 2024-10-23
Add Meinberg Lantime M1000

pcharleston1:


device-types/Meinberg/LANTIME M1000.yml
```diff
@@ -0,0 +1,28 @@
+---
+manufacturer: Meinberg
+model: LANTIME M1000
+slug: meinberg-lantime-M1000
+u_height: 1
+is_full_depth: true
+console-ports:
```
harryajc, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

robinreinhardt:
duplicate of #2240
but this PR seems more comprehensive, so I would close mine.
The airflow is passive. Nothing found about the weight. But it depends on the used modules

harryajc (approved):

PR #2405, opened 2024-10-24
Moved the AP yaml files to the correct directory. They should have be…

imaginejosephishak:
…en in device-types/APC and not the root directory


harryajc (approved):
Thanks for catching that

PR #2421, opened 2024-10-26
add extreme-networks-X460-Series

Kaioguilherme1:
dd extreme-networks-X460-Series

* X460-24t
* X460-48x
* X460-24p
* X460-24x

Reference: https://documentation.extremenetworks.com/summit_16/GUID-B1CEFBEF-7E0B-45D1-A376-59E1F07453FC.shtml


harryajc:
Hi @Kaioguilherme1 Thank you for your contribution.

Looking at the spec sheet the x460 series is POE+ not POE therefore rather than `poe_type: type1-ieee802.3af` you need to use `poe_type: type2-ieee802.3at` for POE+ interfaces.

Also looking at the spec sheet the x460 switches have modular power supplies, therefore these need to be set under `module-bays` and not `power-ports` then the required PSU models need to be created under the module-types/extreme networks/ folder

Kaioguilherme1:
ok I made the requested corrections and made the pull requests for the power supplies modules, thanks for feedback

PR #2452, opened 2024-11-01
formatting

IvanShires:


harryajc:
Hi @IvanShires
Looking at the spec sheet the S4112F-ON has 12 SFP+ ports you have added 12 Copper 10GB ports. It looks like the S4112T-ON has 12 10gb copper ports if that is what you meant to add.

IvanShires:
fixed, I had meant to do the SFP switch, apologies.

harryajc (approved):

PR #2455, opened 2024-11-04
added part number + comments

IvanShires:
Update AS-1114S-WN10RT.yml with datasheet, 10G RJ NICs, and dedicated BMC port.


IvanShires:
I'm somewhat confused about why the runners failed here....

harryajc (approved):

Issue #2473, opened 2024-11-06
MX7000 redaction


zeridon:
Dell MX7000 chassis seems modeled wrong (https://github.com/netbox-community/devicetype-library/blob/master/device-types/Dell/PowerEdge-MX7000.yaml)

The chassis has 8 device bays (front) for blades.

And 6 back IO modules (IOM) for fabric cards.

harryajc:
Yes your right, the `power-ports` should also be modeled as `module-bays` as they are hot swappable.

Create a PR with these changes and we will review it. There is a guide [here[(https://github.com/netbox-community/devicetype-library/blob/master/CONTRIBUTING.md)

PR #2476, opened 2024-11-07
Correct Dell MX7000 chassis

zeridon:
The Chassis is modeled a bit more faithfully now:

 * Sleds are device bays
 * IOM slots are module bays
 * MM slots are module bays


device-types/Dell/PowerEdge-MX7000.yaml
```diff
@@ -2,9 +2,14 @@
...
-u_height: 7
+part_number: '5112814'
+u_height: 7.0
 is_full_depth: true
 subdevice_role: parent
+airflow: front-to-rear
+weight: 105.0
+weight_unit: lb
+comments: '[Product Page](https://www.dell.com/en-us/shop/ipovw/poweredge-mx7000)'
 power-ports:
```
ravench, line 13: These could also be modules. (And the C14 power port being part of that module)
https://github.com/netbox-community/devicetype-library/issues/2473#issuecomment-2461604726

zeridon: According to: https://www.dell.com/support/manuals/en-us/poweredge-mx7000/pemx7000_ism_pub/back-view-of-the-enclosure?guid=guid-fb31612b-d051-46b0-9ec6-98e5ee1f9057&lang=en-us the PSU's are indeed modules but the power ports are on the chasis itself and not part of the PSU's

device-types/Dell/PowerEdge-MX7000.yaml
```diff
@@ -2,22 +2,62 @@
...
-  - name: PSU.Slot.1
+  - name: PSU.IN.1
     type: iec-60320-c22
-  - name: PSU.Slot.2
+    max_draw: 3000
+    label: 1
```
ravench, line 17: Looks like you need to put this in qoutes for the linting to pass.
```yaml
  label: '1'
```

device-types/Dell/PowerEdge-MX7000.yaml
```diff
@@ -2,22 +2,62 @@
...
+    position: B2
+  - name: IOM-C1
+    position: C1
+  - name: IOM-C2
+    position: C2
+  - name: MM-1
+    position: MM-1
+  - name: MM-2
+    position: MM-2
+  - name: PSU.Slot.1
```
ravench, line 55: What would be cool, of course, is if you could also add the most common PSUs as modules to `module-types/Dell/`.

zeridon: I've added 4 PSU's.

Currently we have only `H7TFG` so the pick was mostly out of number of results/availability. The minimum that needs to be present is one AC (e.g. H7TFG) and one DC (e.g. 0155M8).

harryajc (approved):

PR #2491, opened 2024-11-11
Juniper SRX240H2, SRX240H2-POE device type definitions and images added

cieska:
The front and back images were added extra interface properties were updated and module bays were added.


cieska:
Missing information. I will update it.

PR #2494, opened 2024-11-12
Add Extreme X620 switch series with images and PSUs

aaronaxvig:


aaronaxvig:
Hmm, it is not clear to me why this is failing.  Any help?

aaronaxvig:
> Hmm, it is not clear to me why this is failing. Any help?

Not allowed to have quotes when they aren't needed, apparently.

harryajc (approved):

Issue #2461, opened 2024-11-05
Update the test cases to prevent non-ascii characters in yaml files


ravench:
NetBox appears to not be able to handle non-ascii characters.
This causes issues when importing devices or modules that contain non-ascii characters.
Those are usually dashes (long, short, ect) and symbols like (r) or (tm).

In a shell, the files at issue can be found using `grep`: 

```shell
LC_ALL=C  grep -l "[\x80-\xFF]" ./**/*.yaml 
```

There should be a test for this. There seems to be no applicable test in `pre-commit-hooks` and no flag for `yamlfmt` or `yamllint`, so handling this in a own hook or in pytest seems to be necessary.

Unfortunately, I don't understand pytest well enough to write my own tests, therefore this is an issue and not a pull request.

grische:
@ravench all Netbox instances I have access to do handle non-ASCII characters fine, e.g. in label fields. See #3813

PR #2508, opened 2024-11-15
Fix: Prevent non ascii characters in yaml files issue #2461

Kaioguilherme1:
This commit addresses the issue #2461 by updating the test case to detect non-ASCII characters in YAML files. The new check ensures that if any non-ASCII characters are found, the test will fail and provide a clear message listing the problematic characters.
	•	Added a check to identify non-ASCII characters in the content of the YAML files.
	•	If non-ASCII characters are found, the test will now fail with a detailed error message, including the list of non-ASCII characters.

This update helps ensure that YAML files are strictly ASCII-encoded, avoiding potential encoding issues in different environments.

### DIFF
--- a/tests/definitions_test.py
+++ b/tests/definitions_test.py
@@ -159,6 +159,14 @@ def test_definitions(file_path, schema, change_type):
 
     # Load YAML data from file
     definition = yaml.load(content, Loader=DecimalSafeLoader)
+    
+    # Check for non-ASCII characters
+    non_ascii_chars = [char for char in content if ord(char) > 127]
+    if non_ascii_chars:
+        pytest.fail(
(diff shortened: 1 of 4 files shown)


Kaioguilherme1:
I don’t know why it alters the pickle files when making the commit, since my change was only in the test code.

harryajc:
> I don’t know why it alters the pickle files when making the commit, since my change was only in the test code.

I have seen this a few times now, sometimes the netbox-bot triggers the pickle update after a PR when someone makes a change on their own repo which does cause conflict issues when trying to merge, but not sure why this happens

tests/definitions_test.py
harryajc, line 1: Looks good

harryajc (approved):

PR #2529, opened 2024-11-25
Add Palo Alto PA-400 series new models and rack trays

aaronaxvig:


harryajc:
Hi @aaronaxvig 

Not sure modeling these as child devices and the rack mount kits as the parent device is the correct way about modeling these.

I'm not aware we have modeled any rack mount kits in this repo, therefore I would suggest modeling them the same way as they current PA400 series are

aaronaxvig:
I think parent/child is the correct way to do this.  There is at least one other example of exactly the same thing with the Check Point CPAC-1500/3600/3800-RM-DUAL as parent, SG3600 as child.  I suppose it is easy to say that one is wrong too.

This parent/child modelling technique precisely matches what is described in the Device Bay documentation:

> Child devices are first-class Devices in their own right: That is, they are fully independent managed entities which don't share any control plane with the parent. Just like normal devices, child devices have their own platform (OS), role, tags, and components.

The example given in that documentation is a 2U chassis with four blade servers.  My 1U "chassis" happens to not provide power to the two mounted devices (does provide mounting spaces for the power supplies actually, so kind of...) but is otherwise very much the same scenario.

Finally, this parent/child technique does accurately model what is in the rack and how the devices are mounted, and does not interfere with any other modelling that I can see someone else doing.  The devices are not capable of being rack mounted on their own so will always be child devices.  The rack kit would never be used as anything but a parent.

aaronaxvig:
Relevant recommendation to model like this: https://github.com/netbox-community/netbox/issues/9320#issuecomment-1121250281

aaronaxvig:
I believe I have supplied good reasoning why this should be accepted as it.  Could this be merged, or additional commentary provided as to why not?

aaronaxvig:
Can't believe I missed this example (right in the same folder) but the PA-220 firewall and rack kit are modeled with child / parent relationships.  Pull request #1499 is where it was added and has substantial discussion regarding the best way to model this sort of thing:

> Alright so lets talk about the rack mount adapter. There is already a defined way to do this. What you need to do is create a device-type for the rack mount and ensure it has subdevice_role: parent. Then, in the device in question (the PA-220 in this case), set subdevice_role: child and ensure it has u_height: 0

harryajc:
> Can't believe I missed this example (right in the same folder) but the PA-220 firewall and rack kit are modeled with child / parent relationships. Pull request #1499 is where it was added and has substantial discussion regarding the best way to model this sort of thing:
> 
> > Alright so lets talk about the rack mount adapter. There is already a defined way to do this. What you need to do is create a device-type for the rack mount and ensure it has subdevice_role: parent. Then, in the device in question (the PA-220 in this case), set subdevice_role: child and ensure it has u_height: 0

Hi @aaronaxvig 

Thanks for that having a look over that PR and will review the PR again now

harryajc (approved):

PR #2558, opened 2024-12-11
Add Extreme 5520 switches

aaronaxvig:


danner26:
Hello, before I go and comment this on all of the new files, are your `U1` and `U2` ports actually module bays? I.E. are they swappable for the different port configuration you mention? If so, these should be `module-bays` rather than `interfaces`

harryajc (approved):

PR #2565, opened 2024-12-15
Added Netgear GS305E Device Type

qmph22:
Created a YAML file for the Netgear GS305E device type


qmph22 (commented):
Looks good even without the management interface. Verified that only the LF line ending is present

harryajc (approved):

PR #2570, opened 2024-12-16
changes manufacturer "Tripp Lite" to "Eaton"

gurubert:
Tripp Lite is a brand of Eaton, files are in the Eaton directory

### DIFF
--- a/device-types/Eaton/Tripp-Lite-B051-000.yaml
+++ b/device-types/Eaton/Tripp-Lite-B051-000.yaml
@@ -1,5 +1,5 @@
 ---
-manufacturer: Tripp Lite
+manufacturer: Eaton
 model: Tripp Lite B051-000
 slug: tripp-lite-b051-000
 part_number: B051-000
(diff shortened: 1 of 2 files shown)


harryajc:
HI @gurubert 
The slugs also need updating to follow the standard naming

danner26:
I seem to see these are still branded as Tripp Lite, is that correct?

If so, we have reviewed this in the past, and if the manufacturer/brand is still Tripp Lite (even if the owning company is Eaton), we will leave the branding as Tripp Lite unless they change the branding completely to Eaton

If there is an identical product number for Eaton, then this should be duplicated into the Eaton folder. Let me know if you have any questions

danner26:
I am going to reject and close this PR for the reason stated above. If you feel this is incorrect, please feel free to inform us as to why and reopen. Thank you.

gurubert:
I have now moved the two files. The importer https://github.com/netbox-community/Device-Type-Library-Import takes the manufacturer name from the directory. The directory name has to match.

PR #2581, opened 2024-12-20
Add new deviced for LANCOM and QNAP

JanisLinde:
Add new device-types

LANCOM:

- 1793VAW
- 1793VA
- 1926VAG
- Rack-Mount

QNAP:
- TS-431U

Added subdevice_role "child" to LANCOM-Devices to allow usage of the Rack-Mount


JanisLinde:
The tests fails because of a missing power source of the LANCOM/Rack-Mount. But this device-type don't has a power source.

harryajc (approved):

PR #2590, opened 2024-12-27
Add Extreme X690 and X695 switches

aaronaxvig:


danner26 (approved):

aaronaxvig:
Sorry about the chain of commits.  I couldn't figure out what the original issue was from the given errors, so took a shot in the dark.  It was close but in a way was the opposite of what I needed to fix.

Seems to be good now though.

danner26:
Thanks @aaronaxvig I will re-review. In the future you can use `pre-commit run -a` to test, which will give you the actual errors. This is defined in the readme if you need to reference it

danner26 (approved):

PR #2605, opened 2025-01-04
added device-types and module-types

Mr-Nobody-vie:
added device-types and module-types, some generic and others from TP-Link.

Please use the same importer in the webfrontend and in your pull requests .. or pin a message at the top of the forum with instructions "for idiots".

BUT the best thing would be - use a file format for adults! --> JSON

The childish stuff with YAML cost me several hours.
This was my first and last contribution to the Netbox community.
It's a shame that such big hurdles are being put in place here.


harryajc:
Hi @Mr-Nobody-vie 
Thank you for your contribution, however we are currently not modeling Storage, memory or CPU modules in this repository.

PR #2606, opened 2025-01-04
Create 7210-SAS-D-ETR.yaml

jaco1982:
Added Nokia 7210 SAS-D ETR


device-types/Nokia/7210-SAS-D-ETR.yaml
```diff
@@ -0,0 +1,38 @@
+---
+manufacturer: Nokia
+model: 7210 SAS-D ETR
+slug: nokia-7210-sas-d-etr
+part_number: 3HE05676AC
+u_height: 1
```
harryajc, line 6: Hi,
Thank you for contributions, to make sure we have as complete data as possible could you add the `weight` and `weight_unit` properties for this device.

Many Thanks

harryajc (changes requested):

device-types/Nokia/7210-SAS-D-ETR.yaml
jaco1982: I've added the indicated values.

harryajc (approved):

PR #2615, opened 2025-01-07
Add new Arista CCS-720XP model

daven:
This PR adds a CCS-720XP model with rear-to-front airflow.


device-types/Arista/CCS-720XP-48Y6-R.yaml
```diff
@@ -0,0 +1,229 @@
...
+front_image: true
+rear_image: false
+interfaces:
+  - name: Management1
+    type: 1000base-t
+    mgmt_only: true
+  - name: Ethernet1
+    type: 1000base-t
+    poe_mode: pse
+    poe_type: type1-ieee802.3af
```
harryajc, line 29: Looking at the spec sheet these are 30W POE+ interfaces therefore you need to use `poe_type: type2-ieee802.3at` instead

harryajc (changes requested):

device-types/Arista/CCS-720XP-48Y6-R.yaml
daven: Corrected this

harryajc (approved):

PR #2617, opened 2025-01-08
Add Cisco Catalyst 1300 series SMB Switches

joeladria:
This contribution adds the entire series of Cisco C1300 switches. 

I [created some scripts](https://github.com/joeladria/ciscosmb-devicetype-generator) to generate the images and files which may be useful for other switch series in the future.

* Includes Front and Back images from Cisco BX, where available.
* Uses stacking nomenclature (eg. `GigabitEthernet1/0/1`) where on applicable switches, but not on the non-stacking units which typically use the base naming (eg. `GigabitEthernet1`).
* If there is interest, I could produce something similar for the Catalyst 1200, CBS250 and CBS350 using the scripts
* I tested the import ad-hoc and using a [patched version of the import plugin](https://github.com/chatasos/Device-Type-Library-Import-Updated/tree/develop) as there are still issues with the current version available with Netbox 4.x, and they seemed to import well this way


harryajc:
Hi @joeladria 

Thank you for your contribution, to make sure that the data we have is as complete as possible could you add the `airflow` property to all these devices?

harryajc (approved):

PR #2622, opened 2025-01-09
dell-unity-xt-380

zedaprime:
Adding a template for Dell Unity XT 380 Storage Array

### DIFF
--- a/device-types/Dell/Unity-XT-380.yaml
+++ b/device-types/Dell/Unity-XT-380.yaml
@@ -1,0 +1,36 @@
+---
+manufacturer: Dell
+model: Unity XT 380
+slug: dell-unity-xt-380
+part_number: 100-566-075-03
+u_height: 2
+is_full_depth: true
+subdevice_role: parent
+interfaces:
+  - name: SP-A ETH0
+    type: 1000base-t
+    mgmt_only: true
+  - name: SP-B ETH0
+    type: 1000base-t
+    mgmt_only: true
+  - name: SP-A I/O Module 1 Eth Port 0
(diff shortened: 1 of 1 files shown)


zedaprime:
Adding a template for Dell Unity 380 Storage Array.  Using the existing Unity 880F as a base.  

Ports are labeled using the GUI Dashboard names for interfaces.

device-types/Dell/Unity-XT-380.yaml
```diff
@@ -0,0 +1,36 @@
...
+is_full_depth: true
+subdevice_role: parent
+interfaces:
+  - name: SP-A ETH0
+    type: 1000base-t
+    mgmt_only: true
+  - name: SP-B ETH0
+    type: 1000base-t
+    mgmt_only: true
+  - name: SP-A I/O Module 1 Eth Port 0
```
harryajc, line 16: By the looks these are storage controllers therefore, you need to added `device-bays` then create the controller as a separate device with the interfaces for that controller on.

device-types/Dell/Unity-XT-380.yaml
```diff
@@ -0,0 +1,36 @@
+---
+manufacturer: Dell
+model: Unity XT 380
+slug: dell-unity-xt-380
+part_number: 100-566-075-03
+u_height: 2
+is_full_depth: true
+subdevice_role: parent
```
harryajc, line 8: To make sure we have the most complete information, can you add the following parameters, `airflow` `weight` and `weight_unit`

harryajc (changes requested):

PR #2624, opened 2025-01-10
Adding Cisco N540X-6Z18G-SYS-D.

t2ylorgang:
Added Cisco N540X-6Z18G-SYS-D to device types. As far as the interfaces, I set them up to be 1000BASET and 10BASE X SFP+.

According to this documentation:
 https://www.cisco.com/c/en/us/td/docs/iosxr/ncs5xx/hardware/installation/guide/b-ncs540-small-density-routers-hig/m-overview-ncs540-small-density-routers.html#id_75425

"The Cisco N540X-6Z18G-SYS-A/D and N540-6Z18G-SYS-A/D routers have the following network interfaces:
- 18 x 1G SFP+ ports
- 6 x 1G/10G SFP+ ports"

My file has the "1G SFP+ ports" as **1000base-t** and the "1G/10G SFP+ ports" as **10gbase-x-sfpp**

I feel this was the best way to represent a real world application of this router. Obviously, these interface types can change depending on a real world application, but this is a good template IMO.

I can change these to be the predefined SFP ports in Netbox - SFP (1GE) and SFP+ (10GE) if the reviewer thinks this is a better way. Netbox does not have a predefined interface type for a 1/10GE port : (


device-types/Cisco/N540X-6Z18G-SYS-D.yaml
```diff
@@ -0,0 +1,82 @@
...
+    description: 'USB A console interface.'
+  - name: TOD
+    type: rj-45
+    description: 'Time of Day synchronization port.'
+  - name: ALM
+    type: rj-45
+    description: 'Alarm signaling port.'
+interfaces:
+  - name: GigabitEthernet0/0/0/0
+    type: 1000base-t
```
harryajc, line 33: For these 1G SFP ports they need to be modeled using `1000base-x-sfp`

harryajc (changes requested):
Please also fix the issues detected by the pre-commit checks

PR #2632, opened 2025-01-14
XL710BM1-4SP

ravench:
Add FS XL710BM1-4SP network card based on X710-DA4

### DIFF
--- a/module-types/FS/XL710BM1-4SP.yaml
+++ b/module-types/FS/XL710BM1-4SP.yaml
@@ -1,5 +1,5 @@
 ---
-manufacturer: Intel
+manufacturer: FS
 model: XL710BM1-4SP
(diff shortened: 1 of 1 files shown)


module-types/FS/XL710BM1-4SP.yaml
```diff
@@ -0,0 +1,19 @@
+---
+manufacturer: Intel
```
harryajc, line 2: As this is an Intel card this should be located under `module-types/Intel`

harryajc (changes requested):

module-types/FS/XL710BM1-4SP.yaml
ravench: Fixed, thanks

harryajc (approved):

PR #2638, opened 2025-01-16
Add PowerScale F710

Cautious-Software701:
Added the PowerScale F710. This is technically a "node" which forms a PowerScale cluster. If it should be done another way please let me know!

Thanks


device-types/Dell/PowerScale-F710.yaml
```diff
@@ -0,0 +1,42 @@
...
+airflow: front-to-rear
+front_image: true
+rear_image: true
+weight: 22.5
+weight_unit: kg
+console-ports:
+  - name: Console
+    type: de-9
+power-ports:
+  - name: PS1
```
harryajc, line 18: Thank you for contribution. As these are removable PSUs they should be modeled as `module-bays` and then add the required PSU modules in the modules folder if they don't already exist.

harryajc (changes requested):

harryajc (approved):

PR #2653, opened 2025-01-18
Add Supermicro AS-1115SV-WTNRT

mitdfn:
* Add new device type for server Supermicro AS-1115SV-WTNRT
* Add elevation images for AS-1115SV-WTNRT


device-types/Supermicro/AS-1115SV-WTNRT.yml
```diff
@@ -0,0 +1,35 @@
...
+comments: "[User's Manual](https://www.supermicro.com/manuals/superserver/1U/MNL-2649.pdf)<br />[Datasheet](https://www.supermicro.com/en/products/system/datasheet/as-1115sv-wtnrt)"
+u_height: 1
+is_full_depth: true
+weight: 9.53
+weight_unit: kg
+airflow: front-to-rear
+console-ports:
+  - name: Serial
+    type: de-9
+power-ports:
```
harryajc, line 15: As these are removable PSUs they should be modeled as `module-bays` and then add the required modules if they don't already exist

harryajc (changes requested):

device-types/Supermicro/AS-1115SV-WTNRT.yml
mitdfn: got it, will do

PR #2664, opened 2025-01-21
Create Check Point SG5800.yaml

walday:


device-types/Check Point/SG5800.yaml
```diff
@@ -0,0 +1,65 @@
...
+comments: '[Check Point 5800 Data Sheet](https://www.checkpoint.com/downloads/products/5800-security-gateway-datasheet.pdf)'
+airflow: side-to-rear
+console-ports:
+  - name: ttyS0
+    label: CONSOLE
+    type: rj-45
+  - name: ttyS1
+    label: CONSOLE
+    type: usb-c
+power-ports:
```
harryajc, line 19: By the looks of the spec sheet these are replaceable PSUs, therefore they should be modeled as `module-bays` and then add the required PSUs as modules if they don't exist already

harryajc (changes requested):

harryajc (approved):

PR #2672, opened 2025-01-22
Add APC ATSs

aaronaxvig:


harryajc (changes requested):
To make sure that we have the most accurate information possible can you add the `airflow` property to all these devices

harryajc (approved):

PR #2685, opened 2025-01-26
Add ubiquiti unifi security gateway pro images

phillf:


phillf:
@harryajc, thanks for that.

harryajc:
> @harryajc, thanks for that.

It doesn't seem to let me remove the HP images if you can do that it would be brilliant

harryajc (approved):

phillf:
> > @harryajc, thanks for that.
> 
> It doesn't seem to let me remove the HP images if you can do that it would be brilliant

Done.

PR #2698, opened 2025-01-28
Brocade icx7450 32zp

ColtonDx:
Added the Brocade ICX 7450 and the Power Supply and Interface Modules


device-types/Brocade/ICX-7450-32ZP.yaml
```diff
@@ -0,0 +1,102 @@
...
+weight_unit: kg
+airflow: front-to-rear
+is_full_depth: false
+u_height: 1
+interfaces:
+  - name: GigEthernetmgmt1
+    type: 1000base-t
+    mgmt_only: true
+  - name: GigabitEthernet1/1/1
+    type: 1000base-t
```
harryajc, line 16: Looks like these are POE interfaces therefore you need to add the required lines for each POE interface
`poe_mode: pse`
`poe_type: type2-ieee802.3at`
Note these are for standard POE+ the spec sheet says some can provide POE++ which is `type4-ieee802.3bt`

harryajc (changes requested):

device-types/Brocade/ICX-7450-32ZP.yaml
ColtonDx: Added POE interface types and converted the 1/2/x interfaces to a module since they are a removable module

harryajc (approved):

PR #2731, opened 2025-02-07
Adding Legrand J06B0B

danner26:


device-types/Legrand/J06B0B.yaml
```diff
@@ -0,0 +1,28 @@
+---
+manufacturer: Legrand
+model: J06B0B
+slug: legrand-j06b0b
+part_number: 786776043884
+u_height: 1
+is_full_depth: false
```
harryajc, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

harryajc (approved):

PR #2738, opened 2025-02-10
Add LANCOM 1803VA-4G

iSnackyCracky:


iSnackyCracky:
I don't understand why the format validation failed - I only see a few thousand "Done" lines and then it exits with code 1?

Can anyone help me out here?

harryajc (approved):

PR #2747, opened 2025-02-11
Create softline_sl911.yaml

walday:


device-types/Softline/softline_sl911.yaml
```diff
@@ -0,0 +1,31 @@
+---
+manufacturer: Softline
+model: Softline_SL911
+slug: softline_sl911
+u_height: 2
+weight: 20
+weight_unit: kg
+is_full_depth: true
+airflow: front-to-rear
+comments: '2U 8x3.5 bay Storage, Inwin RS208-02M-CR550.H case'
```
harryajc, line 10: Have you got a spec sheet for this. if this is just a Inwin RS208-02M-CR550.H then it probably should be modeled as one of those

device-types/Softline/softline_sl911.yaml
```diff
@@ -0,0 +1,31 @@
...
+power-ports:
```
harryajc, line 11: Looking at the spec sheet for this case, is has removable PSUs therefore they should be modeled as `module-bays` and then add the required psus as modules.

harryajc (changes requested):

device-types/Softline/softline_sl911.yaml
walday: > Looking at the spec sheet for this case, is has removable PSUs therefore they should be modeled as `module-bays` and then add the required psus as modules.

Add PSU and module-bays

harryajc (approved):

PR #2763, opened 2025-02-12
added kemp-lm-x1 device

rusnino:
Introducing Kemp Technologies LoadMaster X1 device


device-types/Kemp Technologies/LM-X1.yaml
```diff
@@ -0,0 +1,31 @@
...
+slug: kemp-technologies-lm-x1
+part_number: LM-X1
+u_height: 1
+is_full_depth: false
+weight: 3.77
+weight_unit: kg
+comments: '[Kemp Technologies LoadMaster X1](https://kemptechnologies.com/server-load-balancing/loadmaster-x1)'
+front_image: true
+rear_image: false
+power-ports:
```
harryajc, line 13: Thank you for your contribution. 
Looking at the spec sheet these are hot swappable PSUs therefore they need to modeled under `module-bays` and then add the required PSU modules.

harryajc (changes requested):

device-types/Kemp Technologies/LM-X1.yaml
rusnino: My bad, checked again and found that for LM-X1 there is just 1 PSU and it not hot swappable(expected) with device max power supply (W): 65.

harryajc (approved):

PR #2768, opened 2025-02-13
Add Dell Poweredge R530 + R540 elevation images

fahhem:


harryajc:
HI @fahhem 

Thank you for your contribution, so that the images link to the device you need to add the `front_image: true` and `rear_image: true` to the devices yaml file

harryajc (approved):

PR #2773, opened 2025-02-13
Update sel-2245-2-dc-analog-input-module.yml

walday:
Drop extra long description


harryajc:
Hi @walday 

If you are doing multiple small changes these can be combined into a single PR, up to 5 devices per PR is preferable. That way it saves a lot of time when reviewing them

Thanks
Harry

harryajc (approved):

PR #2791, opened 2025-02-15
Arista DCS-7280SE-68-F

dorkmatt:


device-types/Arista/DCS-7280SE-68-F.yaml
```diff
@@ -0,0 +1,130 @@
...
+    type: 10gbase-x-sfpp
+  - name: Ethernet48
+    type: 10gbase-x-sfpp
+# QSFP28(100G) ports can be broke into 4x25 with a model of [49-54]/[1-4]
+  - name: Ethernet49/1
+    type: 100gbase-x-qsfp28
+  - name: Ethernet50/1
+    type: 100gbase-x-qsfp28
+power-ports:
+  - name: PS1
```
harryajc, line 123: Looking at the images provided these are hot swappable PSUs, therefore they need to modeled under `module-bays` with the required PSU models added if they don't already exist

harryajc (changes requested):

device-types/Arista/DCS-7280SE-68-F.yaml
dorkmatt: Updated

harryajc (approved):

PR #2809, opened 2025-02-18
Update Tripp-Lite-B051-000.yaml

walday:


harryajc:
This device has been discussed before please see #2570

PR #2818, opened 2025-02-19
Update proliant-xl170r-g9-1u.yaml

walday:


walday:
Fix Error {"u_height":["Child device types must be 0U."]} creating device type: HPE ProLiant XL170r G9 (1U)

harryajc (approved):

PR #2831, opened 2025-02-19
Create storeever-msl6480.yaml

walday:


device-types/HPE/storeever-msl6480.yaml
```diff
@@ -0,0 +1,42 @@
...
+  - name: LTO_Drive_3
+    position: '3'
+  - name: LTO_Drive_4
+    position: '4'
+  - name: LTO_Drive_5
+    position: '5'
+  - name: LTO_Drive_6
+    position: '6'
+power-ports:
+  - name: PSU1
```
harryajc, line 31: Are these modular psus? If so they need to be modeled as module bays

harryajc (changes requested):

walday:
psus bay added

harryajc (approved):

PR #2840, opened 2025-02-20
Add Bachmann 9x power socket

jhofmueller:
Please add the new device type.


Napsty:
Note that Bachmann is not an independent vendor anymore, they were bought by Max Hauri. Maybe add it under Max Hauri? see https://github.com/netbox-community/devicetype-library/blob/master/device-types/Max%20Hauri/bachmann-107686.yaml

harryajc (approved):

PR #2847, opened 2025-02-20
Adding AT-MC1008/SP

danner26:


harryajc:
Hi @danner26 
Is this a normal media converter? Then shouldn't it be a device type rather than the module type?

harryajc:
Stale PR

PR #2856, opened 2025-02-21
Add ThinkSystem SR635

psuet:


device-types/Lenovo/ThinkSystem-SR635.yaml
```diff
@@ -0,0 +1,17 @@
...
+part_number: 7Y98
+u_height: 1
+is_full_depth: true
+subdevice_role: parent
+interfaces:
+  - name: XCC
+    type: 1000base-t
+    mgmt_only: true
+power-ports:
+  - name: PSU0
```
harryajc, line 14: As these are Modular PSUs they need to be modeled as `module-bays` and then add the psu's in the modules folder if they don't already exist.

device-types/Lenovo/ThinkSystem-SR635.yaml
```diff
@@ -0,0 +1,17 @@
+---
+manufacturer: Lenovo
+model: ThinkSystem SR635
+slug: lenovo-thinksystem-sr635
+part_number: 7Y98
+u_height: 1
+is_full_depth: true
```
harryajc, line 7: To make sure we have the most complete information, can you add the following parameters, `airflow` `weight` and `weight_unit` with the required data

device-types/Lenovo/ThinkSystem-SR635.yaml
```diff
@@ -0,0 +1,17 @@
...
+subdevice_role: parent
```
harryajc, line 8: Subdevice_role: parent is not required as this has no 'device-bays'

harryajc (changes requested):

harryajc:
Added through a different PR

PR #2864, opened 2025-02-23
Added Cisco ASR 1004 Router and its elevation-images

AskskwBv8T2nrm4Qnj:
Device-link: https://www.cisco.com/c/en/us/products/routers/asr-1004-router/index.html


device-types/Cisco/ASR1004.yaml
```diff
@@ -0,0 +1,29 @@
...
+rear_image: true
+is_full_depth: true
+airflow: front-to-rear
+console-ports:
+  - name: Con 0
+    type: rj-45
+  - name: Aux 0
+    type: rj-45
+power-ports:
+  - name: ASR1004-PWR-AC-1
```
harryajc, line 20: Looks like these PSUs are modules, therefore they need to be modeled as `module-bays`

device-types/Cisco/ASR1004.yaml
```diff
@@ -0,0 +1,29 @@
...
+    type: iec-60320-c14
+    maximum_draw: 765
+  - name: ASR1004-PWR-AC-2
+    type: iec-60320-c14
+    maximum_draw: 765
+interfaces:
```
harryajc, line 26: Would also be a good idea to add the line cards as `module-bays`

harryajc (changes requested):
Just a few minor things

device-types/Cisco/ASR1004.yaml
AskskwBv8T2nrm4Qnj: Good call, I sleep-walked past that one! I added an updated version.

harryajc (approved):

PR #2866, opened 2025-02-24
QNAP: add TS-2483XU-RP and PWR-PSU-800W-DT01

konimex:
## Description

Self-explanatory.


konimex:
Closing - will make a new one from a branch instead of master.

PR #2873, opened 2025-02-24
Add module types for Cisco UCS-X 9508 chassis

SolidElectronics:
The Cisco UCS-X 9508 chassis has module slots defined, but no modules.  This PR adds some of them to the repo.


SolidElectronics:
One thing to note, CONTRIBUTING.md says to "End each definition file with a blank line", but the pre-commit hook will fail if there are any blank lines.  Is that a mistake, or am I misinterpreting something?

SolidElectronics:
One thing to note, CONTRIBUTING.md says to "End each definition file with a blank line", but the pre-commit hook will fail if there are any blank lines.  Is that a mistake, or am I misinterpreting something?

harryajc (approved):

PR #2887, opened 2025-02-27
HPE: Modify Synergy 12000 and move several modules

konimex:
## Description

Self-explanatory.


konimex:
I'm not sure why it would fail the test since the files are moved from `device-types` to `module-types`.

mwilck:
I have a question about this change. I'm a netbox beginner, so please forgive me if it's dumb.

IIUC, changing these objects from "device" to "module" has the effect that the intefaces associated with the synergy interconnects show up as interfaces on the synergy frame itself. But the frame itself doesn't have any interfaces except the management interface.

The interconnects are independent devices acting (more or less) like a switch. While all ports of a single interconnect are physically connected (with SW-defined filtering between ports), ports of different interconnects are not, unless stacking cables are used to combine the interconnects.

harryajc:
Hi All,

If these are their own separate devices then they should stay as devices and not modules

mwilck:
So how to proceed here? Maybe @konimex can provide a rationale why he tought that the interconnects should be represented as modules rather than devices?

konimex:
Apologies for the late reply since I was on vacation. Also a NetBox beginner (and in fact also a beginner at these Synergy devices).

I was having an impression that these interconnects are ultimately network devices, but looking at #1545, it seems I was mistaken. Closing this PR.

PR #2888, opened 2025-02-27
Create WS-C4948E.yaml

ferrum90:
Add Cisco Catalyst WS-C4948E


device-types/Cisco/WS-C4948E.yaml
```diff
@@ -0,0 +1,123 @@
+---
+manufacturer: Cisco
+model: Catalyst 4948E
+slug: cisco-ws-c4948e
+part_number: WS-C4948E
```
harryajc, line 5: Thank you for your contribution.
To make sure we have the most complete information possible please add the `weight` property and associated `weight_unit` property, and also the `airflow` property.

harryajc (changes requested):

PR #2911, opened 2025-03-06
Add DS916+, DS1512+, DS1812+ and DS1815+ Synology devices

revoltek-daniel:
Add DS916+, DS1512+, DS1812+ and DS1815+ Synology devices


device-types/Synology/DS1512+.yaml
```diff
@@ -0,0 +1,18 @@
+---
+manufacturer: Synology
+model: DS1512+
+slug: synology-ds1512-plus
+part_number: DS1512+
+u_height: 4
```
harryajc, line 6: Hi, Thank you for your contribution.
As these are rack mount devices then they should have a u_height of 0

device-types/Synology/DS1512+.yaml
```diff
@@ -0,0 +1,18 @@
...
+comments: '[Synology DS1512+ Specs](https://www.synology.com/en-us/products/DS1512+#specs)'
+weight: 4.25
+weight_unit: kg
+power-ports:
```
harryajc, line 10: To make sure we have the most complete information possible please add the `airflow` parameter.

harryajc (changes requested):
Please complete these changes for all devices in this PR

device-types/Synology/DS1512+.yaml
revoltek-daniel: They are not rack mountable, if copied this from https://github.com/netbox-community/devicetype-library/blob/master/device-types/Synology/DS1817%2B.yaml#L6

harryajc: Sorry that was a typo, it was supposed to say as these aren't rack mountable they should have a u_height of 0.

revoltek-daniel: So I should fix them on all DiskStations? Because all Synology Models prefixed with DS are not rack mountable and have currently a height of 4 defined.

harryajc (approved):

PR #2914, opened 2025-03-06
Add CyberPower OR600ELCDRM1U ups

revoltek-daniel:
Add CyberPower OR600ELCDRM1U ups


device-types/CyberPower/OR600ELCDRM1U.yaml
```diff
@@ -0,0 +1,42 @@
+---
+manufacturer: CyberPower
+model: OR600ELCDRM1U
+slug: cyberpower-or600elcdrm1u
+part_number: or600elcdrm1u
+u_height: 1
+is_full_depth: false
+weight: 8.1
+weight_unit: kg
```
harryajc, line 9: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

PR #2932, opened 2025-03-10
Added images and more details to C1111-8P [✅ passed] 

fknorn:


elevation-images/Cisco/cisco-isr-1111-8p.front.png
harryajc, line 1: Hi Thank you for your contribution. There should be no white space around the device. Please crop the remaining white/blank space from the image.

elevation-images/Cisco/cisco-isr-1111-8p.rear.png
harryajc, line 1: Hi Thank you for your contribution. There should be no white space around the device. Please crop the remaining white/blank space from the image.

harryajc (changes requested):

device-types/Cisco/ISR-1111-8P.yaml
```diff
@@ -34,4 +39,4 @@ console-ports:
 power-ports:
   - name: PSU0
     type: iec-60320-c14
```
harryajc, line 41: Also for completeness this should be 'type: dc-terminal'

harryajc (changes requested):

elevation-images/Cisco/cisco-isr-1111-8p.front.png
fknorn: Hi Harry,

thanks for your diligence. I assume there has already been some thought / discussion put into this (i.e. regarding devices that are not 19" wide, to be put into 19" racks). I have actually added that transparent padding on purpose, so that – in the rack-view – the device is shown at the correct scale. Maybe netbox could switch to not only fitting the devices width into the horizontal "slot", but also the height?

Anyway, here's a comparison-view with and without padding:

Screenshot 2025-03-11 at 09 41 02

So you insist on removing the padding?

device-types/Cisco/ISR-1111-8P.yaml
fknorn: Done.

elevation-images/Cisco/cisco-isr-1111-8p.front.png
harryajc: Hi @fknorn,

Actually not thought about it like that, from a useability the padding definitely looks better, so happy to keep it as is

fknorn: @harryajc - thanks! But, well, now let me be the doubter.

I guess in terms of future perspectives, the _cropped_ version of the picture would actually be the best to have. So while having padding might improve display _today_, maybe it would be wiser to tweak netbox's own handling of images that are "too narrow" by rather fitting the _height_ of the image into the slot, rather than have it cover the slot. This could be done by changing `slice` to `meet` in the `preserveAspectRatio` attribute of the corresponding `<image>` tag.

**Currently implemented (preserveAspectRatio -> `slice`)**
Screenshot 2025-03-12 at 08 56 06

**Possible solution (preserveAspectRatio -> `meet`)**
Screenshot 2025-03-12 at 08 55 55


However, I'm not sure what the ramifications are and how to properly propose or indeed implement this...

harryajc: Hi @fknorn,
I defiantly recon this is worth raising on the main netbox repository, as how images are handled is all controlled from the main netbox application

fknorn: Done: https://github.com/netbox-community/netbox/issues/18876

PR #2943, opened 2025-03-13
Aruba CX devices

Juniper46:
Added missing CX6200M models and updated poe interfaces on other CX models


harryajc (approved):

rusnino:
Hi @Juniper46 @harryajc 
Aruba added new (updated) switches line with slightly changed Part Number:
instead of R8N87A they sell now R8N87B. (this is not only case, I see other devices changed suffix `A` to suffix `B`)

What do you think, how it can be managed in `devicetype-library` ?

PR #2955, opened 2025-03-21
Create sdwan-ce50n-4t1sm.yaml

walday:


device-types/BiZone/sdwan-ce50n-4t1sm.yaml
```diff
@@ -0,0 +1,30 @@
...
+part_number: sdwan-ce50n-4t1sm
+u_height: 1.0
+is_full_depth: false
+weight: 1.4
+weight_unit: kg
+comments: '[SD-wan Data Sheet](https://bi.zone/catalog/products/secure-sd-wan/?tb=tab4)'
+airflow: passive
+power-ports:
+  - name: PSU 12v 3a
+    type: iec-60320-c14
```
harryajc, line 14: If the PSU is 12v then this should be `dc-terminal`

harryajc (changes requested):

PR #2970, opened 2025-04-01
create S5732-H24S6Q.yaml

trnki:
added new huawei device model


device-types/Huawei/S5732-H24S6Q.yaml
```diff
@@ -0,0 +1,88 @@
...
+is_full_depth: true
+airflow: front-to-rear
+weight: 8.9
+weight_unit: kg
+console-ports:
+  - name: USB
+    type: usb-a
+  - name: console0
+    type: rj-45
+power-ports:
```
harryajc, line 16: Hi Thank you for contribution,
Looking at the spec sheet these are hot swappable PSUs therefore they should be modelled as `module-bays` within the device template with the required PSUs added as modules

harryajc (changes requested):

harryajc (approved):

PR #2985, opened 2025-04-10
Create Dorado-5000-v6.yaml

walday:


harryajc:
Hi @walday
The correct way to model these would be:
- Parent Device (chassis)
- Child Device (controller with the controllers interfaces and any expansion modules like network cards as module bays)

PSUs can stay as module bays with the parent chassis

walday:
> Hi @walday The correct way to model these would be:
> 
> * Parent Device (chassis)
> * Child Device (controller with the controllers interfaces and any expansion modules like network cards as module bays)
> 
> PSUs can stay as module bays with the parent chassis

ok, fix it

harryajc (approved):

PR #2993, opened 2025-04-10
Create 1288H-V7.yaml

walday:


device-types/Huawei/1288H-V7.yaml
```diff
@@ -0,0 +1,31 @@
+---
+manufacturer: Huawei
```
harryajc, line 2: Given these are manufactured by Xfusion they should be modeled under xfusion.

harryajc (changes requested):

device-types/Huawei/1288H-V7.yaml
walday: ok, fix it

harryajc (approved):

PR #2994, opened 2025-04-10
Create 2288H-V7.yaml

walday:


device-types/Huawei/2288H-V7.yaml
```diff
@@ -0,0 +1,30 @@
+---
+manufacturer: Huawei
```
harryajc, line 2: Given these are manufactured by Xfusion they should be modeled under xfusion.

harryajc (approved):

PR #3013, opened 2025-04-15
Add Juniper srx2300

twilley:
Adds the Juniper SRX2300 device-type and images.


device-types/Juniper/SRX2300.yaml
```diff
@@ -0,0 +1,75 @@
...
+    type: 25gbase-x-sfp28
+  - name: et-0/2/2
+    type: 25gbase-x-sfp28
+  - name: et-0/2/3
+    type: 25gbase-x-sfp28
+  - name: et-0/3/0
+    type: 100gbase-x-qsfp28
+  - name: et-0/3/1
+    type: 100gbase-x-qsfp28
+power-ports:
```
harryajc, line 66: Thank you for your contribution,

As the PSUs are field replaceable units they should be modeled as `module-bays` then add the required PSU modules if they don't already exist

harryajc (changes requested):

device-types/Juniper/SRX2300.yaml
twilley: Will do. Thanks.

twilley: I made the changes you requested.

harryajc (approved):

PR #3033, opened 2025-04-23
Create san24b-6.yml

gainedouze:
This is a new device from IBM - SANB24-6
It is a FC switch comon in datacenter worldwide


device-types/IBM/san24b-6.yml
```diff
@@ -0,0 +1,67 @@
+---
+manufacturer: Ibm
+model: SAN24B-6
+slug: ibm-san24b-6
+part_number: 0089690000F24
+is_full_depth: false
+u_height: 1
```
harryajc, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

harryajc (approved):

PR #3044, opened 2025-04-28
Rework of HPE proliant DL360 Gen10 server with related module-bays and power supplies

TerraTalpi:
I have completely worked over the existing **HPE DL360 gen10** server device-type. The servers may ship **without** the **default 4-gbit ethernet NICs** if configured with the vendor/seller. The default NICs are anyway shipped with the onboard LOM port, so they should be labeled as such as well. The interfaces are just numbered in the ILO's webinterface, therefore i have just given them numbers (including the position/module) too.

I have added descriptions for each module in order to identify the ports easier when configuring. I have also added some NICs that I can verify for the respective PCIe, LOM or ALOM (Advanced LOM / Flexible LOM) ports. That way it should be easy when creating the server, to then just configure the LOM adapter and add it to the device. I have taken the descriptions for the NICs from the HPE ILO adapter web interface. I have also adapted the power supply bay description to match the ILO web interface naming, including a description for the actual physical port.

The image that I have added for the front panel of the server has been taken from [www.visiocafe.com](https://www.visiocafe.com/hpe.htm), [direct link](https://www.visiocafe.com/downloads/hp/HPE-Recent_2025-04-03.zip)


TerraTalpi:
Thank you very much @harryajc, I was just about to fix the open issues just now, but you were faster :smiley:. Thanks!

PR #3050, opened 2025-04-29
adding Watchguard T45

dieck:
adding Watchguard T45, and images incl. one with 19" rackmount kit


dieck:
Hi!
Sorry for the delay, I fixed the noted issues.
Thank you for mentioned the device-bay option, I was not aware of that, certainly the best way to do it.

harryajc (approved):

PR #3054, opened 2025-04-29
adding HPE SNQ1100Q 16Gb 2p FC HBA P9D94A

dieck:


harryajc:
@dieck  You haven't added any files only deleted 1, not sure why a file has been deleted?

PR #3055, opened 2025-04-29
added Lenovo x3650 M5

dieck:


dieck:
I see that one check has failed, but the error message is inconclusive to me, I don't know how to fix that?

device-types/Lenovo/system-x3650-M5.yaml
```diff
@@ -0,0 +1,60 @@
...
+part_number: 8871
+u_height: 2
+is_full_depth: true
+comments: '[Lenovo System x3650 M5 product guide](https://lenovopress.com/tips1194.pdf)'
+power-ports:
+  - name: PSU 1
+    type: iec-60320-c14
+  - name: PSU 2
+    type: iec-60320-c14
+rear-ports:
```
harryajc, line 14: Rear ports are designed for patch panels not devices, standard USB ports on servers are not something that we are currently mapping in this repository

device-types/Lenovo/system-x3650-M5.yaml
```diff
@@ -0,0 +1,60 @@
+---
+manufacturer: Lenovo
+model: System x3650 M5
+slug: lenovo-system-x3650-m5
+part_number: 8871
+u_height: 2
+is_full_depth: true
+comments: '[Lenovo System x3650 M5 product guide](https://lenovopress.com/tips1194.pdf)'
+power-ports:
```
harryajc, line 9: As these are hot swappable PSUs they need to be modeled as `module-bays` and then add the require PSUs as separate modules if they don't already exist

device-types/Lenovo/system-x3650-M5.yaml
```diff
@@ -0,0 +1,60 @@
+---
+manufacturer: Lenovo
+model: System x3650 M5
+slug: lenovo-system-x3650-m5
+part_number: 8871
+u_height: 2
+is_full_depth: true
```
harryajc, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

harryajc (approved):

PR #3056, opened 2025-04-29
adding Brocade 6505 FC SAN Switch

dieck:


device-types/Brocade/SAN-Switch-6505.yml
```diff
@@ -0,0 +1,78 @@
...
+    type: 16gfc-sfpp
+console-ports:
+  - name: Console
+    label: Serial Console
+    type: rj-45
+rear-ports:
+  - name: USB
+    label: USB disk
+    type: usb-a
+power-ports:
```
harryajc, line 72: As these appear to be hot swappable PSUs they need to be modeled as `module-bays` with the required modules if they don't exist already

harryajc (changes requested):

harryajc (approved):

PR #3060, opened 2025-04-30
Create GS728TP.yml

Atsharim:


device-types/Netgear/GS728TP.yml
```diff
@@ -0,0 +1,119 @@
...
+    poe_mode: pse
+    poe_type: type2-ieee802.3at
+  - name: '8'
+    type: 1000base-t
+    poe_mode: pse
+    poe_type: type2-ieee802.3at
+  - name: '9'
+    type: 1000base-t
+    poe_mode: pse
+    poe_type: type2-ieee802.3at
```
harryajc, line 51: Looking at the spec sheet only the first 8 ports are POE+ the rest are POE af, unless you are trying to model the GS728TPP

harryajc (changes requested):

harryajc (approved):

PR #3071, opened 2025-05-01
Create PowerEdge-R660.yaml

anadir7:
Adding PowerEdge R660


device-types/Dell/PowerEdge-R660.yaml
harryajc, line 1: There is not enough info for the device, please add the following properties:
`weight`
`weight_unit`
`airflow`
Then add `module-bays` for the PSUs and PICe card slots

harryajc (changes requested):

device-types/Dell/PowerEdge-R660.yaml
anadir7: Added the following properties:
weight
weight_unit
airflow
comments
power-ports
module-bays

anadir7 (commented):
Added the following properties:
weight
weight_unit
airflow
comments
power-ports
module-bays

harryajc (approved):

PR #3082, opened 2025-05-04
adding Qnap TS-439pro II

naos0815:


device-types/QNAP/TS-439proII.yaml
```diff
@@ -0,0 +1,17 @@
+---
+manufacturer: QNAP
+model: TS-439 Pro II
+slug: qnap-ts-439-pro-ii
+part_number: TS-439proII
+u_height: 0
+is_full_depth: false
```
harryajc, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

harryajc (approved):

PR #3103, opened 2025-05-06
Add Cisco NCS6008 device type and compatible module types

segachan:
Add Cisco NCS6008 device type and compatible module types


elevation-images/Cisco/cisco-ncs-6008.front.jpeg
harryajc, line 1: Thank you for your contribution, This image should be cropped so that is only shows the front of the device without any white or empty space.

harryajc (changes requested):

elevation-images/Cisco/cisco-ncs-6008.front.jpeg
segachan: @harryajc I'v cropped it and ramained the front image.

harryajc: @segachan The image should only have the front of the device, with no white or empty spare otherwise it will look terrible in the netbox rack view

segachan: @harryajc How about this image?
image

ref.
https://www.cisco.com/c/ja_jp/products/collateral/routers/network-convergence-system-6000-series-routers/data_sheet_c78-728048.html

harryajc: Much better

PR #3113, opened 2025-05-08
WS-C3750-24FS-S

akiontke:
device WS-C3750-24FS-S added

### DIFF
--- a/device-types/Cisco/WS-C3750-24FS-S.yaml
+++ b/device-types/Cisco/WS-C3750-24FS-S.yaml
@@ -1,0 +1,66 @@
+---
+manufacturer: Cisco
+model: Catalyst 3750-24FS IP Base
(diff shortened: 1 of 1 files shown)


device-types/Cisco/WS-C3750-24FS-S.yaml
```diff
@@ -0,0 +1,66 @@
+---
+manufacturer: Cisco
+model: Catalyst 3750-24FS IP Base
+slug: cisco-ws-c3750-24fs-s
+part_number: WS-C3750-24FS-S
+is_full_depth: false
+u_height: 1
```
harryajc, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

PR #3114, opened 2025-05-08
Add Alcatel-Lucent OS6860-24

bec-proy:


device-types/Alcatel-Lucent/OS6860-24.yaml
```diff
@@ -0,0 +1,82 @@
+---
+manufacturer: Alcatel-Lucent
+model: OS6860-24
+slug: alcatel-lucent-os6860-24
+part_number: 903713-90
+u_height: 1
+is_full_depth: false
```
harryajc, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

device-types/Alcatel-Lucent/OS6860-24.yaml
```diff
@@ -0,0 +1,82 @@
...
+console-ports:
+  - name: Console
+    type: usb-micro-b
+power-ports:
```
harryajc, line 11: Are these hot swappable PSUs? If they are they should be modeled as `module-bays` with the PSUs being their own modules

device-types/Alcatel-Lucent/OS6860-24.yaml
```diff
@@ -0,0 +1,82 @@
...
+    type: 10gbase-x-sfpp
+  - name: 1/1/27
+    type: 10gbase-x-sfpp
+  - name: 1/1/28
+    type: 10gbase-x-sfpp
+  - name: 1/1/29
+    type: 40gbase-x-qsfpp
+  - name: 1/1/30
+    type: 40gbase-x-qsfpp
+  - name: Loopback0
```
harryajc, line 81: Virtual interfaces is not something we are modeling in this repo.

harryajc (changes requested):

device-types/Alcatel-Lucent/OS6860-24.yaml
bec-proy: Could you check if it is now correct both the module-bay and the module_type for the PSU?

harryajc (approved):

PR #3123, opened 2025-05-12
Add QNAP TS-h1277AXU-RP

mathieumd:


harryajc (approved):

device-types/QNAP/TS-h1277AXU-RP.yml
```diff
@@ -4,20 +4,12 @@ model: TS-h1277AXU-RP
...
 rear_image: true
-power-ports:
-  - name: PSU 1
-    type: iec-60320-c14
-    maximum_draw: 550
-    allocated_draw: 370
-  - name: PSU 2
-    type: iec-60320-c14
-    maximum_draw: 550
-    allocated_draw: 370
```
mathieumd: It doesn't look like PSU are modulable, as both are available by default: [*Power Supply Unit 	550W(x2), 100-240V*](https://www.qnap.com/en/product/ts-h1277axu-rp/specs/hardware)

harryajc: As the PSUs are Field replaceable units they should be modeled as module-bays

PR #3144, opened 2025-05-15
added HPE StoreOnce 5260, based on a DL385 Gen10 chassis

demik:
Added this backup Appliance.

It's based off a DL385 Gen10 chassis, but some PCIe slots are unuseable so not specified, and there is no onboard network ports as the network is on an OCP 3.0 slot.

It's usually shipped with some kind of RJ-45 OCP Ethernet module (either 4x1G BCM5719 or 2x10G BCM57416)


device-types/HPE/StoreOnce-5260.yaml
```diff
@@ -0,0 +1,28 @@
+---
+manufacturer: HPE
+model: StoreOnce 5260
+slug: hpe-storeonce-5260
+u_height: 2
+is_full_depth: true
+airflow: front-to-rear
```
harryajc, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

device-types/HPE/StoreOnce-5260.yaml
harryajc: I would go with the minimum weight of the chassis

harryajc (approved):

PR #3158, opened 2025-05-20
Added device type Zyxel ES-1024B

dienteperro:


dienteperro (commented):
Changed value to integer after check.

device-types/Zyxel/ES-1024B.yaml
```diff
@@ -0,0 +1,63 @@
+---
+manufacturer: Zyxel
+model: ES-1024B
+slug: zyxel-es1024b
+u_height: 1
+is_full_depth: false
+comments: '[Zyxel ES-1024B 24-port Unmanaged Rackmount Ethernet Switch](https://www.zyxelguard.com/datasheets/Switches/ES-1024B.pdf)'
+weight: 1.45
```
harryajc, line 8: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

harryajc (approved):

PR #3159, opened 2025-05-20
Create fortigate-7121f.yaml

visezion:


device-types/Fortinet/fortigate-7121f.yaml
```diff
@@ -0,0 +1,49 @@
...
+  - name: FPM Slot 9
+    position: '9'
+  - name: FPM Slot 10
+    position: '10'
+  - name: FPM Slot 11
+    position: '11'
+  - name: FPM Slot 12
+    position: '12'
+power-ports:
+  - name: PSU1
```
harryajc, line 38: As these PSUs are FRUs they should be modeled as `module-bays`

harryajc (approved):

PR #3198, opened 2025-05-26
Add QNAP ts 463 u rp

nativenoble:


device-types/QNAP/TS-463U-RP.yaml
```diff
@@ -0,0 +1,34 @@
+---
+manufacturer: QNAP
+model: TS-463U-RP
+slug: qnap-ts-463u-rp
+part_number: TS-463U-RP
+u_height: 1
+airflow: front-to-rear
+is_full_depth: false
+comments: '[TS-463U-RP | Hardware Specs | QNAP](https://www.qnap.com/de-de/product/ts-463u-rp/specs/hardware)'
+power-ports:
```
harryajc, line 10: PSUs that are FRUs, should be modeled as `module-bays`

PR #3207, opened 2025-05-31
Add Apple Mac Pro 2013 "Trash Can"

JunielKatarn:
See https://support.apple.com/en-us/112025

### DIFF
--- a/device-types/Apple/A1481.yaml
+++ b/device-types/Apple/A1481.yaml
@@ -1,0 +1,18 @@
+---
+manufacturer: Apple
+model: MacPro6,1
+slug: apple-macpro61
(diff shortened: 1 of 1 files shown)


device-types/Apple/A1481.yaml
```diff
@@ -0,0 +1,21 @@
+---
+manufacturer: Apple
+model: MacPro6,1
+slug: apple-macpro61
+part_number: A1481
+u_height: 0
```
harryajc, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

PR #3208, opened 2025-05-31
Add Apple Mac Pro 2019

JunielKatarn:
See
- https://support.apple.com/en-us/118461
- https://support.apple.com/en-us/111907

### DIFF
--- a/device-types/Apple/A1991.yaml
+++ b/device-types/Apple/A1991.yaml
@@ -5,6 +5,9 @@ slug: apple-macpro71
 part_number: A1991
 u_height: 0
 is_full_depth: false
+airflow: rear-to-front
+weight: 18
+weight_unit: kg
 comments: |
   See https://support.apple.com/en-us/118461
 power-ports:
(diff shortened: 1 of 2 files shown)


device-types/Apple/A1991.yaml
```diff
@@ -0,0 +1,20 @@
+---
+manufacturer: Apple
+model: MacPro7,1
+slug: apple-macpro71
+part_number: A1991
+u_height: 0
```
harryajc, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

device-types/Apple/A2304.yaml
```diff
@@ -0,0 +1,20 @@
+---
+manufacturer: Apple
+model: MacPro7,1
+slug: apple-rack-macpro71
+part_number: A2304
+u_height: 5
```
harryajc, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

PR #3211, opened 2025-06-04
Create FlashSystem_5300.yaml

Animus82:


device-types/IBM/FlashSystem-5300.yaml
```diff
@@ -0,0 +1,67 @@
...
+  - name: node2 10
+    type: 10gbase-x-sfpp
+  - name: node1 11
+    type: 10gbase-x-sfpp
+  - name: node2 11
+    type: 10gbase-x-sfpp
+console-ports:
+  - name: con0
+    type: usb-c
+power-ports:
```
harryajc: As theses PSUs are FRUs they should be modeled as `module-bays` with the required PSUs added as modules if they don't already exist

device-types/IBM/FlashSystem-5300.yaml
```diff
@@ -0,0 +1,67 @@
...
+manufacturer: IBM
+model: FlashSystem 5300
+slug: ibm-fs-5300
+part_number: 4662-7H2
+is_full_depth: true
+u_height: 1
+airflow: front-to-rear
+weight: 5.76
+weight_unit: kg
+interfaces:
```
harryajc: For Chassis with storage controllers the best way to model them is by modeling the controllers as `device-bays` then create the controller as a separate device. Also if any of the interfaces are part of expansion cards within that controller then they should be modeled as `module-bays` within the controller device.

device-types/IBM/FlashSystem-5300.yaml
Animus82: is changed

device-types/IBM/FlashSystem-5300.yaml
Animus82: is changed

PR #3223, opened 2025-06-06
Add Cisco SG220-26P2

bastianleicht:


device-types/Cisco/SG220-26P.yml
```diff
@@ -0,0 +1,123 @@
+---
+manufacturer: Cisco
+model: SG220-26
+part_number: SG220-26P
+slug: cisco-sg220-26p
+u_height: 1
+weight: 3.7
```
harryajc, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (approved):

PR #3231, opened 2025-06-07
Add support for AVM FRITZ!Box 7490

RbVs:
Introduces device specifications for FRITZ!Box 7490 including
interfaces, power port details, and airflow properties. Enhances
compatibility for device management and integration.


device-types/AVM/FRITZBox-7490.yaml
```diff
@@ -0,0 +1,43 @@
+---
+manufacturer: AVM
+model: FRITZ!Box 7490
+slug: avm-fritzbox-7490
+part_number: '20002678'
+u_height: 0
+is_full_depth: false
+airflow: passive
```
harryajc, line 8: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

PR #3233, opened 2025-06-11
Add device type Fortinet FG-401F

aopdal:
Add Fortinet device FG-401F


device-types/Fortinet/FG-401F.yaml
```diff
@@ -0,0 +1,90 @@
...
+  - name: port13
+    type: 1000base-t
+  - name: port14
+    type: 1000base-t
+  - name: port15
+    type: 1000base-t
+  - name: port16
+    type: 1000base-t
+  - name: port17
+    type: 1000base-x-sfp
```
harryajc, line 60: Looking at the spec sheet 8 of these sfp ports are SFP+, for SFP+ ports they should be `type: 10gbase-x-sfpp`

harryajc (changes requested):

aopdal:
I have changed interface name and type on 10Gig interfaces.

harryajc (approved):

PR #3238, opened 2025-06-12
Update CRS312-4C+8XG-RM

JunielKatarn:
Corrects network port types:

- 12 ports are 10G-BaseT RJ-45
- Only 4 ports are SFP+ \
  While port groups 1T-4T and 1F-4F are actually just 4 combo ports (either Ethernet or SFP), there **is** a total of 16 physical connections which should be reflected in the device type spec.
- Added missing PSU (this device type has two)

See https://mikrotik.com/product/crs312_4c_8xg_rm


device-types/MikroTik/CRS312-4C-Plus-8XG-RM.yaml
```diff
@@ -4,34 +4,70 @@ model: CRS312-4C+8XG-RM
...
+    label: MGMT/BOOT
+    mgmt_only: true
 console-ports:
-  - name: serial0
+  - name: serial1
     type: rj-45
 power-ports:
-  - name: PSU0
-    type: dc-terminal
+  - name: PSU1
```
harryajc, line 68: Are these hot swappable PSUs? If so they should be modeled as `module-bays`.

JunielKatarn: They are not swappable.
They are redundant internal PSUs to prevent a power-related single point of failure.

JunielKatarn: See https://www.servethehome.com/mikrotik-crs312-4c-8xg-rm-12-port-10gbe-switch-review/

image

PR #3242, opened 2025-06-12
Add Supermicro Motherboard X8STi

rccarmenaty:
This pull request introduces a new device definition for the Supermicro Motherboard X8STi.

The details for this definition have been sourced from the official product page: [Supermicro Motherboard X8STi](https://www.supermicro.com/products/archive/motherboard/x8sti)

A quick note regarding the console port: I initially tried to set the `type` to `vga`, but this doesn't appear to be a registered choice in the project's schema. As a temporary workaround, it can be set to `other`. If there's a more suitable existing type or if you'd prefer I submit a separate issue to discuss adding `vga` as a valid type, please let me know.

Thank you for your time and consideration.

### DIFF
--- a/device-types/Supermicro/MBD-X8STi.yaml
+++ b/device-types/Supermicro/MBD-X8STi.yaml
@@ -1,0 +1,52 @@
+---
+manufacturer: Supermicro
+model: Motherboard X8STi
+slug: supermicro-mbd-x8sti
+part_number: MBD-X8STi
+u_height: 0
+is_full_depth: true
+airflow: front-to-rear
+comments: Supermicro motherboard X8STi (https://www.supermicro.com/products/archive/motherboard/x8sti)
+subdevice_role: child
+module-bays:
+  - name: PCI-E 1
+    position: "1"
+    description: PCI-E 2.0 x16
+  - name: PCI-E 2
+    position: "2"
+    description: PCI-E 2.0 x16
+  - name: PCI-E 3
+    position: "3"
+    description: PCI-E x4 (in x8 slot)
+  - name: PCI 1
+    position: "4"
+    description: 32-bit 33MHz PCI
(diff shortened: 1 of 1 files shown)


device-types/Supermicro/MBD-X8STi.yaml
```diff
@@ -0,0 +1,40 @@
...
+    mgmt_only: false
+  - name: eth1
+    type: 1000base-t
+    mgmt_only: false
+power-ports:
+  - name: PSU1
+    type: iec-60320-c14
+device-bays: []
+module-bays: []
+inventory-items:
```
harryajc: Inventory items are being phased out so not something we would be looking to add

device-types/Supermicro/MBD-X8STi.yaml
```diff
@@ -0,0 +1,40 @@
+---
+manufacturer: Supermicro
+model: Motherboard X8STi
+slug: supermicro-mbd-x8sti
+part_number: MBD-X8STi
+u_height: 1
```
harryajc: As this is a motherboard it should have `subdevice_role: child` and as part of that it needs to be u_height: 0

device-types/Supermicro/MBD-X8STi.yaml
```diff
@@ -0,0 +1,40 @@
...
+comments: Supermicro X8STi motherboard - LGA1366, 2x1GbE, 6xDDR3, IPMI supported
+
+interfaces:
+  - name: eth0
+    type: 1000base-t
+    mgmt_only: false
+  - name: eth1
+    type: 1000base-t
+    mgmt_only: false
+power-ports:
```
harryajc: Power should be provided by the chassis

harryajc (changes requested):

PR #3026, opened 2025-04-18
Add rack-types

bastianleicht:
As described in https://github.com/netbox-community/devicetype-library/issues/2618

Implementation of the importer in: https://github.com/netbox-community/Device-Type-Library-Import/pull/170

PR #3244, opened 2025-06-13
feat - add 42U variant of HP 10000 G2 series rack

kquinsland:
This is predicated on https://github.com/netbox-community/devicetype-library/pull/3026.

Adds another Rack.


bastianleicht (commented):
Looks good!

bastianleicht (commented):
Could you please add the `desc_units` Property?

That was missing in the original PR. For your Rack it should be `desc_units: false`

harryajc (approved):

PR #3246, opened 2025-06-13
Fix HPE Aruba 8325 Bundle SKU, added HPE Aruba 8325 Power Supplies and Fans

robinreinhardt:
The existing 8325 models had the SKUs from the bundles (with fans and power supplies).
I have completed the bundle specification and added the specification for the standalone SKUs.
I also added the power supply and fan models for the 8325s.


harryajc:
Hi @robinreinhardt

Devices that have removal PSUs and Fans should have those modeled at Module-bays, even if they come as a bundle from the manufacturer. Having the bundles as separate devices to the standard chassis seems like a lot of duplication within this repo, I would recommend we keep it as the 1 new device that you have created with the modules and remove the bundle devices.

PR #3249, opened 2025-06-14
Firewalla-Gold-Plus

ultimatejedi:


ultimatejedi:
OK, I am 19 commits in, and I am not smart enough to get past the yamlfmt test.  I know it is something simple I am missing, but if anybody can provide any insight, it would be greatly appreciated.  I am new to git in general and the pre-scripts that run.  I have been working with yaml for several years, but this one is getting the best of me.

harryajc (approved):

ultimatejedi:
Thanks!  Is that all I was missing was the references to the images?  Sorry for the mislabel on the pics.

harryajc:
> Thanks! Is that all I was missing was the references to the images? Sorry for the mislabel on the pics.

Hi, Think the lint was failing because of the `'` on the weight_unit line, it can be a bit odd what causes it to fail some times.

PR #3251, opened 2025-06-15
Feat - Add the `NA3` variant of popular `ap8959` PDU

kquinsland:
US/nema-5-15 version of `ap8959`

### DIFF
--- a/device-types/APC/AP8959NA3.yaml
+++ b/device-types/APC/AP8959NA3.yaml
@@ -1,0 +1,117 @@
+# yaml-language-server: $schema=../../schema/devicetype.json
+---
+manufacturer: APC
+model: AP8959NA3
+slug: apc-ap8959na3
+part_number: AP8959NA3
+u_height: 0
+is_full_depth: false
+comments: |
+  Rack PDU 2G, Switched, ZeroU, 16A, 120V, (21) C13 & (3) C19
+console-ports:
+  - name: Serial
+    type: rj-12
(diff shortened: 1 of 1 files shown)


device-types/APC/AP8959NA3.yaml
```diff
@@ -0,0 +1,116 @@
+---
+manufacturer: APC
+model: AP8959NA3
+slug: apc-ap8959na3
+part_number: AP8959NA3
+u_height: 0
```
harryajc, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

device-types/APC/AP8959NA3.yaml
kquinsland01: 1. It's a long/tall/skinny PDU with no vents. All cooling is passive radiation but I guess you could make a weak argument that because it's usually installed at the back of the rack in a vertical orientation you get f2b airflow.

2. Good catch, I can add that. Quick google shows `13.3` kg seems to be the weight sans shipping materials.

kquinsland01: Resolved: https://github.com/netbox-community/devicetype-library/pull/3251/commits/08523b48cf16a0e45256bae129b592446b141e65

PR #3258, opened 2025-06-17
Create Tripp-Lite-BP72V15-2U.yaml

jemery27:


device-types/Eaton/Tripp-Lite-BP72V15-2U.yaml
```diff
@@ -0,0 +1,15 @@
+---
+manufacturer: Eaton
+model: Tripp Lite BP72V15-2U
+slug: eaton-tripp-lite-bp72v15-2u
+part_number: '037332124333'
+u_height: 2
+is_full_depth: true
+weight: 72.4
```
harryajc, line 8: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

PR #3261, opened 2025-06-20
Add/ckg2 rm

kquinsland:
rack mount adapter for cloud key gen2 devices


device-types/Ubiquiti/CKG2-RM.yaml
```diff
@@ -0,0 +1,27 @@
...
+is_full_depth: false
+weight: 600
+weight_unit: g
+airflow: passive
+front_image: true
+rear_image: true
+# This is always a parent device. When standalone, the cloud key is technically is not a parent or child.
+# When inserted, it should be modeled as a child of this parent device.
+# Unclear how best to model this for this repo; probably best to let user modify the cloud-key manually post import?
+subdevice_role: parent
```
harryajc: If this is a parent device, then it should have a `device-bay` for the child device

harryajc (changes requested):

device-types/Ubiquiti/CKG2-RM.yaml
kquinsland01: Resolved: https://github.com/netbox-community/devicetype-library/pull/3261/commits/11944aaa00ed05cb5b1749ea7fb50712968f07a8

harryajc: This should be a `device-bay` rather than a `module-bay` as the Cloud Key is a separate device

kquinsland01: Resolved: https://github.com/netbox-community/devicetype-library/pull/3261/commits/45d880e3c4d53c45fdce5eb5392a3ac94e3cb864

harryajc (approved):

PR #3266, opened 2025-06-24
Cable Matters variant of 1U keystone patch panel

kquinsland:
Cable Matters variant of the generic 1U 24port keystone

### DIFF
--- a/device-types/Cable Matters/180013.yaml
+++ b/device-types/Cable Matters/180013.yaml
@@ -1,0 +1,61 @@
+---
+manufacturer: Cable Matters
+model: 24 Port Keystone Patch Panel
+slug: cable-matters-24-port-keystone-patch-panel
+# Cable Matters appears to use sequential model numbers.
+# I am using a generic but appropriate model number here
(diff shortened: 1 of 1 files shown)


device-types/Cable Matters/180013.yaml
```diff
@@ -0,0 +1,61 @@
+---
+manufacturer: Cable Matters
+model: 24 Port Keystone Patch Panel
+slug: cable-matters-24-port-keystone-patch-panel
+# Cable Matters appears to use sequential model numbers.
+# I am using a generic but appropriate model number here
+# The part number matches their website
+# # https://www.cablematters.com/pc-386-162-rack-or-wall-mount-24-port-blank-patch-panel.aspx
+part_number: '180013'
+u_height: 1
```
harryajc, line 10: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

PR #3285, opened 2025-06-30
Update M4300-12X12F.yaml

jaydenthorup:
Add interfaces to this template and remove modules

### DIFF
--- a/device-types/Netgear/M4300-12X12F.yaml
+++ b/device-types/Netgear/M4300-12X12F.yaml
@@ -9,29 +9,29 @@ comments: 'Documentation: [https://www.netgear.it/support/product/m4300-12x12f.a
 subdevice_role: parent
 interfaces:
   - name: 1/0/1
-    type: 10gbase-t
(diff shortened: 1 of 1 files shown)


device-types/Netgear/M4300-12X12F.yaml
```diff
@@ -7,10 +7,56 @@ is_full_depth: false
...
+  - name: 1/0/9
+    type: 10gbase-t
+  - name: 1/0/10
+    type: 10gbase-t
+  - name: 1/0/11
+    type: 10gbase-t
+  - name: 1/0/12
+    type: 10gbase-t
+  - name: 1/0/13
+    type: 10gbase-t
```
harryajc, line 36: Looking at the spec sheet half of these ports are SFP+ therefore half of these ports should have the `type: 10gbase-x-sfpp`

jaydenthorup: Correcting these now, good catch.

harryajc (approved):

PR #3291, opened 2025-07-03
Added Brocade ICX6450-48P

frantathefranta:
Using Brocade ICX6430/ICX6450 spec sheet [from here](https://webresources.ruckuswireless.com/pdf/datasheets/ds-icx-6430-6450.pdf).


device-types/Brocade/ICX-6450-48P.yaml
```diff
@@ -0,0 +1,128 @@
...
+u_height: 1
+airflow: side-to-rear
+weight: 14.11
+weight_unit: lb
+interfaces:
+  - name: GigEthernetmgmt1
+    type: 1000base-t
+    mgmt_only: true
+  - name: GigabitEthernet1/1
+    type: 1000base-t
```
harryajc, line 16: Looking at the spec sheet these are POE+ ports therefore all POE+ ports need the following:
`poe_mode: pse`
`poe_type: type2-ieee802.3at`

harryajc (changes requested):

device-types/Brocade/ICX-6450-48P.yaml
frantathefranta: Thanks for noticing that, should now be added to all ports

harryajc (approved):

PR #3293, opened 2025-07-04
Added Juniper device elevation images.

David-Kahn-CA:
Added missing elevation images for the following devices:
- Juniper QFX5110-48s
- Juniper QFX5120-32c
- Juniper QFX5120-48y
- Juniper QFX5220-32cd
- Juniper SRX345
- Juniper SRX380
- Juniper SRX4600


David-Kahn-CA:
> All images should be cropped so that they don't have any white or empty space around the device. Also you need to add the following to the device yaml for these devices so that the images link to the required devices: `front_image: true` `rear_image: true`

Thank you for the feedback, images have been cropped down. 

Can you please clarify something for me before I begin adding the image information to the device yaml files?

For various models Juniper offers various airflow and PSU options in the form of AFI / AFO / AC / DC variants. The image files Juniper provides for these are the same regardless of the PSU/airflow configuration. (Orange for AFO, Blue for AFI.) Should the image files be named as the "base" device name then I add the requested `front_image` and `rear_image` information to the device yaml files where the image is applicable?

For example, the QFX-5110-48s image is named 'juniper-qfx5110-48s.front' and 'juniper-qfx5110-48s.rear'. This image set would apply to all of the following devices in the repo:

* QFX5110-48S-AFI
* QFX5110-48S-AFO
* QFX5110-48S-DC-AFI
* QFX5110-48S-DC-AFO

Alternatively, do the device images in this case need to be renamed to the full device name and then duplicated for each variant they apply to?

IE:
* QFX5110-48S-AFI -- QFX5110-48S-AFI.front and QFX5110-48S-AFI.rear
* QFX5110-48S-AFO -- QFX5110-48S-AFO.front and QFX5110-48S-AFO.rear
* QFX5110-48S-DC-AFI -- QFX5110-48S-DC-AFI.front and QFX5110-48S-DC-AFI.rear
* QFX5110-48S-DC-AFO -- QFX5110-48S-DC-AFO.front and QFX5110-48S-DC-AFO.rear

EDIT: I reviewed the readme, the images need to match the slug value for each device. I'll get them copied.

PR #3298, opened 2025-07-08
Create Netgate

gcassista:
Add netgate Category to the elevation image


harryajc:
This is not required, the folder will automatically be generated once a file has been added to it

PR #3315, opened 2025-07-18
Add Swich Linksys LGS528

vsc55:
Information and images of the Linksys LGS528 switch


device-types/Linksys/LGS528.yml
```diff
@@ -0,0 +1,110 @@
...
+  - name: gi25
+    type: 1000base-t
+  - name: gi26
+    type: 1000base-t
+  - name: gi27
+    type: 1000base-t
+  - name: gi28
+    type: 1000base-t
+  - name: Loopback
+    type: virtual
```
harryajc: Virtual port are not something that is currently being modeled in this repository

device-types/Linksys/LGS528.yml
```diff
@@ -0,0 +1,110 @@
...
+    type: lag
+    enabled: false
+  - name: Po7
+    type: lag
+    enabled: false
+  - name: Po8
+    type: lag
+    enabled: false
+module-bays:
+  - name: sfp27
```
harryajc: SFP ports should be modeled as interfaces with the correct type

vsc55: The gi27 and spf27 ports are the same logical interface, as are the gi28 and sfp28 ports. You can use either the gi27 (ethernet) port or the SFP27 port, but not both at the same time. That's why the SFP port is defined as "module-bays," since you can add different types of SFP devices to that slot.
If this isn't correct, how should I define it? Do you have an example?

device-types/Linksys/LGS528.yml
vsc55: perfect, remove the virtual port

device-types/Linksys/LGS528.yml
harryajc: I think the best way i have seen this done is like: https://github.com/netbox-community/devicetype-library/blob/29fcb13d2f829d7354b38b4af7d0fe416aba5518/device-types/D-Link/DWS-3160-24TC.yaml#L69
Where you have the Port F and T versions, other option that quite a few have done is just put the port down as a SFP port and then add a note that it can be a rj45 port in the description

vsc55: Thanks @harryajc for the info.
Let's see how it goes now with the latest update.

harryajc (approved):

PR #3321, opened 2025-07-21
Update Cisco Nexus 7700 device and module types

m-hau:
Improve and complete the three existing Cisco Nexus 7700 router device types and eight module types, while also adding six new module types. The updates are for the most part a human-readable description, comment with a link to the data sheet, weight and airflow.

The name of a module bay now corresponds to the slot names that the `show inventory` cli command shows/uses, and the label matches the actual physical label.

One note-worthy change is to the way the management ethernet interface and serial console is modelled.
From a OS perspective, there is only a single configurable `mgmt0` ethernet interface, which corresponds to the physical management ethernet interface of the "active" supervisor module. The management ethernet interface of the "standby" supervisor module (if any) is basically unused, until the supervisor module becomes the "active" one of course. The allow proper modelling and documentation of all this, the chassis now comes with a virtual `mgmt0` ethernet interface, which can be used for the logical interface configuration (IPs and such). Each supervisor module now comes with a `mgmtX` interface, which can be used to properly model the physical connections (cables).
It is a bit different for the serial console ports. The chassis itself does not have any serial console ports, only the supervisor modules do. There is no switch-over like with the management ethernet interface, the OS can see each of these independently. `con0` is always the serial console port of the first supervisor module, and `con1` of the second (if installed). So these console ports are now part of the supervisor modules, allowing proper configuration and cabling. Due to the limitation of the (auto-)naming possibilities, they will be called `con5` and `con6` and need to be manually corrected (if needed).


harryajc:
Hi @m-hau 

One thing by renumbering the positions numbers on the PSUs they when modules with the {{module}} variable won't get PS1 they will get PS30 or whichever, might become a bit confusing.

m-hau:
Good point, I changed all positions so they "reset" per slot-type, since it would also effect fabric and fan modules. In the specific case of PSUs they still get renumbered from the range `0 ~ n-1` to `1 ~ n`, because that's the way the Cisco documentation and CLI numbers them. But because all slot names are now similar, the validation complains about `duplicated positions within module-bays`. :sweat: So this may not be possible after all.

For completeness I also adjusted the 2- and 18-slot device variants. But I don't have access to actual hardware of these and can't verify the slot names and labels, so it's best-effort based on documentation and may not be absolutely accurate.

harryajc:
Hi @m-hau 
Very strange, we have many instances within this repo of devices with different module bays using the same position numbers so not sure why it doesn't like it. 

Done some testing looks like it doesn't like having the module names as different numbers to the positions I did a test on 02 with the below and passed successfully:
```
module-bays:
  - name: Slot 1
    position: '1'
    label: '1'
    description: Line Card
  - name: Slot 2
    position: '2'
    label: '2'
    description: Supervisor Module
  - name: PS1
    position: '1'
    label: PS 1
    description: Power Supply 1
  - name: PS2
    position: '2'
    label: PS 2
    description: Power Supply 2
  - name: Fan1
    position: '1'
    label: Fan
    description: Fan
```

m-hau:
The reason is in the [`validate_component_names()`](https://github.com/netbox-community/devicetype-library/blob/d18d144e/tests/device_types.py#L180) method, which gets called via [`validate_components()`](https://github.com/netbox-community/devicetype-library/blob/d18d144e/tests/device_types.py#L271) from the [`test_definitions`](https://github.com/netbox-community/devicetype-library/blob/d18d144e/tests/definitions_test.py#L196) test. The same `position` is allowed if the `name`s don't share a common prefix.

I am going to swap name and description, that should pass the validation.

harryajc:
> The reason is in the [`validate_component_names()`](https://github.com/netbox-community/devicetype-library/blob/d18d144e/tests/device_types.py#L180) method, which gets called via [`validate_components()`](https://github.com/netbox-community/devicetype-library/blob/d18d144e/tests/device_types.py#L271) from the [`test_definitions`](https://github.com/netbox-community/devicetype-library/blob/d18d144e/tests/definitions_test.py#L196) test. The same `position` is allowed if the `name`s don't share a common prefix.
> 
> I am going to swap name and description, that should pass the validation.

Yeah i think it is to make sure that the positions match up with the slot naming, although I think the error needs some adjusting so it makes more sense

PR #3341, opened 2025-07-24
Create aruba-cx-6000-48g-4sfp.yaml

visezion:


harryajc:
The aruba CX 6000 model already exists in the repo

PR #3342, opened 2025-07-24
Create CX-8400.yaml

visezion:


harryajc:
The Aruba CX 8400 already exists in this repo

PR #3344, opened 2025-07-26
Dell MX9116n and MX7116n

connect-cloud:


device-types/Dell/MX7116n-Fabric-Expander-Module.yaml
harryajc, line 1: As this is a network card this should probably be modeled as a module

device-types/Dell/MX9116n-Fabric-Switching-Engine.yaml
harryajc, line 1: As this is a network card this should probably be modeled as a module

harryajc (changes requested):

device-types/Dell/MX9116n-Fabric-Switching-Engine.yaml
connect-cloud: Hi Harry, the MX9116n is a switch which can be put in a Dell MX7000 chassis. I moddeled it as a Child device because it's a device with it's own IP, management and configuration. Because we also have the HPE Synergy systemes I moddeled them the same: e.g. https://github.com/netbox-community/devicetype-library/blob/master/device-types/HPE/Virtual-Connect-SE-40Gb-F8-Module-for-Synergy.yaml
Please advise

harryajc: HI @connect-cloud 
Thank you for the information. Okay yes this should stay as a device type if it has it own management. For Devices we are asking for the following fields to be populated if possible:
`weight` and associated `weight unit`
`airflow`

harryajc:
No repose in over 9 months

PR #3346, opened 2025-07-26
Adds new Huawei Device Type: S6730-H48Y6C-V2

AlmeidaCaio:
Adds a new Device Type:
* **Manufacturer:** Huawei
* **Model:** S6730-H48Y6C-V2
* **Device Specifications (Reference):** https://support.huawei.com/enterprise/en/doc/EDOC1100214356/7a23a1e6/s6730-h48y6c-v2


device-types/Huawei/S6730-H48Y6C-V2.yaml
```diff
@@ -0,0 +1,150 @@
+---
+manufacturer: Huawei
+model: S6730-H48Y6C-V2
+slug: huawei-s6730-h48y6c-v2
+part_number: 02354NGU
+u_height: 1
+is_full_depth: true
```
harryajc, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

device-types/Huawei/S6730-H48Y6C-V2.yaml
AlmeidaCaio: I just pushed a commit with fields you asked for; however, I'm still trying to configure the pre-commit, it keeps failing on the .yaml format (yamlfmt). Maybe it's because I'm using Windows?

Anyways, thanks for your reply and review! I'm glad I could contribute.

PR #3358, opened 2025-07-31
Add Supermicro SYS-E300-12D-4CN6P

robobenklein:
Using one of these, but can't recall which 2 GbE ports are the I350-AM2 vs the I210-IT network chipset, if someone knows how to check on freebsd let me know or just add port descriptions.

### DIFF
--- a/device-types/Supermicro/SYS-E300-12D-4CN6P.yml
+++ b/device-types/Supermicro/SYS-E300-12D-4CN6P.yml
@@ -1,0 +1,31 @@
+---
+manufacturer: Supermicro
+model: IoT SuperServer SYS-E300-12D-4CN6P
+slug: supermicro-superserver-sys-e300-12d-4cn6p
+u_height: 1
+is_full_depth: false
+airflow: front-to-rear
+part_number: SYS-E300-12D-4CN6P
+weight: 1.6
+weight_unit: kg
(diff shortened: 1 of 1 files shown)


device-types/Supermicro/SYS-E300-12D-4CN6P.yml
```diff
@@ -0,0 +1,30 @@
+---
+manufacturer: Supermicro
+model: IoT SuperServer SYS-E300-12D-4CN6P
+slug: supermicro-superserver-sys-e300-12d-4cn6p
+u_height: 1
+is_full_depth: false
+part_number: SYS-E300-12D-4CN6P
+weight: 1.6
+weight_unit: kg
```
harryajc, line 10: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

device-types/Supermicro/SYS-E300-12D-4CN6P.yml
robobenklein: Added based on fan orientation, although in practice the default two fans behave more like mixed.

harryajc (approved):

PR #3364, opened 2025-08-01
Add "airflow" and "profile" to moduletype test

iriseden:
This commit add "airflow" and "profile" to moduletype tests to allow using those objects.

- airflow : Usefull for fan and power supply
- profile : Introduced in netbox v4.3 (https://netboxlabs.com/docs/netbox/models/dcim/moduletypeprofile/)

### DIFF
--- a/schema/moduletype.json
+++ b/schema/moduletype.json
@@ -74,6 +74,9 @@
         },
         "comments": {
             "type": "string"
+        },
+        "profile": {
+            "type": "string"
         }
     },
     "required": ["manufacturer", "model"],
--- a/schema/moduletype.json
+++ b/schema/moduletype.json
@@ -12,6 +12,9 @@
         "part_number": {
             "type": "string"
         },
+        "airflow": {
+            "$ref": "urn:devicetype-library:generated-schema#/definitions/airflow"
+        },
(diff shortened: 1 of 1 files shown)


schema/moduletype.json
```diff
@@ -71,6 +74,9 @@
         },
         "comments": {
             "type": "string"
+        },
+        "profile": {
```
harryajc, line 84: I can add the airflow, however the profile option will require some more work

iriseden: Hello, what kind of fix would you like? The problem with profile is that it is not standard and can be defined by the user.

harryajc: First we would only support the default netbox profiles in this repo, secondly I was having a look the other day, but still not quite sure how the profile attributes are imported/or if that is even possible at this stage, I know in 4.3.4 they added support for importing the module profiles themselves, but currently don't see any information on importing the attributes

iriseden: Okay, I have no idea if we can import the attributes, I'll test it. Regarding the default profiles, I'll correct the commit.

harryajc:
Module Profiles have been added through a different PR

PR #3369, opened 2025-08-04
Treetech sdg

emanuel-bs:


device-types/Treetech/SDG.yaml
```diff
@@ -0,0 +1,15 @@
+---
+manufacturer: Treetech
+model: SDG
+slug: treetech-sdg
+u_height: 1
+is_full_depth: false
+is_powered: true
+airflow: passive
```
harryajc, line 8: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

device-types/Treetech/SDG.yaml
```diff
@@ -0,0 +1,15 @@
...
+comments: Fully aggregate information from any equipment compatible with Modbus and/or DNP3 protocols and redistribute it in a highly customizable way in
+  these protocols and even in those of the IEC 61850 standard.
+power-ports:
```
harryajc, line 11: You have created some modules as part of this change, are they to go into this device therefore you need to create the associated `module-bays`

harryajc (changes requested):

harryajc:
No response in over 9 months

PR #3389, opened 2025-08-07
Improve ZPE Service Routers

m-hau:
Improve the existing types of ZPE's Service Router product line by updating the existing types, and adding a few new devices and modules.

The used part numbers are the ones reported by the devices themself, which only identify the "base chassis" and are not listed in the data sheet. The data sheet SKUs are more akin to pre-configured bundles with different memory-/storage-/expansion-configurations. This is not so prevalent with the NSR because it is a modular chassis, and more noticeable with the GSR and BSR. The actual slots of the device are modelled as module-bays, and can be populated using the appropriate modules to document the actual configuration.
The same is true for some of NSRs modules (namely storage and M.2), where the data sheet SKUs are pre-assembled bundles but the modules themselves report a generic part number. That is the reason these module types have again module-bays of their own, into which you can insert the actual harddisks/M.2 modules. Note that importing module-bays of modules was only added in v4.2.9 (see netbox-community/netbox#19311). While the actual M.2 cards do have their own SKUs, the hard disks don't, so it would be better for the storage module to use the new "Hard disk" module type profile; but that is not available yet (see #3364).
While the RAM is technically also a FRU, I opted to not model the RAM slots as module-bays since they are not intended to be changed, and the few people that actually do that and want to document it that detail degree can create the module-bays themselves using the comments. The same goes for the integrated storage in NSR.

For the most part the only real difference between the "base chassis" is the cpu, which is why as model I used the full product name and the cpu-core-count (and PSU type for the non-Lite NSR). I did not add all of the NSR chassis power variants, namely the SAC/POE/DC variants are still missing in the library, and need to be added later if required. Apart from the power supply they should be identical to the DAC variant.

The device definitions use the `bridge` attribute on the interfaces to model which ports are connected to the integrated switch chip. Note that the import into NetBox does currently not support this, but hopefully soon (see netbox-community/netbox#20042).

The (non-Lite) NSR and GSR have a single switch chip inside, but two backplane interfaces to it. To properly model this the `backplane0` can be considered the main switch interface, and `backplane1` is bridged by `backplane0`. Any ports that are added through modules will need to be manually assigned post-creation to the backplane0 bridge/interface.
By default the NSR comes with rear-exhaust fans, which is why the airflow is set to `side-to-rear`. When using a rear-to-side configuration using different fans, this needs to be set manually; and the fans can be documented using the module-bays.

I wasn't sure what to do with the digital IO ports of the GSR. In the end I modelled them as console server ports, since that was the closest fit, and allows you to cable them if wanted.

### DIFF
--- a/device-types/ZPE/BSR-T8-BASE.yaml
+++ b/device-types/ZPE/BSR-T8-BASE.yaml
@@ -62,11 +62,6 @@ power-ports:
     label: PWR IN
     type: dc-terminal
 interfaces:
-  - name: lo
-    type: virtual
-  - name: loopback0
-    type: virtual
-
   - name: eth0
     label: ETH 0
     type: 1000base-t
--- a/device-types/ZPE/GSR-T8-BASE.yaml
+++ b/device-types/ZPE/GSR-T8-BASE.yaml
@@ -81,11 +81,6 @@ power-ports:
     label: V2
     type: dc-terminal
 interfaces:
-  - name: lo
-    type: virtual
-  - name: loopback0
-    type: virtual
-
   - name: eth0
     label: ETH 0
     type: 1000base-t
--- a/device-types/ZPE/NSR-BASE-DAC.yaml
+++ b/device-types/ZPE/NSR-BASE-DAC.yaml
@@ -38,11 +38,6 @@ power-ports:
     type: iec-60320-c14
     description: Rear Center
 interfaces:
-  - name: lo
-    type: virtual
-  - name: loopback0
-    type: virtual
-
   - name: eth0
     label: ETH 0
     type: 1000base-t
--- a/device-types/ZPE/NSR-LITE-DAC.yaml
+++ b/device-types/ZPE/NSR-LITE-DAC.yaml
@@ -38,11 +38,6 @@ power-ports:
     type: iec-60320-c14
     description: Rear Center
 interfaces:
-  - name: lo
-    type: virtual
-  - name: loopback0
-    type: virtual
-
   - name: eth0
     label: ETH 0
     type: 1000base-t
--- a/device-types/ZPE/NSR-TOP1-DAC.yaml
+++ b/device-types/ZPE/NSR-TOP1-DAC.yaml
@@ -38,11 +38,6 @@ power-ports:
     type: iec-60320-c14
     description: Rear Center
 interfaces:
-  - name: lo
-    type: virtual
-  - name: loopback0
-    type: virtual
-
   - name: eth0
     label: ETH 0
     type: 1000base-t


device-types/ZPE/BSR-T8-BASE.yaml
```diff
@@ -0,0 +1,116 @@
...
+airflow: passive
+weight: 1.2
+weight_unit: kg
+description: 2-Core Intel CPU, 8x RJ45 Serial rolled, 5x GbE, Single AC Power Supply
+comments: '[ZPE Nodegrid Bold SR Data Sheet](https://go.zpesystems.com/rs/004-BTR-463/images/Datasheet-Nodegrid-Bold-SR.pdf)'
+console-ports:
+  - name: Console
+    label: CONSOLE
+    type: rj-45
+  - name: Monitor
```
harryajc: VGA ports are not something currently being modeled in this repo

device-types/ZPE/BSR-T8-BASE.yaml
```diff
@@ -0,0 +1,116 @@
...
+    type: usb-a
+    description: USB 3.0
+  - name: usbS4
+    label: '4'
+    type: usb-a
+    description: USB 3.0
+power-ports:
+  - name: PS
+    label: PWR IN
+    type: other
```
harryajc: If this a dc input the best type is dc-terminal

device-types/ZPE/BSR-T8-BASE.yaml
```diff
@@ -0,0 +1,116 @@
...
+interfaces:
+  - name: lo
+    type: virtual
```
harryajc: We don't usually include virtual ports in this repo

device-types/ZPE/GSR-T8-BASE.yaml
```diff
@@ -1,78 +1,168 @@
...
+airflow: passive
+weight: 0.9
+weight_unit: kg
+description: 4-Core Intel CPU, 8x RJ45 Serial rolled, 2x SFP+, 5x GbE, 4x PoE+, Dual DC input (wide range DC)
+comments: '[ZPE Nodegrid Gate SR Data Sheet](https://go.zpesystems.com/rs/004-BTR-463/images/Datasheet-Nodegrid-Gate-SR.pdf)'
 console-ports:
   - name: Console
+    label: CONSOLE
     type: rj-45
+  - name: Monitor
```
harryajc: Monitor outputs are not something we are currently modeling in this repo

harryajc (changes requested):
If you can make these changes on all devices that it applies

PR #3391, opened 2025-08-08
Enforce component properties

m-hau:
Tighten the validation, so that components (like device/module-types already do) can only have a defined set of properties. This helps to detect typos and other mistakes.

Simply enabling this caused a lot of validation errors (~700), of which most (~500) were fixed by updating/improving the json schema. The remaining ~200 were true positive failures, and properly fixed.

One notable change to the validation is the renaming of the custom power source marker from `poe` to `_is_power_source` to better reflect its intention.

The `generated_schema.json` file was manually edited, because the upstream version is missing an enumeration. I already created an upstream issue to fix that (see netbox-community/netbox#20056).

The individual commits contain more detailed information.


schema/components.json
```diff
@@ -18,14 +17,18 @@
         "type": {
           "$ref": "urn:devicetype-library:generated-schema#/definitions/console-port/properties/type"
         },
-        "poe": {
-          "type": "boolean"
+        "description": {
+          "type": "string"
+        },
+        "_is_power_source": {
```
harryajc, line 23: is the property _is_power_source used in netbox? as this can only be changed if that is the property netbox is expecting.

harryajc (changes requested):

schema/components.json
harryajc: @m-hau Apologies just realised that this is for validation and not something that is actually imported into netbox

PR #3392, opened 2025-08-08
Proposal: Update Extreme Networks Switch device-types

0lini:
This PR contains an exemplary device type for an Extreme Networks switch. 
The changes I propose are renaming interfaces to conform with the OS naming scheme, and labels that follow the physical markings.
This also includes replacing SFP interfaces with module bays.

### DIFF
--- a/device-types/Extreme Networks/7520-48Y-8C-FabricEngine.yaml
+++ b/device-types/Extreme Networks/7520-48Y-8C-FabricEngine.yaml
@@ -13,6 +13,7 @@ console-ports:
     type: rj-45
 interfaces:
   - name: MGMT
+    type: 1000base-t
     mgmt_only: true
 module-bays:
   - name: PS#1


device-types/Extreme Networks/5320-48P-8XE.yaml
harryajc, line 1: Happy with renaming of interface names, as they should match whatever the switch calls them, however currently we are not modeling SFP ports as module bays, they are modeled as the appropriate interface type. This has been the standard for every device in this repository, for this to change it would have to be the recommended approach by netbox

0lini: Hi, this is the case it seems. https://netboxlabs.com/blog/sfp-modeling-modules-over-inventory-items/

aaronaxvig:
Hello, a while back I added a lot of the Extreme devices in here.  Happy to see someone else doing the same.

As I imagine you know, some of these Extreme switches are odd in that they can run either EXOS or FabricEngine.  These are two completely different OSs.  Hopefully this information is helpful to other readers who may not have hands-on experience.

I personally don't use FabricEngine so am not super familiar with it, but apparently the interfaces are named 1/1, 1/2, and 1/3 in it.  In EXOS they are named 1, 2, and 3.  Or if the switch is in stacking mode they would be 1:1, 1:2, and 1:3.  On the switch they are physically printed as 1, 2, and 3.

The guidance from documentation is that the interface name should be: "The name of the interface, as reported by the device's operating system."

Because of this guidance, I think there is a reasonable case for keeping the existing precedent of different template definitions for EXOS and FabricEngine.

Your slugs will have to be modified to something like `extreme-networks-5320-48p-8xe-fabricengine` as I believe those have to be unique within this entire git repository.

PR #3393, opened 2025-08-08
Add neu folder for module type profiles with an exemplary SFP profile

0lini:
This PR includes a new folder, that will contain future module type profiles. Also it contains a profile for SFP transceiver based on one of the latest [blog](https://netboxlabs.com/blog/sfp-modeling-modules-over-inventory-items/) posts.

### DIFF
--- a/module-type-profiles/sfp-transceiver.yaml
+++ b/module-type-profiles/sfp-transceiver.yaml
@@ -1,0 +1,29 @@
+---
+name: SFP Transceiver
+schema: {
+    "properties": {
+        "connector_type": {
+            "description": "Physical connector type",
+            "enum": [
+                "LC",
+                "SC",
+                "MPO",
+                "RJ45"
+            ],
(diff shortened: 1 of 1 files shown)


harryajc:
If this is the direction that netbox want to go I would imagine that they would integrate this module profile into netbox, don't think there are really any other module profiles that are missing therefore I don't think this is something that we need to add into this repo

walday:
there are a lot of the same sfp modules taking into account different connectors, speeds and wavelengths.. and sometimes it is important to understand the manufacturer.. I don’t think that netbox will fill all this in themselves.. here, a library is most likely needed..

0lini:
I created this issue, this will give us an idea what they think. https://github.com/netbox-community/devicetype-library/pull/3393

0lini:
https://github.com/netbox-community/netbox/issues/20070#issuecomment-3189271254

PR #3394, opened 2025-08-08
Add Fortinet FortiGate 90E

mkolus:


device-types/Fortinet/FG-90E.yaml
```diff
@@ -0,0 +1,53 @@
+---
+manufacturer: Fortinet
+model: FortiGate 90E
+slug: fortinet-fg-90e
+part_number: FG-90E
+u_height: 1
+is_full_depth: false
+weight: 1.31
+weight_unit: kg
```
harryajc, line 10: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

device-types/Fortinet/FG-90E.yaml
mkolus: There it is, I have one at home so it was easy to figure out 😀
Btw, this is my very first pull request, thanks for your patience.

harryajc (approved):

PR #3403, opened 2025-08-11
Rename custom `is_powered` property

m-hau:
~~Small~~ follow-up to #3391.

~~Fix the `poe` to `_is_power_source` rename for ports in the README, forgot that in the original PR.~~

Rename the device-type-level custom property `is_powered` to `_is_powered`. The leading underscore makes it obvious that it's a custom property, and is then consistent with the ports' `_is_power_source`. Also remove it from device types if it's true (the default).

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,9 +80,9 @@ The following fields may **optionally** be declared:
 - `comments`: A string field which allows for comments to be added to the device. (**Default: None**)
   - Type: String
   - :test_tube: Example: `comments: This is a comment that will appear on all NetBox devices of this type`
-- `is_powered`: A boolean which indicates whether the device type does not take power. This is mainly used as a workaround for validation testing on non powered devices (i.e. rackmount kits or patch pannels.) (**Default: True**)
+- `_is_powered`: Indicates that the device type requires power to operate. Only used internally for power validation testing on non-powered devices (i.e. rackmount kits or patch pannels.) (**Default: True**)
   - Type: Boolean
-  - :test_tube: Example: `is_powered: false`
+  - :test_tube: Example: `_is_powered: false`
 - `weight`: A number representing the numeric weight value. Must be a multiple of 0.01 (2 decimal places). (**Default: None**)
   - Type: Number
   - Value: must be a multiple of 0.01
--- a/device-types/ADC/7033-Fibre-Panel.yaml
+++ b/device-types/ADC/7033-Fibre-Panel.yaml
@@ -9,7 +9,7 @@ is_full_depth: false
 airflow: passive
 weight: 1.14
 weight_unit: kg
-is_powered: false
+_is_powered: false
 front_image: true
 module-bays:
   - name: '1'
--- a/device-types/ADVA/f7-48csm-1hu-19600-19130.yaml
+++ b/device-types/ADVA/f7-48csm-1hu-19600-19130.yaml
@@ -5,7 +5,7 @@ slug: adva-f7-48csm-1hu-19600-19130
 part_number: 1078708781-01
 u_height: 1
 is_full_depth: false
-is_powered: false
+_is_powered: false
 airflow: passive
 front-ports:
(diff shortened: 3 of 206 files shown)


harryajc:
I think changing the property is probably not a good idea given it is used a lot within this repo

m-hau:
I believe it to be very beneficial, because a lot of the time people won't read the docs but just copy&paste some other type, and adjust it as needed, not realising that some fields are not imported into NetBox, carrying them over in a cargo-cult-like manner. Distinguishing fields from/for NetBox and repo-internal-only by using the underscore-prefix should make it obvious to anybody at a glance that later are different from all the others and thus probably "special" in some way. And since it shouldn't be used outside of the repo, renaming it should be pretty safe.

But that's just my opinion. If you want it, I can remove the rename and only leave the doc fixes.

harryajc:
HI @m-hau 
Okay, I think the best option is, if you can create a separate PR with the doc changes then I can get that approved and I'll get another maintainer to review this PR and see what they think of making this property change

m-hau:
Done. Moved the doc-update to #3404, and left the property-rename here.

harryajc:
@danner26 @DanSheps @ryanmerolle Thoughts on this?

DanSheps:
I need to review the ramifications of this.  This is the first time I am seeing this and this is not a standard attribute

harryajc:
> I need to review the ramifications of this. This is the first time I am seeing this and this is not a standard attribute

This is a attribute specific to this repo for devices that aren't powered so that the power validation scripts pass successfully.

DanSheps:
I am more wondering if this is even needed or if there is a way to work this so that it doesn't check for that.

Reason being is it is possible to have a chassis without PSU's or interfaces so it would be an "unpowered" device however we handle this fine.

I think the better way to do this is;

* Check for interfaces and console server ports, if there are no interfaces, console ports, console server ports, power outlets, or device bays we assume it is "passive device"
* If the device is passive, simply pass the test
* If the device is not passive, continue to validate

m-hau:
So instead of explicitly defining a device type as non-/powered and based on that validating the allowed/required components, the powered status would be inferred from the defined components and based on that validate the components. Sounds reasonable, but has the disadvantage that there is no "escape hatch" for corner cases, where you would still want to manually override/set it.
But you could also combine both: By default infer it from the components, but still allow it to be set manually in the definition. This would drastically reduce the number of manual-sets because the default would already do "the right thing" in most cases.

I did a quick-and-dirty implementation of this (applies to `master`), and after tweaking the checked-for components a bit (device-bay is allowed for non-powered, but not module-bay) all device-types still pass validation. Feel free to play around with it yourself.
```diff
diff --git a/tests/device_types.py b/tests/device_types.py
index 24919397..95c9d139 100644
--- a/tests/device_types.py
+++ b/tests/device_types.py
@@ -75,6 +75,11 @@ class DeviceType:
     def validate_power(self):
         CUSTOM_POWER_SOURCE_PROPERTY = '_is_power_source'
 
+        # ignore the power status from the definition, instead infer it from the components
+        self.definition['is_powered'] = False
+        for property_name in ('console-ports', 'console-server-ports', 'power-ports', 'power-outlets', 'interfaces', 'module-bays'):
+            self.definition['is_powered'] |= bool(self.definition.get(property_name, False))
+
         # Check if power-ports exists
         if self.definition.get('power-ports', False):
             # Verify that is_powered is not set to False. If so, there should not be any power-ports defined
```

m-hau:
It's been a while. May I re-raise the question how to continue with this?

harryajc:
I Think given how widely this is used this should remain how it is

m-hau:
I still believe it has merit to properly distinguish NetBox properties from DeviceTypeLibrary-only properties. And closing this leaves a bit of an inconsistent state, because now you have `_is_power_source` and `is_powered`.

PR #3428, opened 2025-08-21
Create Ubiquiti R5AC-Lite

GoldenCryer101:
Create entry for Ubiquiti R5AC-Lite device type under Ubiquiti


GoldenCryer101:
Fixed the slug format as per the test failure

harryajc (approved):

PR #3439, opened 2025-08-23
Adds a new CheckPoint Device Type: Quantum 16000 Security Gateway

AlmeidaCaio:
Adds a new Device Type:
* **Manufacturer:** Check Point
* **Model:** Quantum 16000 Security Gateway
* **Device Specifications (Reference):** https://www.checkpoint.com/downloads/products/16000-security-gateway-datasheet.pdf

### DIFF
--- a/device-types/Check Point/Quantum-16000-Security-Gateway.yaml
+++ b/device-types/Check Point/Quantum-16000-Security-Gateway.yaml
@@ -38,34 +38,34 @@ interfaces:
     label: SYNC
     type: 1000base-t
   - name: eth1-01
-    label: '40G 1'
+    label: 40G 1
     type: 40gbase-x-qsfpp
   - name: eth1-02
-    label: '40G 2'
+    label: 40G 2
     type: 40gbase-x-qsfpp
   - name: eth2-01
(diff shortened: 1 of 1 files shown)


AlmeidaCaio:
I don't understand the format error on pipeline's validation, could any of the reviewers help me? Thanks in advance

device-types/Check Point/Quantum-16000-Security-Gateway.yaml
```diff
@@ -0,0 +1,93 @@
...
+    type: 1000base-t
+    mgmt_only: true
+  - name: lom
+    label: LOM
+    type: 1000base-t
+    mgmt_only: true
+  - name: sync
+    label: SYNC
+    type: 1000base-t
+  - name: eth1-01
```
harryajc, line 40: The most accurate way of modeling this, as these ports are part of expansion cards is to add 4 module bays to this device and then create a module for all of the different expansion cards that exist.

AlmeidaCaio: I will update this

PR #3442, opened 2025-08-25
Added Meraki-MS150-48FP-4X fixed

Kompromatas:
Added device type Meraki-MS150-48FP-4X fixed
Fixed part number for Meraki access points MRxx to match Cisco Support part numbers


device-types/Cisco/Meraki-MS150-48FP-4X.yaml
```diff
@@ -0,0 +1,222 @@
+---
+manufacturer: Cisco
+model: Meraki MS150-48FP-4X
+slug: cisco-meraki-ms150-48fp-4x
+part_number: MS150-48FP-4X
+weight: 5.4
```
harryajc, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

PR #3455, opened 2025-08-27
Create 46M4137.yaml

walday:


device-types/IBM/46M4137.yaml
```diff
@@ -0,0 +1,118 @@
+---
+manufacturer: IBM
+model: 46M4137
+slug: ibm-46m4137
+part_number: AP8888
+u_height: 0
+is_full_depth: false
+airflow: passive
```
harryajc, line 10: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

PR #3459, opened 2025-08-27
Create dt Check Point - SG9300.yaml

jo-bzh:


device-types/Check Point/SG9300.yaml
```diff
@@ -0,0 +1,65 @@
...
+comments: '[Check Point 9300 Data Sheet](https://www.checkpoint.com/downloads/products/quantum-force-9300-datasheet.pdf)'
+airflow: side-to-rear
+console-ports:
+  - name: ttyS0
+    label: CONSOLE
+    type: rj-45
+  - name: ttyS1
+    label: CONSOLE
+    type: usb-c
+power-ports:
```
harryajc: As these are field replaceable PSU they should be modeled as module-bays

jo-bzh:
Thanks @harryajc didn't take the time to update

PR #3460, opened 2025-08-27
Add APC PDU AP7822B

iriseden:
Add APC 2U 16 Power Outlet PDU (AP7822B)

### DIFF
--- a/device-types/APC/AP7822B.yml
+++ b/device-types/APC/AP7822B.yml
@@ -1,0 +1,67 @@
+---
+manufacturer: APC
+model: AP7822B
+slug: apc-ap7822b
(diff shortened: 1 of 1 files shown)


device-types/APC/AP7822B.yml
```diff
@@ -0,0 +1,67 @@
+---
+manufacturer: APC
+model: AP7822B
+slug: apc-ap7822b
+part_number: AP7822B
+u_height: 2
```
harryajc, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

PR #3465, opened 2025-08-28
added Cisco/WS-C3650-24TD-S, IP Base licensed version of WS-C3650-24TD-L

demik:
Like title says

The hardware is identical, it's just the IP Base SKU before the Smart licensing era


walday:
This device use RTU license on new IOS

device-types/Cisco/WS-C3650-24TD-S.yaml
```diff
@@ -0,0 +1,75 @@
+---
+manufacturer: Cisco
+model: Catalyst 3650-24TD-S
+slug: cisco-ws-c3650-24td-s
+part_number: WS-C3650-24TD-S
+is_full_depth: false
```
harryajc, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

demik:
Thank you. I' also added the airflow and weight on WS-C3650-24TD-L as well since it was missing

harryajc (approved):

PR #3478, opened 2025-09-01
Create DS1517+.yaml

walday:


device-types/Synology/DS1517+.yaml
```diff
@@ -0,0 +1,21 @@
+---
+manufacturer: Synology
+model: DS1517+
+slug: synology-ds1517-plus
+part_number: DS1517+
+u_height: 4
```
harryajc, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (approved):

PR #3481, opened 2025-09-01
Cisco APIC M4 and required Module Types

kmschellens:
Device:

- Cisco-APIC-M4

Module Types (required by APIC M4):

- APIC-P-I8D25GF.yaml
- APIC-P-ID10GC.yaml
- APIC-PCIE-C25Q-04.yaml


device-types/Cisco/APIC-M4.yaml
```diff
@@ -0,0 +1,42 @@
...
+part_number: APIC-M4
+u_height: 1
+is_full_depth: true
+airflow: front-to-rear
+weight: 17
+weight_unit: kg
+console-ports:
+  - name: Con
+    type: rj-45
+power-ports:
```
harryajc: Are these field replaceable PSUs? If so they should be modeled as `module-bays`

kmschellens: You're quite right - fixed!

PR #3499, opened 2025-09-03
Add R1208WFTYSR.yaml

bonktree:
Intel R1208WFTYSR (https://ark.intel.com/content/www/us/en/ark/products/192586/intel-server-system-r1208wftysr.html)


device-types/Intel/R1208WFTYSR.yaml
```diff
@@ -0,0 +1,33 @@
...
+manufacturer: Intel
+model: R1208WFTYSR
+slug: intel-r1208wftysr
+u_height: 1
+is_full_depth: true
+airflow: front-to-rear
+weight: 12.3
+weight_unit: kg
+comments: R1208WFTYSR (https://ark.intel.com/content/www/us/en/ark/products/192586/intel-server-system-r1208wftysr.html)
+power-ports:
```
harryajc: There are already module-bays for psus, are these required?

bonktree: I was having trouble making definitions for the matching PSU module types at first, that's the reason for the confusion.

The server platform has no built-in power ports. So this is not necessary.

PR #3505, opened 2025-09-04
Add FAN-040A-[BF].yaml

bonktree:
https://support.huawei.com/enterprise/en/doc/EDOC1000019246/fcf3265f/fan-040a-series-fan-modules

These can be installed in `CE5855-48T4S2Q-EI`, `CE5855-24T4S2Q-EI`.


bonktree:
> Failed: module-types/Huawei/FAN-040A-B.yaml failed validation: Additional properties are not allowed ('airflow' was unexpected)

> Failed: module-types/Huawei/FAN-040A-F.yaml failed validation: Additional properties are not allowed ('airflow' was unexpected)

That is unfortunate.
The chassis does not enforce an air flow direction, it is determined by the installed fan modules only. So an absent `airflow` field for the switch would make sense, if fan modules would have their own `airflow`.
Removing.

harryajc:
> > Failed: module-types/Huawei/FAN-040A-B.yaml failed validation: Additional properties are not allowed ('airflow' was unexpected)
> 
> > Failed: module-types/Huawei/FAN-040A-F.yaml failed validation: Additional properties are not allowed ('airflow' was unexpected)
> 
> That is unfortunate. The chassis does not enforce an air flow direction, it is determined by the installed fan modules only. So an absent `airflow` field for the switch would make sense, if fan modules would have their own `airflow`. Removing.

If you want to add support for airflow, like what was suggested under #3364 then that would be accepted as the issue with that PR is to do with the profile implementation

bonktree:
> > The chassis does not enforce an air flow direction, it is determined by the installed fan modules only. So an absent `airflow` field for the switch would make sense, if fan modules would have their own `airflow`. Removing.
> 
> If you want to add support for airflow, like what was suggested under #3364 then that would be accepted as the issue with that PR is to do with the profile implementation

@harryajc, that would be great. I've submitted a PR to address that.

harryajc:
I have merged that PR so you should be able to add the airflow property now

harryajc (approved):

PR #3514, opened 2025-09-05
Update EDGE-01U.yaml

kmorin18:
Edge 01u has 3 drawers (1-3) and 4 slots (A-D), not just 2 drawers

image

### DIFF
--- a/device-types/Corning/EDGE-01U.yaml
+++ b/device-types/Corning/EDGE-01U.yaml
@@ -25,3 +25,11 @@ module-bays:
     position: 02C
   - name: 02D
     position: 02D
+  - name: 03A
+    position: 03A
+  - name: 03B
+    position: 03B
+  - name: 03C
+    position: 03C
+  - name: 03D
+    position: 03D


harryajc:
Hi @kmorin18 

Looking at cornings website here https://ecatalog.corning.com/optical-communications/EMEA/en_GB/Fibre-Optic-Hardware/Housings/EDGE%E2%84%A2-Housing/p/EDGE-01U

It suggest that it only has 8 slots total, Are you sure the image from the device you have listed is the EDGE-01U

kmorin18:
Good find

It seems like it's the "-SP" SKU that has 3 drawers
https://ecatalog.corning.com/optical-communications/US/en/Fiber-Optic-Hardware/Housings/EDGE%E2%84%A2-Housing/p/EDGE-01U-SP

harryajc:
Okay, so the best option is to create a seperate device yaml for this sku, you can copy the current one and just make the required changes. Also note for new device we are asking for the `weight` and `airflow` properties be defined

harryajc:
Been over 7 months since last update

PR #3518, opened 2025-09-08
Create mediant-2600-SBC.yaml

fabiovalverde:
Added the Mediant 2600 SBC using official datasheet.

### DIFF
--- a/device-types/AudioCodes/mediant-2600-SBC.yaml
+++ b/device-types/AudioCodes/mediant-2600-SBC.yaml
@@ -6,17 +6,9 @@ part_number: Mediant 2600 SBC
 u_height: 1.0
 is_full_depth: true
 airflow: front-to-rear
-subdevice_role: parent
 weight: 11.7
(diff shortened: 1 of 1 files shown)


device-types/AudioCodes/mediant-2600-SBC.yaml
```diff
@@ -0,0 +1,36 @@
...
+part_number: Mediant 2600 SBC
+u_height: 1.0
+is_full_depth: true
+airflow: front-to-rear
+subdevice_role: parent
+weight: 11.7
+weight_unit: lb
+comments: '[Mediant 2600 SBC Datasheet](https://www.audiocodes.com/media/2414/mediant-2600-sbc-datasheet.pdf)'
+
+power-ports:
```
harryajc: Looking at the spec sheet, these look to be field replaceable PSUs, therefore they should be modeled as `module-bays` and then with the required PSU modules defined under the `module-types`

PR #3526, opened 2025-09-09
Added Niagara TAP 3225-PTC-2

Guibi1:
Added Niagara passive TAP 3225-PTC-2 with two single width taps.


module-types/Niagara/3211-S-LR-50-50.yaml
```diff
@@ -0,0 +1,19 @@
+---
+manufacturer: Niagara
+model: Network TAP Passive Fiber Optics Module Single Width 1310nm 1550nm 50/50
+part_number: 3211-S-LR-50/50
+comments: '[TAP Datasheet](https://info.niagaranetworks.com/hubfs/Product%20Data%20Sheets/3225/3225.pdf)'
+interfaces:
```
harryajc, line 6: If these are passive taps then they won't have 100GB interfaces they will just be LC ports

module-types/Niagara/3211-S-SR-50-50.yaml
```diff
@@ -0,0 +1,19 @@
+---
+manufacturer: Niagara
+model: Network TAP Passive Fiber Optics Module Single Width 850nm 50/50
+part_number: 3211-S-SR-50/50
+comments: '[TAP Datasheet](https://info.niagaranetworks.com/hubfs/Product%20Data%20Sheets/3225/3225.pdf)'
+interfaces:
+  - name: Port/{module}/A_Monitor
```
harryajc, line 7: Again if these are passive they should just be LC ports

device-types/Niagara/3225-PTC-2.yaml
```diff
@@ -0,0 +1,60 @@
+---
+manufacturer: Niagara
+model: Network TAP Passive Fiber Optics 3225-PTC-2
+slug: niagara-tap-3225-ptc-2
+part_number: 3225-PTC-2
+u_height: 1
+is_full_depth: false
```
harryajc, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

device-types/Niagara/3225-PTC-2.yaml
Guibi1: I didn't have the weight in the datasheet, which makes sense as most of the weight come from the modules.

module-types/Niagara/3211-S-LR-50-50.yaml
Guibi1: Yes they are passive but I can't put lc as an interface type. Should I move them to the front/rear ports?

harryajc: Yes these should be probably front ports linked to the same rear ports, I guess is the best way

harryajc (changes requested):

harryajc (approved):

PR #3534, opened 2025-09-12
Add Quantum Scalar i3

mathieumd:


device-types/Quantum/Scalar-i3.yaml
```diff
@@ -0,0 +1,29 @@
+---
+manufacturer: Quantum
+model: Scalar i3
+slug: quantum-scalar-i3
+description: LTO library for long-term data archiving
+u_height: 3.0
+is_full_depth: true
```
harryajc, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

device-types/Quantum/Scalar-i3.yaml
mathieumd: Of course, but unfortunately I could not find any mention of these into [the official datasheet](https://cdn.allbound.com/iq-ab/2021/11/Scalar-Tape-DS00559A.pdf). And ours is already in production, so I cannot weight it anymore.

PR #3559, opened 2025-09-18
added SNR-S2982G-24T-POE

pirrazium:
added new device


device-types/SNR/S2982G-24T-POE.yaml
```diff
@@ -0,0 +1,178 @@
+---
+manufacturer: SNR
+model: S2982G-24T-POE
+slug: snr-s2982g-24t-poe
+part_number: S2982G-24T-POE
+u_height: 1.0
+is_full_depth: false
+airflow: right-to-left
```
harryajc, line 8: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc: These Datapoints Have not been added

PR #3575, opened 2025-09-25
Add SG19200

adam5isalive:
Adding Check Point SG19200


adam5isalive (commented):
Should I also add the modules I created, or should I do each of those seperately?

harryajc:
I'll merge this device, then you can create a new PR for the modules

PR #3580, opened 2025-09-26
Update shelf-drawer-1u-full-depth.yaml

LLFT:
Updated the model to manage the parent/child role


walday:
Why exactly 2 slots?

LLFT:
> Why exactly 2 slots?

no reason, but at least 1 because this type of product is necessarily **Parent** of something.

PR #3594, opened 2025-10-01
Create 1751-V.yaml

walday:


device-types/Cisco/1751-V.yaml
```diff
@@ -0,0 +1,29 @@
...
+module-bays:
+  - name: VIC-WIC0
+    position: '0'
+  - name: VIC-WIC1
+    position: '1'
+  - name: VIC2
+    position: '2'
+power-ports:
+  - name: PSU
+    type: iec-60320-c6
```
harryajc: Looking at the spec sheet it suggests that this is a dc-terminal

harryajc (changes requested):

device-types/Cisco/1751-V.yaml
walday: i'm fix it

PR #3614, opened 2025-10-02
Create DES-1210-08P.yaml

walday:


device-types/D-Link/DES-1210-08P.yaml
```diff
@@ -0,0 +1,48 @@
+---
+manufacturer: D-Link
+model: DES-1210-08P
+slug: d-link-des-1210-08p
+part_number: DES-1210-08P
+comments: '[D-Link DES-1210-08P](https://www.dlink.ru/ru/products/1/2054.html)'
+is_full_depth: false
```
harryajc, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

device-types/D-Link/DES-1210-08P.yaml
walday: Hi! Fix it

PR #3623, opened 2025-10-04
Add rackifyUS devices

phillf:
(Seeking input)

Adding devices from [RackifyUS Etsy Shop](https://www.etsy.com/shop/RackifyUS?listing_id=4357803151)


phillf:
@harryajc / @jeremystretch, would you mind reviewing and letting me know if I need to make and changes?

phillf:
> For devices that you have added images for you need to uncomment the image: true lines.
I have done this and removed the superfluous rear_image lines where applicable.

Intent is to add the rest of the images at a later date and uncomment the YAML when appropriate.

>     1. Please add the `airflow` data point if you have it or can find it. Often these are listed in the datasheet, but sometimes they are not
Air flow is dictated by the device that mounts inside the bracket.

>     2. Please add the `weight` and associated `weight_unit` data point if you have it or can find it. Often these are listed in the datasheet, but sometimes they are not
I am awaiting input from the designer of the items but I don't think he ever thought to weigh them.

PR #3632, opened 2025-10-06
Fortinet: add FIM-7941F

fabienarchambault:


device-types/Fortinet/FIM-7941F.yaml
```diff
@@ -0,0 +1,79 @@
+---
+manufacturer: Fortinet
+model: Fortinet Interface Module
+part_number: FIM-7941F
+slug: fortinet-fim-7941f
+is_full_depth: true
```
harryajc, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

device-types/Fortinet/FIM-7941F.yaml
```diff
@@ -0,0 +1,79 @@
...
+# SFP28
+  - name: m3
+    type: 25gbase-x-sfp28
+  - name: m4
+    type: 25gbase-x-sfp28
+console-ports:
+  - name: Console
+    type: rj-45
+# Add fictive power source
+# All power supplies are within the chassis
```
harryajc: If this goes within a chassis then there are 2 options:
1. Either this device just provides ports to the main chassis and has none of it own management therefore it should be modeled as a 'module-type'
2. or that this is a device in it's own right therefore having it own management, then rather than adding module-bays that don't exist just add 'subdevice_role: child'

harryajc (changes requested):

PR #3636, opened 2025-10-07
Fortinet: add FPM-7620F

fabienarchambault:


device-types/Fortinet/FPM-7620F.yaml
```diff
@@ -0,0 +1,43 @@
...
+    type: 25gbase-x-sfp28
+  - name: port7
+    type: 25gbase-x-sfp28
+  - name: port8
+    type: 25gbase-x-sfp28
+  - name: port9
+    type: 25gbase-x-sfp28
+  - name: port10
+    type: 25gbase-x-sfp28
+module-bays:
```
harryajc: If this is supposed to be part of a chassis as well then then use the following: 'subdevice_role: child'

harryajc (changes requested):

PR #3642, opened 2025-10-07
Modify PID/image, Cisco 8818 device-type.

segachan:


segachan:
@harryajc 
> What is the reason for changing the model name?

The reason is why the model name is mistook.

PR #3643, opened 2025-10-07
Add Nvidia SN5600 switch

fabienarchambault:


device-types/Nvidia/SN5600.yaml
```diff
@@ -0,0 +1,148 @@
+---
+manufacturer: Nvidia
+model: SN5600
+slug: nvidia-sn5600
+u_height: 2
+is_full_depth: true
```
harryajc, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

device-types/Nvidia/SN5600.yaml
```diff
@@ -0,0 +1,148 @@
...
+comments: |
+  Data sheet: https://nvdam.widen.net/s/mmvbnpk8qk/networking-ethernet-switches-sn5000-datasheet-us
+console-ports:
+  - name: con0
+    type: rj-45
+power-ports:
```
harryajc: Looking at the spec sheet these are Field replaceable PSUs therefore they should be modeled as 'module-bays'

harryajc (changes requested):

device-types/Nvidia/SN5600.yaml
fabienarchambault: Thank you for the review. Can you detail more how to do it?
If I compare to the SN4600 from Mellanox (previous model), the `power-ports` are modelled in the same way.

device-types/Nvidia/SN5600.yaml
fabienarchambault: Commited information

device-types/Nvidia/SN5600.yaml
fabienarchambault: Found the solution in anothe Mellanox switch ;)

PR #3644, opened 2025-10-07
Add Nvidia QM9700 InfiniBand switch

fabienarchambault:


device-types/Nvidia/QM9700.yaml
```diff
@@ -0,0 +1,139 @@
+---
+manufacturer: Nvidia
+model: QM9700
+slug: nvidia-qm9700
+u_height: 1
+is_full_depth: true
+console-ports:
+  - name: uart
+    type: rj-45
+power-ports:
```
harryajc: Are these Field Replaceable PSUs, if so they should be modeled as 'module-bays'

device-types/Nvidia/QM9700.yaml
```diff
@@ -0,0 +1,139 @@
+---
+manufacturer: Nvidia
+model: QM9700
+slug: nvidia-qm9700
+u_height: 1
```
harryajc, line 5: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

PR #3646, opened 2025-10-08
Local

LLFT:
Correction following Harryajc's remarks


device-types/HPE/MSA-3600-LFF-Disk-Enclosure.yaml
```diff
@@ -8,12 +8,18 @@ is_full_depth: true
...
 module-bays:
-  - name: PS1
-    position: PS1
-  - name: PS2
-    position: PS2
-  - name: Ctrl A
-    position: A
-  - name: Ctrl B
-    position: B
+  - name: PSU1
```
harryajc, line 14: What is the reason for changing these module-bay names?

LLFT: You're right, the bays don't need to be renamed. To save time, we integrated the module name directly into the bay. But we need to add the fan bays.
We also need to create the "HPE 3.5 LFF I/O" modules if they don't exist.

PR #3652, opened 2025-10-10
Add device Raisecom ISCOM2948GF-4C-AC/DC/D

sysadmin-pg:
Raisecom ISCOM2948GF-4C-AC/DC/D Enhanced Aggregation Ethernet Switch

### DIFF
--- a/device-types/Raisecom/ISCOM2948GF-4C.yaml
+++ b/device-types/Raisecom/ISCOM2948GF-4C.yaml
@@ -1,0 +1,129 @@
+---
+manufacturer: Raisecom
+model: ISCOM2948GF-4C-AC/DC/D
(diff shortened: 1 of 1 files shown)


device-types/Raisecom/ISCOM2948GF-4C.yaml
```diff
@@ -0,0 +1,129 @@
...
+part_number: ISCOM2948GF-4C-AC/DC/D
+u_height: 1.0
+is_full_depth: false
+airflow: front-to-rear
+weight: 8.0
+weight_unit: kg
+console-ports:
+  - name: Console
+    type: rj-45
+power-ports:
```
harryajc, line 14: Are these field replaceable PSUs? if so they should be modeled as 'module-bays'

PR #3660, opened 2025-10-16
Adding template for Juniper ACX7020-AC

modonovan:
This pull request adds a device type template and front elevation image (taken from the Juniper Datasheet for the device) for [Juniper ACX7020-AC router](https://www.juniper.net/us/en/products/routers/acx-series/acx7000-family-cloud-metro-routers/acx7000-line/acx7020-cloud-metro-routers-datasheet.html)


device-types/Juniper/ACX7020-AC.yaml
```diff
@@ -0,0 +1,69 @@
...
+  - name: et-0/0/16
+    type: 10gbase-x-sfpp
+  - name: et-0/0/17
+    type: 10gbase-x-sfpp
+  - name: et-0/0/18
+    type: 10gbase-x-sfpp
+  - name: et-0/0/19 # This interface is disabled on the router if PTP (precision time protocol) is enabled
+    type: 10gbase-x-sfpp
+module-bays:
+  - name: PSM 0
```
harryajc, line 62: As these are built in PSU they should be modeled as `power-ports`.

device-types/Juniper/ACX7020-AC.yaml
```diff
@@ -0,0 +1,69 @@
...
+    # Comment: Power Supply Module 0
+    position: '0'
+  - name: PSM 1
+    # Comment: Power Supply Module 1
+    position: '1'
+  - name: Fan Module 0
```
harryajc: Looks like the fan is built-in too, so this is not required.

device-types/Juniper/ACX7020-AC.yaml
modonovan: Updated now via new commit - thanks for spotting this.

PR #3674, opened 2025-10-20
Optimize pre-commit download

### LINKED ISSUE #3322: Optimize download amount of PyTest pre-commit check
When running the PyTest pre-commit check with the (default) configuration of [`USE_LOCAL_KNOWN_SLUGS = False`](https://github.com/netbox-community/devicetype-library/blob/5feee391/tests/test_configuration.py#L36), it will fetch an up-to-date copy of the slug pickles from Github. To get this ~400KiB of needed data, it will do a fresh [full clone](https://github.com/netbox-community/devicetype-library/blob/5feee391/tests/definitions_test.py#L139) of this here `devicetype-library` git repository to a temporary directory, which comes to a total download size of ~610MiB (and about double that in used disk/ram space). That's an overhead of over 1500%! For every single invocation of the hook!

My first thought was: Easy fix, make the clone shallow and limit it to the last commit only. But even then the download is ~526MiB, which is better but still not great and still way too much. Looking at the shallow copy, ~95% of the space is taken by elevation images. So to get the download size down further, you need to make the repo clone a sparse one, where you only fetch the relevant `tests` sub-directory of the repo.

Doing the repo clone using `--local` or `--shared` could also speed things up, but there is no guarantee that the pre-commit check will be run from an actual git repository, it could also be using a shell from an unpacked source archive.

At this point I was taking a step back, and asked myself: Why even bother with a git clone at this point, if all you need is the latest version of two small files. It may be easier to replace the whole repo clone with a simple HTTPS download of these two files. If you don't want to pull in another dependency like `requests` for this, it can also be done using the python stdlib (`urllib.request` and friends). Doing it naively by fetching the files raw from the `master` branch, you don't have the guarantee that both files belong to the same commit (imagine a new commit is pushed while you are downloading the first file). But this can be prevented by doing three simple HTTP requests: Get the current commit (id) of the `master` branch, get file A from that exact commit, and then get file B from that commit.

Any thoughts on this?

m-hau:
Optimize the amount of data the pre-commit hook downloads, significantly reducing it from ~600MiB to ~500KiB. This is done by using a Git feature named [Partial Clone](https://git-scm.com/docs/partial-clone), which is supported since 2.18 (released 2018). With older Git versions it falls back to the previous behaviour of doing a full clone.

Fixes #3322


harryajc:
@danner26 @DanSheps Thoughts on this?

danner26:
@m-hau I will have to dig into this further, but I believe this is just cloning the head commit (i.e. the latest in the branch), is that correct?

m-hau:
Using `git clone https://...` to do a "normal" clone, it will a) download the whole history (all revisions of all past and present files) into the `.git` directory and b) populate the working directory with the files/state of the last commit. This will download 654MiB of data, and consume 1,31GiB of disk-space (or RAM in the case of tmpfs). That is what the hook is currently doing.

Using `git clone --depth=1 https://...` you are doing a "shallow clone", which reduces a) from the whole history to just the last commit. This will download 561MiB of data, and consume 1,22GiB of disk-space (or RAM in the case of tmpfs). Most of this data are the elevation-images, which are not required for the hook.

Using `git clone --depth=1 --no-checkout https://...` you are skipping b) entirely, and can then selectively checkout the files you actually need in your working copy. This will download 561MiB of data, and consume 562MiB of disk-space (or RAM in the case of tmpfs). That is what the new hook is doing as fallback.

Using `git clone --depth=1 --no-checkout --filter=blob:none https://...` you are doing a "partial clone", reducing a) to just the metadata of the last commit (skipping any file contents) and skipping b) entirely. You can then selectively checkout the files you actually need in your working copy, which will automatically trigger a download of the required file contents. This will download 372KiB of data, and consume 868KiB of disk-space (or RAM in the case of tmpfs). That is what the new hook is doing.

So yeah, reducing the download amount and disk/ram-usage required a bit more than "just cloning the head commit", because the elevation-images make even just the last commit huge.

marcinpsk:
@m-hau that's a nice one, cut pre-commit check for me from 'let's cross fingers and throw salt over a left shoulder' to under 10s, on crappy hotel wifi.

PR #3691, opened 2025-10-24
added device definition and elevation images for CyberPower PR2000RT2UC

pkropf:
Added .yaml device definition file plus front and back elevation images for a CyberPower PR2000RT2UC.


pkropf:
I need to make some changes to the definition. It's missing the module bay, the image names are incorrect and they need to be referenced properly. Will submit another request once these changes have been made.

PR #3699, opened 2025-10-28
Add Edgecore AS4625-54T

jzehnter:
Add Edgecore models (with images):
- AS4625-54T-O-AC-F (front-to-rear)
- AS4625-54T-O-AC-B (rear-to-front


device-types/Edgecore/AS4625-54T-O-AC-B.yaml
```diff
@@ -0,0 +1,146 @@
...
+  - name: Console
+    type: rj-45
+
+module-bays:
+  - name: PSU1
+    position: '1'
+  - name: PSU2
+    position: '2'
+
+inventory-items:
```
harryajc: These should be modeled as `module-bays`

device-types/Edgecore/AS4625-54T-O-AC-F.yaml
```diff
@@ -0,0 +1,146 @@
...
+  - name: Console
+    type: rj-45
+
+module-bays:
+  - name: PSU1
+    position: '1'
+  - name: PSU2
+    position: '2'
+
+inventory-items:
```
harryajc: These should be modeled as `module-bays`

device-types/Edgecore/AS4625-54T-O-AC-B.yaml
jzehnter: Did you mean the inventory-items (FANs?)
They are fixed according to the datasheet:
> 9. Fixed 2 + 1 redundant fans (FAN1, FAN2, FAN3)

harryajc: If they are fixed then they are not required

device-types/Edgecore/AS4625-54T-O-AC-F.yaml
jzehnter: removed blank lines, inventory-items and lo-interface

PR #3707, opened 2025-11-03
Add CP Plus CP-ANW-HPU24G2F2-N30

kaydee123:


device-types/CP Plus/cp-anw-hpu24g2f2-n30.yaml
```diff
@@ -0,0 +1,148 @@
+---
+manufacturer: CP Plus
+model: CP-ANW-HPU24G2F2-N30
+slug: cp-plus-cp-anw-hpu24g2f2-n30
+part_number: CP-ANW-HPU24G2F2-N30
+u_height: 1.0
+is_full_depth: false
+weight: 2.65
+weight_unit: kg
```
harryajc, line 10: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

PR #3710, opened 2025-11-03
Create Juniper MX2020 device-type.

segachan:


segachan:
@harryajc Please review this.

PR #3724, opened 2025-11-10
New Rack-types for a few Hammond racks

ppk-cyber:
Here are a few Rack-types for Hammond racks


ppk-cyber:
Model names were revised to match file names

PR #3725, opened 2025-11-11
add Nvidia SN3420

cambid:
This PR adds the Nvidia SN3420 (based on https://marketplace.nvidia.com/en-us/enterprise/networking/sn3420/)


device-types/Nvidia/SN3420.yml
```diff
@@ -0,0 +1,140 @@
...
+comments: Nvidia SN3420 (https://nvdam.widen.net/s/gbk7knpsfd/sn3000-series)
+slug: nvidia-sn3420
+u_height: 1
+is_full_depth: true
+weight: 8.5
+weight_unit: kg
+console-ports:
+  - name: con0
+    type: rj-45
+power-ports:
```
harryajc: The spec sheet says these are hot swappable PSUs therefore they should be modeled as `module-bays`

device-types/Nvidia/SN3420.yml
```diff
@@ -0,0 +1,140 @@
+---
+manufacturer: Nvidia
+model: SN3420
+comments: Nvidia SN3420 (https://nvdam.widen.net/s/gbk7knpsfd/sn3000-series)
+slug: nvidia-sn3420
+u_height: 1
+is_full_depth: true
+weight: 8.5
+weight_unit: kg
```
harryajc, line 9: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

device-types/Nvidia/SN3420.yml
cambid: The SN3420 is available with both airflows front-to-rear and rear-to-front.

device-types/Nvidia/SN3420.yml
cambid: I've changed the PSU to module-bays.

PR #3743, opened 2025-11-15
Add device type HT-EM4 for Hitron Technologies

cyberpunk2350:
adding Hitron HT-EM4 Moca Adaptor


cyberpunk2350:
So I'm not sure what is causing the format test to fail, I could be wrong but it seems like it's some kind of time out.

harryajc:
The empty fields was causing the pre checks to fail, The manufacturer issue will need sorting also there is no defined power-ports

Issue #3747, opened 2025-11-17
Dell R7525 has rear image of Dell R7515


Howard-A2Labs:
Hello there,

I noticed the [rear image for the Dell R7525](https://github.com/netbox-community/devicetype-library/blob/master/elevation-images/Dell/dell-poweredge-r7525.rear.jpg) is not an R7525 when adding my servers to my racks after pulling the library into my new Netbox instance. It appears to be a copy of [the R7515 rear](https://github.com/netbox-community/devicetype-library/blob/master/elevation-images/Dell/dell-poweredge-r7515.rear.jpg) which is totally different.

Here you can see what the rear of them looks like, though I am not sure where to find a clean version without the labeling numbers and lines.

<img width="1630" height="508" alt="Image" src="https://github.com/user-attachments/assets/d631a2e7-18eb-4bc9-9be3-001bd37e27db" />

PR #3748, opened 2025-11-17
Replacement of R7525 Rear Image

Juniper46:


Juniper46:
Fixes issue discussed in #3747

PR #3752, opened 2025-11-20
add FS C6P-U24FT1U

cambid:
This PR adds the FS U24FT1U (https://www.fs.com/uk/products/69180.html)

### DIFF
--- a/device-types/FS/C6P-U24FT1U.yaml
+++ b/device-types/FS/C6P-U24FT1U.yaml
@@ -4,81 +4,84 @@ model: C6P-U24FT1U
 slug: fs-c6p-u24ft1u
 part_number: '69180'
 u_height: 1
+is_full_depth: false
+airflow: passive
+is_powered: false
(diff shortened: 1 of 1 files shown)


device-types/FS/C6P-U24FT1U.yaml
```diff
@@ -0,0 +1,181 @@
+---
+manufacturer: FS
+model: C6P-U24FT1U
+slug: fs-c6p-u24ft1u
+part_number: '69180'
+u_height: 1
+is_full_depth: false
+airflow: passive
```
harryajc, line 8: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

cambid: Unfortunately the weight is not mentioned on the datasheet or the package of the panel.

harryajc (approved):

PR #3759, opened 2025-11-21
Add device type for Cisco Catalyst 9800-80-K9

lukejenkins:
Add Cisco C9800-80-K9 Wireless Controller. Not too many changes from the existing C9800-40-K9.yaml.


device-types/Cisco/C9800-80-K9.yaml
```diff
@@ -0,0 +1,49 @@
...
+    type: 10gbase-x-sfpp
+  - name: TenGigabitEthernet0/0/4
+    type: 10gbase-x-sfpp
+  - name: TenGigabitEthernet0/0/5
+    type: 10gbase-x-sfpp
+  - name: TenGigabitEthernet0/0/6
+    type: 10gbase-x-sfpp
+  - name: TenGigabitEthernet0/0/7
+    type: 10gbase-x-sfpp
+power-ports:
```
harryajc: These are hot swappable PSUs therefore they should be modeled as `module-bays`

lukejenkins: Done, I've also added the expansion slot and module-type files for the supporting power supply types and the missing expansion slot types that are supported.

PR #3761, opened 2025-11-21
fix(module): extra parentheses in filename

NeverBehave:
slug function in test will remove parentheses from filename. 

This make test happy


NeverBehave:
MB, the module type has a separate replace rule.

NeverBehave:
The difference seems to be module type dont have the three extra rules: `.replace("(", "").replace(")", "").replace(";", "")`. 

Is there a reason why?

PR #3785, opened 2025-12-07
Add EFB Elektronik 1U Shelf Drawer Model 691651ATS8

Napsty:
Add a 1U Shelf Drawer from EFB Elektronik, with front image.


Napsty:
Thanks @harryajc . I was wondering why the lint would fail. So it's because of the slug model name requiring small caps?

harryajc:
> Thanks @harryajc . I was wondering why the lint would fail. So it's because of the slug model name requiring small caps?

Yes, the slug must only be lowercase, numbers or -

PR #3792, opened 2025-12-11
Add new smartoptics module

netnoddy:
Added smartopics passive DWDM module


device-types/Smartoptics/H-MD-09-921-928-4C.yaml
harryajc, line 1: Looking at the spec sheet it looks like this is a module that goes into a chassis, therefore this should be modeled as a module

netnoddy: No this is a standalone passive DWDM device no need for a chassis, there's no power or anything like that.
You can mount it directly into a rack if you want.

harryajc (approved):

PR #3799, opened 2025-12-15
Add new device type

netnoddy:


netnoddy:
Add new Cisco MDS fiber-channel switch

PR #3803, opened 2025-12-19
Add missing MGMT port for MikroTik CRS354-48G/P

grische:


grische:
@harryajc do you need anything from me on this one and the other mgmt port?

harryajc:
@grische Usually mgmt ports follow a different naming than the standard network ports, unless this is different in Mikrotik

grische:
That's how it shows up in the console for those models. Not sure about other models though.

PR #3999, opened 2026-03-08
Update to match new Advanced Port Mappings

odiseo123:
Old port mappings syntax no longer working after v4.5.2 (2026-02-03) update. 

Refer to https://github.com/netbox-community/netbox/issues/20564, https://netboxlabs.com/docs/netbox/release-notes/version-4.5#advanced-port-mappings-20564 and  https://github.com/netbox-community/netbox/issues/21539

All credits to @arthanson

### DIFF
--- a/device-types/Generic/24-port-copper-patch-panel.yaml
+++ b/device-types/Generic/24-port-copper-patch-panel.yaml
@@ -8,124 +8,173 @@ airflow: passive
 is_powered: false
 front_image: true
 front-ports:
-  - name: Port 1
+  - name: Front 1
     type: 8p8c
-    rear_port: Port 1
-  - name: Port 2
+  - name: Front 2
     type: 8p8c
-    rear_port: Port 2
-  - name: Port 3
+  - name: Front 3
     type: 8p8c
-    rear_port: Port 3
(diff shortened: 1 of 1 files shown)


odiseo123:
The units test fails as it seems to also be following the old port mapping syntax which requires a "rear_port" property specified in each "front_port", instead of using the mappings.

`FAILED tests/definitions_test.py::test_definitions[device-types/Generic/24-port-copper-patch-panel.yaml-schema0-M] - Failed: device-types/Generic/24-port-copper-patch-panel.yaml failed validation: 'rear_port' is a required property`

I am not a full time programmer so I am not confident editing the unit tests. However, I can confirm that the modified device-type works correctly on dockerized Netbox v4.5.3

image

Maybe one of the maintainers could update the unit tests, while reviewing this PR. Otherwise, the solution to this issue is documented here: https://github.com/netbox-community/netbox/issues/21539

harryajc:
@odiseo123 This change will require that the pre-commit scripts and schema is updated to support this change

harryajc:
This has been updated as part of the commit merging the new port mapping schema

PR #4051, opened 2026-03-19
Add Vertiv PowerIT horizontal rack PDUs

aaronaxvig:
Add the horizontal PDU models that are listed on page 4 here: https://www.vertiv.com/494d03/globalassets/shared/vertiv-rack-pdu-br-en-na-sl-20881_238947_01.pdf

Two notes for discussion:

1. I believe most products in the repo use the model number as the file name.  I chose to use the part number as the file name.  I think Vertiv uses part and model kind of backwards.  They use the part number as the "friendly" customer facing name, and model number as long complex thing.
Example
https://www.geistglobal.com/power-distribution-unit/VP5N11A0
Part #: VP5N11A0 (larger text, also used in the product list--first PDF I linked to)
Model Number: 
MNS5M1R5-06S153-1ST5A0H10-S (smaller secondary text, complex, difficult to read and refer to)
So that is my rationale for what I did.  If that is unsuitable or there is a hard requirement otherwise, I would propose leaving the file names and swapping the model and part number values in the yaml files.

2. A lot of these PDUs have the combo C13/19 power outlets.  It is kind of funny to use an Eaton named thing on a Vertiv product.  The eaton-c39 outlet type was added here: https://github.com/netbox-community/netbox/issues/17471.  There appears to be some unwillingness to add a generic combo-c13-c19 port type: https://github.com/netbox-community/netbox/issues/18510  I may write up a new proposal since that one died for lack of effort I think.  But in the meantime I chose to use the eaton-c39 port on these PDUs.  There is one other APC template and two ZPAS ones in the repo that use it.  And no Eaton ones...


aaronaxvig:
I realized that the model number is what shows in the UI so swapped the model and part_number values.  And fixed slugs to match.

harryajc (approved):

PR #4096, opened 2026-04-01
Add platform definitions and default_platform support

NeverBehave:
## Summary

Addresses #4094 — adds platform support to the device type library.

- **New `platforms/` directory** with `<Manufacturer>/<slug>.yaml` structure for platform definitions
- **New `schema/platformtype.json`** validating platform YAML (manufacturer, name, slug, description)
- **`default_platform` field** added to `schema/devicetype.json` as optional slug string
- **Cross-validation** ensures `default_platform` references an existing platform file
- **Seed platforms**: Arista EOS, Cisco IOS-XE, Cisco NX-OS, Juniper Junos, Palo Alto PAN-OS

### Why a separate `platforms/` directory?

A bare `default_platform` string would be a dangling reference — import tools can't resolve a platform that doesn't exist in the target NetBox instance. The `platforms/` directory gives import tools a complete dependency chain: create platforms first, then device types.

Additionally, a device's platform doesn't always match its manufacturer (e.g., Edgecore switches running Cumulus Linux). Platforms need their own manufacturer field, which is why they live in dedicated files under their own manufacturer directory.

### Test changes

- Platform files validated against `platformtype.json` schema
- Platform filename must match `slug` field
- `default_platform` in device types must reference an existing `platforms/*/<slug>.yaml`
- Duplicate platform slug detection across manufacturers
- Precomputed platform slug lookup for performance
- Device-only validations (verify_filename, validate_components) skipped for platforms

## Test plan

- [x] All 5 seed platform files pass schema validation
- [x] Invalid slug pattern rejected by schema
- [x] Missing required fields rejected by schema
- [x] Extra fields rejected by `additionalProperties: false`
- [x] Cross-validation correctly resolves valid platform slugs
- [x] Cross-validation correctly rejects nonexistent platform slugs
- [x] Filename=slug enforcement verified

### DIFF
--- a/platforms/Arista/arista-eos.yaml
+++ b/platforms/Arista/arista-eos.yaml
@@ -1,0 +1,5 @@
+---
+manufacturer: Arista
+name: Arista EOS
+slug: arista-eos
+description: Arista Extensible Operating System
--- a/platforms/Cisco/cisco-ios-xe.yaml
+++ b/platforms/Cisco/cisco-ios-xe.yaml
@@ -1,0 +1,5 @@
+---
+manufacturer: Cisco
+name: Cisco IOS-XE
+slug: cisco-ios-xe
+description: Cisco Internet Operating System XE
--- a/platforms/Cisco/cisco-nxos.yaml
+++ b/platforms/Cisco/cisco-nxos.yaml
@@ -1,0 +1,5 @@
+---
+manufacturer: Cisco
+name: Cisco NX-OS
+slug: cisco-nxos
+description: Cisco Nexus Operating System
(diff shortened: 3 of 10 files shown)


cflanderscpc:
Would love to see this one get added/merged.  Getting as many pre-seed options into the library as reasonably possible is a huge help for initial onboarding

NeverBehave:
We have migrated away from this community repo as upstream features are stale and lack of feedback.

Thus unfortunately I might not be able to push this forward in the near future. Feel free to pick it up tho.

PR #4105, opened 2026-04-02
Added NetApp StorageGRID SG5812 device type

Solitarystate:
Added Netapp StorageGrid Model SG5812 Appliance under the netapp folder including its rear image borrowed from hwu netapp website. 
I have run the precommit and everything looks good except file naming conflict for another section that I had not contributed to
```
=================================== FAILURES ===================================
______ test_definitions[module-types/Bastion/DL801.yaml-schema5613-skip] _______
module-types/Bastion/DL801.yaml appears to be duplicated. Found 2 matches: DL801, DL801
______ test_definitions[module-types/MegaTec/DL801.yaml-schema6933-skip] _______
module-types/MegaTec/DL801.yaml appears to be duplicated. Found 2 matches: DL801, DL801
=========================== short test summary info ============================
FAILED tests/definitions_test.py::test_definitions[module-types/Bastion/DL801.yaml-schema5613-skip] - Failed: module-types/Bastion/DL801.yaml appears to be duplicated. Found 2 m...
FAILED tests/definitions_test.py::test_definitions[module-types/MegaTec/DL801.yaml-schema6933-skip] - Failed: module-types/MegaTec/DL801.yaml appears to be duplicated. Found 2 m...
================== 2 failed, 7315 passed, 1 skipped in 37.55s ==================
```
kindly check and merge to master.


device-types/NetApp/storagegrid-sg5812.yaml
```diff
@@ -0,0 +1,38 @@
...
+manufacturer: NetApp
+model: StorageGRID SG5812
+slug: netapp-sg5812
+u_height: 2
+is_full_depth: true
+weight: 29.45
+weight_unit: kg
+rear_image: true
+power-ports:
+  - name: PSU1
```
harryajc, line 12: As these are FRUs, they should be modeled as `module-bays`

device-types/NetApp/storagegrid-sg5812.yaml
```diff
@@ -0,0 +1,38 @@
...
+    type: iec-60320-c14
+  - name: PSU2
+    type: iec-60320-c14
+interfaces:
```
harryajc, line 16: The correct way to model SAN devices with 2 seperate storage controllers is that the chassis has 2 `device-bays` and then the storage controllers are a modeled as a seperate device which has the specific interfaces

harryajc (changes requested):

device-types/NetApp/storagegrid-sg5812.yaml
Solitarystate: Thanks for the review and the pointers about how SAN devices are typically modeled.

In this specific case, the SG5812 is a StorageGRID storage + compute appliance, not a generic dual‑controller SAN shelf:

- In NetApp Hardware Universe, SG5812 is listed as a “Storage + Compute Node” and the specs show Controller Redundancy: Simplex – there is only a single storage controller in this chassis, not two hot‑swappable controller FRUs. The storage node being an iscsi direct connected storage part of the same chassis

- The compute and storage functions are integrated into the same 2U appliance, and there are no separate storage controller canisters that would make sense to model as independent child devices in NetBox.

Because of that, I modeled ‎`storagegrid-sg5812` as a single device with:

- IEC power inlets (for the built‑in PSUs), and

- the four 10/25GbE data ports plus the 1GbE management ports used by StorageGRID.

If we modeled this as a chassis with two storage controller device‑bays, it would misrepresent the actual hardware (and imply a dual‑controller SAN shelf that doesn’t exist here).

Does it make sense to keep this as a single monolithic appliance device type for SG5812, consistent with the Hardware Universe “Simplex” controller model? I’m happy to tweak names/labels or add extra detail if you’d like it shaped a bit differently.

device-types/NetApp/storagegrid-sg5812.yaml
Solitarystate: See my comment below explaining why I didn't choose module-bays. Let me know if you approve

device-types/NetApp/storagegrid-sg5812.yaml
Solitarystate: @harryajc Can you read my previous comment and let me know if you can merge this.

Solitarystate: @harryajc Can we take this forward?

harryajc: @Solitarystate Okay, happy for the controller and compute to be part of the device, my only question is are the PSUs hot swappable? Because it looks like they are and if that is the case they should be `module-bays`

Solitarystate: @harryajc You are right. I've done the necessary changes and updated this PR. Please approve and acceept the request. Thanks!

harryajc (approved):

PR #4111, opened 2026-04-07
Add mikrotik devices

Flo-95:
I added the following devices; their weight is not listed in the datasheet.
For the "RB960PGS-PB," the website states a weight of 0.7–0.9 kg. I left it blank since I don’t have exact information.


device-types/MikroTik/RB960PGS-PB.yaml
```diff
@@ -0,0 +1,35 @@
+---
+manufacturer: MikroTik
+model: PowerBox Pro
+slug: mikrotik-powerbox-pro-rb960pgs-pb
+part_number: RB960PGS-PB
+comments: '[MikroTik RB960PGS-PB Datasheet](https://mikrotik.com/product/RB960PGS-PB)'
+is_full_depth: false
+u_height: 0
```
harryajc, line 8: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

device-types/MikroTik/RB960PGS-PB.yaml
Flo-95: See my First Message  .... 

`For the "RB960PGS-PB," the website states a weight of 0.7–0.9 kg. I left it blank since I don’t have exact information.`

Should I use the average or the maximum value?

harryajc (approved):

PR #4115, opened 2026-04-08
Add Huawei OceanStor Dorado 2100

miquelmauri:


device-types/Huawei/OceanStor-Dorado-2100.yaml
harryajc, line 1: The correct way to model SAN devices with 2 separate storage controllers is that the chassis has 2 `device-bays` and then the storage controllers are a modeled as a separate device which has the specific interfaces

device-types/Huawei/OceanStor-Dorado-2100.yaml
```diff
@@ -0,0 +1,69 @@
...
+  - name: CTE0.B.P2
+    type: 1000base-t
+  - name: CTE0.B.P3
+    type: 1000base-t
+  - name: CTE0.B.P4
+    type: 10gbase-x-sfpp
+  - name: CTE0.B.P5
+    type: 10gbase-x-sfpp
+
+power-ports:
```
harryajc, line 65: As these look to be hot swappable PSUs they should be modeled as `module-bays`

harryajc (changes requested):

PR #4168, opened 2026-05-04
Lenovo Thinksystem updates

tfoks:
Add new Lenovo systems:
- SR630 v2
- SR630 v3
- SR650 v3
- SR650 v4
- SR680a v3
- DG5200

Adjusted definitions:
- SR630
- SR650 v2
- SR650

Moved modules from device-types to module-type and added some more modules


device-types/Lenovo/ThinkSystem-DG5200.yaml
harryajc, line 1: The correct way to model SAN devices with 2 separate storage controllers is that the chassis has 2 `device-bays` and then the storage controllers are a modelled as a separate device which has the specific interfaces

device-types/Lenovo/ThinkSystem-DG5200.yaml
```diff
@@ -0,0 +1,55 @@
...
+u_height: 2
+is_full_depth: true
+airflow: front-to-rear
+front_image: true
+rear_image: true
+comments: '[Lenovo System DG5200 product guide](https://lenovopress.lenovo.com/lp2074)'
+subdevice_role: parent
+weight: 25.3
+weight_unit: kg
+power-ports:
```
harryajc, line 15: As these are FRUs, they should be modelled as 'module-bays'

harryajc (changes requested):

device-types/Lenovo/ThinkSystem-DG5200.yaml
tfoks: In essence I agree. Unfortunately there're no "products" for these controllers, i.e. you cannot order the system without these and there's no choice between different controllers. The same applies to the PSUs. Although the controllers and PSUs are FRUs they cannot be ordered separately.
Therefore I included them into the device and not as separate module-types.
It is different for the NIC modules in the controllers. Thus these are modelled as "module-bays".

device-types/Lenovo/ThinkSystem-DG5200.yaml
tfoks: see above

tfoks:
Revision needed

PR #4175, opened 2026-05-13
add HPE S0M82A S0M84A S0M85A

aopdal:
Add three HPE Aruba switches


harryajc (changes requested):

aopdal:
The airflow on these switches is Front/Right/Left to Back so I took the mixed alternative. And its installed and not shipping weight.

harryajc (approved):

PR #4183, opened 2026-05-21
Update QFX5100-48S-6Q.yaml

arbitraryc0de:
The current template for the QFX5100-48S-6Q model both ge-* and xe-* interfaces for the same 48 physical SFP+ ports.

On this platform, the 48 SFP+ cages are physically shared ports that can operate at either 1G (ge-*) or 10G (xe-*) depending on optics/configuration, so defining both creates duplicate physical interfaces in NetBox.

The hardware layout should be:

xe-0/0/0 through xe-0/0/47 (48x SFP+)
et-0/0/48 through et-0/0/53 (6x QSFP+)

Using both ge-* and xe-* results in 102 data interfaces being created for a 54-port switch.

Unless we are expected to manually delete half the file, I would suggest:

keeping only the xe-* SFP+ interfaces
keeping the 6x et-* QSFP+ interfaces
removing the duplicate ge-* entries

https://github.com/netbox-community/devicetype-library/blob/master/device-types/Juniper/QFX5100-48S-6Q.yaml

Thanks for maintaining the library.


harryajc (approved):

harryajc:
@arbitraryc0de I think quite a few of the Juniper switches have duplicates in the same way, might need to have a look at that

PR #4196, opened 2026-05-24
new device type wth front and rear pictures added. 

cieska:


device-types/HPE/Aruba-2530-24G-PoEP-2SFPP.yaml
```diff
@@ -0,0 +1,122 @@
+---
+manufacturer: HPE
+model: Aruba 2530-24G-PoE+-2SFP+
+slug: hpe-aruba-2530-24g-poep-2sfpp
+part_number: J9854A
+u_height: 1
+is_full_depth: false
```
harryajc, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

PR #4198, opened 2026-05-27
Added Various Cisco UCS C Series images

mkarel:


harryajc:
@mkarel Thank you for your contribution, but in order for these images to be added to the associated devices within netbox you need to add the following lines to the device Yamls:
`front_image: true`
`rear_image: true`

mkarel:
I didn't realize that. I will look to address that.

~Matt Karel


Two roads diverged in a wood, and I--
I took the one less traveled by,
And that has made all the difference.
-Robert Frost (The Road Not Taken)

PR #4200, opened 2026-05-27
Adding Arista PWR module-types as per their vendor published product numbers

guidogz1:
Adding Arista PWR module-types as per their published product numbers. Reference https://www.arista.com/assets/data/pdf/Datasheets/7368X4-Datasheet.pdf and recent price listl


module-types/Arista/PWR-1900AC-R.yaml
```diff
@@ -0,0 +1,9 @@
+---
+manufacturer: Arista
+model: PWR-1900AC-R
+part_number: PWR-1900AC-R
+# airflow: rear-to-front
```
harryajc, line 5: These can be uncommented as modules support airflow

module-types/Arista/PWR-1900AC-F.yaml
```diff
@@ -0,0 +1,9 @@
+---
+manufacturer: Arista
+model: PWR-1900AC-F
+part_number: PWR-1900AC-F
+# airflow: front-to-rear
```
harryajc, line 5: These can be uncommented as modules support airflow

harryajc (changes requested):

PR #4202, opened 2026-05-27
Arista module type update

guidogz1:
Adding Arista PWR module-types as per their published product numbers.


guidogz1:
@harryajc - New MR submitted, includes airflow parameter.

harryajc (approved):

PR #4216, opened 2026-05-30
Adding PaloAlto PA-5xx series devices

pmsharp2:
Adding PA-510, PA-520, PA-540, PA-550, PA-540, PA-560


aaronaxvig:
Nice work on these.  I was verifying a few things vs. the docs and noticed that there are PA-545-POE and PA-555-POE models too.  If you don't get those added I will try to remember, sometime.

There are [3 new tray types too](https://docs.paloaltonetworks.com/hardware/pa-500-hardware-reference/pa-500-series-firewall-installation/install-the-pa-500-series-firewall-in-an-equipment-rack) which would be nice to have for these:

- PAN-1RU-4POST-RACK-10
- PAN-1RU-4POST-RACK-11
- PAN-1RU-4POST-RACK-12

Example of a PA-400 series tray [here](https://github.com/netbox-community/devicetype-library/blob/master/device-types/Palo%20Alto/PAN-PA-400-RACKTRAY.yaml).

Which reminds me--we should have `subdevice_role: child` set on all these 500 series devices, as in #1499 and #2529.

harryajc (approved):

PR #4220, opened 2026-06-03
add Ribbon Communications SBC5400

slauzon:


device-types/Ribbon Communications/SBC-5400.yaml
```diff
@@ -0,0 +1,57 @@
+---
+manufacturer: Ribbon
+model: SBC 5400
+slug: ribbon-communications-sbc-5400
+u_height: 2
+is_full_depth: true
```
harryajc, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

device-types/Ribbon Communications/SBC-5400.yaml
slauzon: weight and associated weight_unit have been added to PR

harryajc (approved):

PR #4228, opened 2026-06-05
Add PDU PremiumCord PDU-E10G10

sysadmin-pg:
Rack PDU
Power sockets: 10x IEC C13
Power cord:
- Plug type: CEE E/F 7/7
- Length: 2m

### DIFF
--- a/device-types/PDU-E10G10.yaml
+++ b/device-types/PDU-E10G10.yaml
@@ -1,0 +1,44 @@
+---
+manufacturer: PremiumCord
+model: Rack PDU 10x IEC C13, 2m cable
+slug: premiumcord-pdu-e10g10
+description: 10x C13, 16 A, 230 V, 50 Hz
+part_number: PDU-E10G10
+u_height: 1.0
+is_full_depth: false
+airflow: passive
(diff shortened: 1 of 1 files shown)


device-types/PDU-E10G10.yaml
```diff
@@ -0,0 +1,44 @@
+---
+manufacturer: PremiumCord
+model: Rack PDU 10x IEC C13, 2m cable
+slug: premiumcord-pdu-e10g10
+description: 10x C13, 16 A, 230 V, 50 Hz
+part_number: PDU-E10G10
+u_height: 1.0
+is_full_depth: false
```
harryajc, line 8: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

device-types/PDU-E10G10.yaml
sysadmin-pg: Hi Harry,
I cannot find this info from the official site of the manufacturer.
I have tried several e-shops that sell this PDU, but havemn't found it.
I cancel the merge request.

PR #4249, opened 2026-06-10
Add some Eaton PDUs, update Daisy Chain Ports

pmatura:
The Daisy Chain ports are interface ports that should be connected together.


harryajc (changes requested):
Surely these Daisy Chain ports are better modelled as console-ports?

pmatura:
The problem is that one PDU is supposed to have an "upstream" network port connected and a second PDU can be daisy-chained by then connecting the daisy-chain ports together, which is not allowed by the Netbox data model.

aaronaxvig:
In some Vertiv PDUs that I was just modeling, it was similar.  They have an IMD-5M module in a module bay, but interfaces could maybe done the same here.

I did `Port 0` and `Port 1` as the 1000BASE-T ports, and both of them connected to a bridge port `Bridge 0`.  This is pretty much exactly how they are laid out in the management interface.

I believe there are no examples of this in the devicetype-library.  Here is an export of what I built:

```
---
profile: null
manufacturer: Vertiv
model: IMD-5M
part_number: ''
description: ''
weight: null
weight_unit: null
airflow: null
attribute_data: null
comments: ''
console-ports:
- name: RS232
  type: rj-45
  label: ''
  description: ''
interfaces:
- name: Bridge 0
  type: bridge
  enabled: true
  mgmt_only: true
  label: ''
  description: ''
  bridge: null
  poe_mode: null
  poe_type: null
  rf_role: null
- name: Port 0
  type: 1000base-t
  enabled: true
  mgmt_only: true
  label: ''
  description: ''
  bridge: Bridge 0
  poe_mode: null
  poe_type: null
  rf_role: null
- name: Port 1
  type: 1000base-t
  enabled: true
  mgmt_only: true
  label: ''
  description: ''
  bridge: Bridge 0
  poe_mode: null
  poe_type: null
  rf_role: null
- name: Remote sensor
  type: other
  enabled: true
  mgmt_only: false
  label: ''
  description: ''
  bridge: null
  poe_mode: null
  poe_type: null
  rf_role: null

pmatura:
> I did `Port 0` and `Port 1` as the 1000BASE-T ports, and both of them connected to a bridge port `Bridge 0`. This is pretty much exactly how they are laid out in the management interface.

Sounds like a good idea. I added that to the definitions.

harryajc (approved):

PR #4256, opened 2026-06-10
HPE Alletra-6050: Missing ports, module slots

pmatura:
Move Power ports to PSU module slots, and FC ports to PCIe module slots


device-types/HPE/Alletra-6050.yaml
harryajc, line 1: For dual controller systems the best way to model them is the chassis has 2 `device-bays` then create a child device for the controller that has all of the interfaces/module-bays for the controller

harryajc (changes requested):

pmatura:
So, like this?

harryajc (approved):

pmatura:
Very nice. Thanks for all of your reviews @harryajc!

PR #4260, opened 2026-06-10
Add Avocent ACS48 variant

pmatura:


device-types/Avocent/ACS48.yaml
```diff
@@ -0,0 +1,114 @@
+---
+manufacturer: Avocent
+model: ACS48
+slug: avocent-acs48
+u_height: 1
```
harryajc, line 5: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

device-types/Avocent/ACS48.yaml
pmatura: I added weight information, but I only know the airflow is side-to-side. I can't confirm whether it's left-to-right or right-to-left.

harryajc (approved):

PR #4261, opened 2026-06-10
Add Cisco Catalyst 3750E-48TD variant and some missing information/ports

pmatura:


device-types/Cisco/WS-C3750E-48TD.yaml
```diff
@@ -0,0 +1,119 @@
+---
+manufacturer: Cisco
+model: Catalyst 3750E-48TD
+slug: cisco-ws-c3750e-48td
+part_number: WS-C3750E-48TD
+is_full_depth: false
```
harryajc, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

harryajc (approved):

PR #4264, opened 2026-06-10
Dell PowerSwitch S3048: Add missing power port

pmatura:


device-types/Dell/PowerSwitch-s3048-on.yaml
```diff
@@ -12,6 +12,8 @@ console-ports:
 power-ports:
   - name: PS1
     type: iec-60320-c14
+  - name: PS2
```
harryajc: By the looks these are hot swappable PSUs, therefore they should be modelled as `module-bays`

harryajc (changes requested):

pmatura:
I changed them to module slots, however I couldn't verify the specific model numbers for the PSU modules that go in there.

harryajc (approved):

PR #4265, opened 2026-06-10
Add Dell PowerVault MD1400, MD2412

pmatura:


device-types/Dell/PowerVault-MD2412.yaml
```diff
@@ -0,0 +1,31 @@
...
+  - name: EMM1 SAS Port 1
+    type: other
+  - name: EMM1 SAS Port 2
+    type: other
+  - name: EMM2 SAS Port 1
+    type: other
+  - name: EMM2 SAS Port 2
+    type: other
+power-ports:
+  - name: PSUA
```
harryajc: Are these Hot Swappable PSUs? If so they should be modelled as `module-bays`

device-types/Dell/PowerVault-MD1400.yaml
```diff
@@ -0,0 +1,41 @@
...
+  - name: EMM1 SAS Port 1
+    type: other
+  - name: EMM1 SAS Port 2
+    type: other
+  - name: EMM1 SAS Port 3
+    type: other
+  - name: EMM1 SAS Port 4
+    type: other
+power-ports:
+  - name: PS1
```
harryajc: Are these Hot Swappable PSUs? If so they should be modelled as `module-bays`

harryajc (changes requested):

pmatura:
Yes, you're right, I updated the definitions.

harryajc (approved):

PR #4270, opened 2026-06-12
Add devices

slauzon:


device-types/AudioCodes/mediant-1000.yaml
```diff
@@ -0,0 +1,41 @@
...
+part_number: M1KB
+u_height: 1
+is_full_depth: true
+weight: 3.3
+weight_unit: kg
+console-ports:
+  - name: Console
+    type: de-9
+power-ports:
+  - name: PSU1
```
harryajc: Are these hot swappable PSUs? if so they should be modelled as `module-bays`

device-types/Moxa/NPort-5610-16.yaml
```diff
@@ -0,0 +1,48 @@
+---
+manufacturer: Moxa
+model: NPort 5610-16
+slug: moxa-nport-5610-16
+u_height: 1
+is_full_depth: false
+weight: 2490
```
harryajc, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

1. Please add the `airflow` data point if you have it or can find it. Often these are listed in the datasheet, but sometimes they are not.

harryajc (changes requested):

device-types/AudioCodes/mediant-1000.yaml
slauzon: They are hot swappable.  I missed that on the design. I have also added airflow direction.  I have edited the file accordingly.

device-types/Moxa/NPort-5610-16.yaml
slauzon: Added airflow.

harryajc (approved):

PR #4271, opened 2026-06-14
Add Raritan/PX3TS-5191CR. Fix typo in Raritan/PX3TS-5184CR

nickhilliard:
Add Raritan/PX3TS-5191CR, which is the CE certified equivalent of the UL certified Raritan/PX3TS-5184CR model.

PX3TS-5184CR / 5191CR only has one C19 outlet, not two.


device-types/Raritan/PX3TS-5191CR.yaml
```diff
@@ -0,0 +1,56 @@
+---
+manufacturer: Raritan
+model: PX3TS-5191CR
+slug: raritan-px3ts-5191cr
+part_number: PX3TS-5191CR
+u_height: 1
```
harryajc, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

device-types/Raritan/PX3TS-5191CR.yaml
nickhilliard: no problem, now added + also for the PX3TS-5184CR.

harryajc (approved):

PR #4277, opened 2026-06-16
Add Opengear ACM5003-M-F-E

draggeta:
This PR is to add the Opengear ACM5003-M-F-E model with 3 ports, 1 ethernet and power.


device-types/Opengear/ACM5003-M-F-E.yaml
```diff
@@ -0,0 +1,21 @@
+---
+manufacturer: Opengear
+model: ACM5003-M-F-E
+slug: opengear-acm5003-m-f-e
+part_number: ACM5003-M-F-E
+u_height: 1
+is_full_depth: false
```
harryajc, line 10: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

device-types/Opengear/ACM5003-M-F-E.yaml
draggeta: I'll add those. Another question, this device has a built in modem and interface. Is there a way to model an RJ11 interface?

draggeta: I think I managed to fix everything

harryajc (approved):

PR #4282, opened 2026-06-16
Add Apple iPhone device types (iPhone X through 17, SE, Air, e)

vince6e74:
Adds 36 device-type definitions for Apple iPhones, from iPhone X up to the iPhone 17 line, including all Pro / Pro Max / Plus / mini variants as well as iPhone SE (1st–3rd generation), iPhone Air, and the 16e / 17e models.


harryajc:
@vince6e74  As per #2351 I don't think iPhones really fit in this repo

vince6e74:
@harryajc I’ve included iPhones because they’re managed via MDM.

PR #4295, opened 2026-06-19
updated

visezion:


harryajc:
This device has already been added in a different PR

PR #4306, opened 2026-06-24
Create AP36.yaml

tippenring:
I'm not sure if there is a preference between the -US or -WW (worldwide) versions. I'm adding the US version. 

The only difference between -US and -WW devices is the worldwide does not offer standard power. This relates to the comment "  \n- Low Power Indoor and Standard Power (US only)".


tippenring (commented):
formatting change

PR #4316, opened 2026-06-30
Add Ubiquiti UPS

MikiahG:
Add the Ubiquiti 2U UPS

First time PR


harryajc (changes requested):
For the images they should be cropped so that they have no white or empty space around the device, otherwise they won't fit properly in the rack view in netbox

harryajc (approved):

PR #4317, opened 2026-07-01
Add a few Lancom devices

FloEisen:
Added the following Lancom Device Types:

- LX-6400
- LX-6500
- OAP-1700B
- LN-860
- OX-6402
- GS-4554X
- 1900EF
- ISG-5000


harryajc (changes requested):

harryajc (approved):

PR #4323, opened 2026-07-01
Add Nvidia Jetson Orin Nano Super

tdiego1:
## Summary
- Adds device type definition for the NVIDIA Jetson Orin Nano Super Developer Kit, sourced from the official NVIDIA datasheet.
- Includes `eth0` (1000base-t) and `wlan0` (802.11ac) interfaces, 4x USB-A, 1x USB-C, and 1x DisplayPort front-ports, and a DC-terminal power port.

## Test Plan
- [x] YAML passes pre-commit checks (trailing-whitespace, end-of-file, check-yaml, yamlfmt, yamllint)
- [x] PyTest case checks passed

### DIFF
--- a/device-types/Nvidia/Jetson-Orin-Nano-Super.yaml
+++ b/device-types/Nvidia/Jetson-Orin-Nano-Super.yaml
@@ -1,0 +1,32 @@
+---
+manufacturer: Nvidia
+model: Jetson Orin Nano Super
+slug: nvidia-jetson-orin-nano-super
+part_number: 945-137766-0000-000
+u_height: 0
+is_full_depth: false
+comments: >-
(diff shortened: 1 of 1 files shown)


device-types/Nvidia/Jetson-Orin-Nano-Super.yaml
```diff
@@ -0,0 +1,20 @@
+---
+manufacturer: Nvidia
+model: Jetson Orin Nano Super
+slug: nvidia-jetson-orin-nano-super
+part_number: 945-137766-0000-000
+u_height: 0
```
harryajc, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

PR #4331, opened 2026-07-04
Add device definitions

potvinp:
- Add HPE Plexxi PX-S3EQ-F and PX-S3EQ-R definitions
- Add Celestica DX010 definition

### DIFF
--- a/device-types/Celestica/DX010.yaml
+++ b/device-types/Celestica/DX010.yaml
@@ -1,0 +1,88 @@
+---
+manufacturer: Celestica
+model: DX010
+slug: celestica-dx010
+part_number: DX010
+u_height: 1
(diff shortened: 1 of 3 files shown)


device-types/Celestica/DX010.yaml
```diff
@@ -0,0 +1,88 @@
+---
+manufacturer: Celestica
+model: DX010
+slug: celestica-dx010
+part_number: DX010
+u_height: 1
+airflow: front-to-rear
```
harryajc, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

PR #4340, opened 2026-07-09
Adds Philips Hue Bridge

katelovescode:
Philips Hue Bridge 2.1

### DIFF
--- a/device-types/Philips/hue-bridge-2-1.yaml
+++ b/device-types/Philips/hue-bridge-2-1.yaml
@@ -1,0 +1,17 @@
+---
+manufacturer: Philips
+model: Hue Bridge 2.1
+slug: philips-hue-bridge-2-1
+part_number: 3241312018A
+u_height: 0
+is_full_depth: false
+weight: 210
+weight_unit: g
+comments: See [Product documentation](https://www.philips-hue.com/en-us/support).
+power-ports:
+  - name: USB Power
+    type: usb-micro-b
+    maximum_draw: 5
+interfaces:
+  - name: Ethernet
+    type: 1000base-t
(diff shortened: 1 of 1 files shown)


katelovescode:
Updated slug to `philips-hue-bridge-2-1` (dot → hyphen), consistent with the existing pattern for devices with dotted version numbers (e.g. `tp-link-cpe210-v3-20`, `thomas-krenn-ra1112-asepn-v2-0`).

Note: the CI `test` check will still fail for the same reason it fails on my other open PR (#4341). There are two bugs in `tests/device_types.py` that create an impossible constraint for new devices:

1. `_slugify_manufacturer()` doesn't replace dots, so manufacturers like `GL.iNet` produce a prefix of `gl.inet-`
2. `_slugify_model()` doesn't replace dots either, so models like `Hue Bridge 2.1` produce an expected suffix of `hue-bridge-2.1`

Both conflict with the slug schema (`^[-a-z0-9_]+$`) which forbids dots. Existing devices with dotted versions (TP-Link, Thomas-Krenn, etc.) pass only because their slugs are grandfathered in `known-slugs.json` and skip the manufacturer/model checks.

The fix would be to add `.replace(".", "")` or `.replace(".", "-")` to both `_slugify_manufacturer()` and `_slugify_model()`.

device-types/Philips/hue-bridge-2-1.yaml
```diff
@@ -0,0 +1,15 @@
+---
+manufacturer: Philips
+model: Hue Bridge 2.1
+slug: philips-hue-bridge-2-1
+part_number: 3241312018A
+u_height: 0
```
harryajc, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

katelovescode:
Ready for re-review, failed build notwithstanding

danner26:
@katelovescode the reason this is not running is because you edited the test file as well. Please remove that chance from this PR, and it should run the automated tests.

harryajc (approved):

PR #4341, opened 2026-07-09
Adds GLiNet mini router

katelovescode:
GLiNet mini router (my use case is Basement Defender hub)

**Note on CI failure:** The `verify_slug` test in `tests/device_types.py` derives the expected manufacturer prefix via `_slugify_manufacturer()`, which does not replace dots. For `GL.iNet` this produces `gl.inet-`, but the slug schema (`^[-a-z0-9_]+$`) forbids dots — making it impossible to satisfy both constraints simultaneously for new GL.iNet devices.

Slug `gl-inet-gl-mt300n-v2` is consistent with all existing GL.iNet devices in the repo (which pass only because they are grandfathered into `known-slugs.json` and skip the manufacturer prefix check). The fix would be to add dot handling to `_slugify_manufacturer()` (e.g. `.replace(".", "")`).


device-types/GL-iNet/GL-MT300N-V2.yaml
```diff
@@ -0,0 +1,19 @@
+---
+manufacturer: GL.iNet
+model: GL-MT300N-V2
+slug: gl-inet-gl-mt300n-v2
+part_number: GL-MT300N-V2
+u_height: 0
```
harryajc, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

katelovescode:
Ready for re-review, failed build notwithstanding

tests/device_types.py
danner26, line 1: Please remove the edit to this file, and the automated tests will run correctly.

danner26 (dismissed):

harryajc (approved):

PR #4342, opened 2026-07-09
Adds GMKtec NucBox

katelovescode:
GMKtec NucBox G3 Plus


device-types/GMKtec/NucBox-G3-Plus.yaml
```diff
@@ -0,0 +1,17 @@
+---
+manufacturer: GMKtec
+model: NucBox G3 Plus
+slug: gmktec-nucbox-g3-plus
+part_number: G3 PLUS
+u_height: 0
```
harryajc, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

katelovescode:
Ready for re-review

danner26 (approved):

harryajc (approved):

PR #4345, opened 2026-07-09
Fix slugify functions to replace dots, aligning with slug schema pattern

### LINKED ISSUE #4343: Bug: _slugify_manufacturer and _slugify_model don't replace dots, conflicting with slug schema pattern
## Description

I wanted to flag what appears to be a bug that creates an impossible validation constraint for new device types from manufacturers whose names or model numbers contain dots.

## The conflict

The slug schema (`schema/devicetype.json`) requires:

```json
"pattern": "^[-a-z0-9_]+$"
```

Dots are not permitted. However, `_slugify_manufacturer()` and `_slugify_model()` in `tests/device_types.py` do not replace dots:

```python
def _slugify_manufacturer(self):
    return self.manufacturer.casefold().replace(" ", "-").replace("sfp+", "sfpp")...
    # no .replace(".", ...) — dots are preserved

def _slugify_model(self):
    slugified = self.model.casefold().replace(" ", "-").replace("sfp+", "sfpp")...
    # no .replace(".", ...) — dots are preserved
```

This means:
- A manufacturer like `GL.iNet` produces expected prefix `gl.inet-`
- A model like `Hue Bridge 2.1` produces expected suffix `hue-bridge-2.1`

Both contain dots, which the schema forbids. New devices with dots in their manufacturer name or version number cannot satisfy both `verify_slug` and the schema simultaneously.

## Suggested fix

Adding dot handling to both slugify functions would resolve the conflict:

```python
def _slugify_manufacturer(self):
    return self.manufacturer.casefold().replace(" ", "-").replace(".", "")...

def _slugify_model(self):
    slugified = self.model.casefold().replace(" ", "-").replace(".", "")...
```

(Whether dots should become hyphens or be removed entirely is a style call — the existing `known-slugs.json` entries use hyphens, e.g. `v3-20` for `v3.20`.)

## Affected PRs

I ran into this on two open PRs of mine: #4340 and #4341.

katelovescode:
## Summary

The three slugify functions in `tests/device_types.py` — `_slugify_manufacturer`, `_slugify_model`, and `_slugify_part_number` — did not replace dots. The slug schema pattern (`^[-a-z0-9_]+$`) does not permit dots, so new devices from manufacturers or models with dots in their names had no valid slug that could satisfy both the schema validation and the `verify_slug` prefix/suffix checks simultaneously.

**Example:** A new GL.iNet device would require a slug starting with `gl.inet-` (per `verify_slug`) but that dot fails schema validation. Similarly, `Hue Bridge 2.1` produced an expected suffix of `hue-bridge-2.1`.

Existing affected devices are unaffected — their slugs are already in `known-slugs.json` and skip the prefix/suffix checks.

## Changes

- `_slugify_manufacturer`: added `.replace(".", "-")`
- `_slugify_model`: added `.replace(".", "-")`
- `_slugify_part_number`: added `.replace(".", "-")` for consistency

This matches the pattern already established by existing grandfathered slugs (e.g. `tp-link-cpe210-v3-20` for model `CPE210 v3.20`).

Closes #4343


danner26:
Thank you for creating this PR.

The slugify fix looks right to me and implementation choice for hyphens matches the existing slugs format.

One small issue: the same `_slug_model` and `_slug_part_number` values also get used by `verify_filename`, not just `verify_slug`. To that extent, `verify_filename` does not skip known devices the way the prefix/suffix checks do. So, the note about existing devices being unaffected isnt accurate.

I reviewed the current files against the patched code and there are 7 `device-types` filenames that still have dots, and thus will fail validation once `_slug_model` uses hyphens.

- device-types/TP-Link/CPE210-v1.1.yaml
- device-types/TP-Link/CPE210-v3.0.yaml
- device-types/TP-Link/CPE210-v3.20.yaml
- device-types/Thomas-Krenn/oss-capacity-v3.0.yml
- device-types/Thomas-Krenn/ra1112-asepn-v2.0.yml
- device-types/Thomas-Krenn/ri1208-smxs-v1.0.yml
- device-types/Uptime Industries/Bladerunner-1.0-20-blade.yaml

They don't show up in this PR's CI because it only validates changed files, but a full run (`pre-commit run -al`) catches them. They'll trip up any future PR that touches those files or when a maintainer/contributor runs a full check.

The fix is just to rename them to use the hyphen form. The slugs already use hyphens and stay in known-slugs.json, and images are matched by slug not filename, so the renames are safe.

I also wanted to mention that `Most/A10-1.6.yaml` and `Sysmatrix/PD-0802.000.yaml` are okay in their current form since their dotted value is in `part_number`, which `verify_filename` matches directly.

danner26:
@katelovescode please either make these updates yourself, or provide me write access to your repo and I can make them for you. Thank you.

danner26:
@katelovescode I apologize in advance and hope you didn't spend any time on this yet. Following up on my own suggestion above, I tested the rename approach in a codespace and it does not work as I originally expected it to. Scratch that part.. And please accept my apologies again.

When I rename the files our CI fails, but not for the reason I expected. Git does detect them as renames (`change_type` `R`), but the test method (`_get_diff_from_upstream`) resolves each rename to the old path and tries to open it. Since the file has been renamed, that path no longer exists and we get `FileNotFoundError` on all 7. The harness is grabbing the wrong side of the rename diff.

So realistically we need to also update the test method, which is a secondary issue that I do not believe should be contained within this PR.

So that all being said, please give me a little bit of time to update the test harness and resolve this issue. After doing so, we can update your branch and resolve the file names. You may need to do some merge conflict fixing, but we shall see. I will get back to you shortly, thanks.

danner26:
@katelovescode I really appreciate this. I folded your code in to a PR and fixed the subsequent files. Thank you again

katelovescode:
No worries at all, I had no time to look at it today, thanks for carrying it across the finish line!  I don't have a ton of devices in my home lab but wanted to contribute back where I could.

danner26:
> No worries at all, I had no time to look at it today, thanks for carrying it across the finish line! I don't have a ton of devices in my home lab but wanted to contribute back where I could.

That is much appreciated! Thank you.

PR #4349, opened 2026-07-11
Add new device (Gowin 1U Rack Mount Router (GW-BS-1UR2-25G))

jetsure12:
Add Gowin 1U Rack Mount Router (GW-BS-1UR2-25G)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,8 +26,14 @@ When writing new definitions, there are some important guidelines to follow:
   different).
 - Use the complete form of interface names where applicable. For example, use `TenGigabitEthernet1/2/3` instead of
 `Te1/2/3`.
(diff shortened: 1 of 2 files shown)


jetsure12 (commented):
First line is correct and the last line is ended ith a type as in every other file

device-types/Gowin/GW-BS-1UR2-25G.yaml
```diff
@@ -0,0 +1,43 @@
+---
+manufacturer: GoWin
+model: GW-BS-1UR2-25G
+slug: gowin-gw-bs-1ur2-25g
+part_number: GW-BS-1UR2-25G
+u_height: 1
+is_full_depth: false
```
harryajc: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

PR #4352, opened 2026-07-13
New device WS-C2960X-24PSQ-L

nuskad:
Added WS-C2960X-24PSQ-L


nuskad:
The switch model WS-C2960X-24PSQ-L is a bit of a strange 2960-X model
As it does only have the first 8 ports available for PoE, as the power budget is only 110W, as it is passively cooled.
Link to the [datasheet](https://www.cisco.com/c/en/us/products/collateral/switches/catalyst-2960-x-series-switches/datasheet_c78-728232.html)

harryajc (approved):

PR #4358, opened 2026-07-14
Add and fix Panduit fiber module trays

aaronaxvig:


harryajc (changes requested):
Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add to theses devices if possible.

aaronaxvig:
All are passive airflow so I added that.

I added weights for the ones that I could fine.

harryajc (approved):

PR #4359, opened 2026-07-14
Add Panduit Fiber Adapter Panels

aaronaxvig:
This is mostly the current series, ending in Z.

A few existed already but were not up to current standards so I redid those.


device-types/Panduit/FRME3.yaml
```diff
@@ -0,0 +1,28 @@
+---
+manufacturer: Panduit
+model: Opticom Rack Mount Fiber Enclosure, 3 RU
+slug: panduit-frme3
+part_number: FRME3
+u_height: 3
```
harryajc, line 6: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

device-types/Panduit/FRME3.yaml
aaronaxvig: I moved that file (with airflow and weight added) to my other pull request with other similar enclosures.

harryajc (approved):

PR #4370, opened 2026-07-21
Add Lenovo ThinkSystem DG5200 and supporting module-types

tfoks:
feat: add ThinkSystem DG5200 and supporting module-types


tfoks:
Hi @harryajc,
The tests failed for the DG5200 definition because it does nct define any power source:
`device-types/Lenovo/ThinkSystem-DG5200.yaml has does not appear to have a valid power source`
How to deal with this? The chassis in fact has no power source. Power is provided by the controller modules which are inserted into the module bays.

BTW: Netbox is completely fine with this definition and does not complain any missing power source.

device-types/Lenovo/ThinkSystem-DG5200.yaml
```diff
@@ -0,0 +1,17 @@
+---
+manufacturer: Lenovo
+model: ThinkSystem DG5200
+slug: lenovo-thinksystem-dg5200
+part_number: 7DHY
+u_height: 2
+is_full_depth: true
```
harryajc, line 7: To bypass the power check you can use `is_powered: false`

module-types/Lenovo/C4A5.yaml
harryajc, line 1: This controller should be modelled as a `device-type` given that the chassis has device bays not module bays

harryajc (changes requested):

device-types/Lenovo/ThinkSystem-DG5200.yaml
tfoks: fixed

module-types/Lenovo/C4A5.yaml
tfoks: My fault. That was the wrong version of the definition and should have module-bays. Fixed as well.

harryajc (approved):

PR #4375, opened 2026-07-23
Add Infoblox TE-1506 appliance devicetype

markturczan:


device-types/Infoblox/TE-1506.yaml
```diff
@@ -0,0 +1,50 @@
...
+subdevice_role: parent
+airflow: front-to-rear
+weight: 20.0
+weight_unit: lb
+console-ports:
+  - name: Console Port
+    type: de-9
+    description: Male DE-9 serial port
+power-ports:
+  - name: PSU1
```
harryajc, line 17: As these are hot swappable PSUs they should be modelled as `module-bays` with a separate module created for the PSUs if required.

device-types/Infoblox/TE-1506.yaml
```diff
@@ -0,0 +1,50 @@
...
+    maximum_draw: 475
+    allocated_draw: 375
+interfaces:
+  - name: HA Port
+    type: 1000base-t
+    enabled: true
+    mgmt_only: false
+    description: RJ-45 10/100/1000 Base-T auto-sensing gigabit Ethernet port through which the active node in an HA pair connects to the network.
+  - name: iLO/IPMI Port
+    type: other
```
harryajc: should this be 1000base-t?

harryajc (changes requested):

device-types/Infoblox/TE-1506.yaml
markturczan: Thank you for the review. You make a good point. The PSUs are field-replaceable; I'll make the recommended modifications and resubmit.

I should also add the DC version of the PSU module, as it is an available option.

device-types/Infoblox/TE-1506.yaml
markturczan: I wasn't sure how to type the IPMI port. The vendor documentation does not list the IPMI port as one of its "Network Ports", and I wasn't sure if using type 1000base-t would make it confusing. All of the other ports are described as "RJ-45 10/100/1000 Base-T auto-sensing gigabit Ethernet ports," while the IPMI is described as "RJ-45 10/100/1000 iLO/IPMI port". 

If it makes more sense for Netbox to type the interface as 1000base-t, I'll make the modification.

https://docs.infoblox.com/space/te1506/314966027/Infoblox+TE-1506+Series+Hardware+Components

harryajc (approved):

PR #4387, opened 2026-07-26
Create P550.yaml

tanfwc:


device-types/Yeastar/P550.yaml
```diff
@@ -0,0 +1,48 @@
...
+  - name: WAN
+    type: 1000base-t
+    mgmt_only: false
+    description: Wide Area Network / Primary Uplink
+  - name: LAN
+    type: 1000base-t
+    mgmt_only: false
+    description: Local Area Network
+  - name: Port 1
+    type: other
```
harryajc, line 22: Not sure if it fits but is `e1` a better type fit for this type of port?

harryajc (changes requested):

device-types/Yeastar/P550.yaml
tanfwc: @harryajc those port are actually RJ11, i have choose others as it does not fit as e1 port type

harryajc (approved):

PR #4398, opened 2026-07-31
Add Sophos XG 210

sysadmin-pg:
Entry-level unified threat management (UTM) appliance
designed for small office and branch office deployments.


device-types/Sophos/XG-210.yaml
```diff
@@ -0,0 +1,55 @@
...
+  - name: Gig-E 5
+    type: 1000base-t
+  - name: Gig-E 6
+    type: 1000base-t
+  - name: SFP(1)
+    type: 1000base-x-sfp
+  - name: SFP(2)
+    type: 1000base-x-sfp
+module-bays:
+  - name: PSU-EXT
```
harryajc: If this is a connector then it should be a power port with probably something like `type: dc-terminal`

harryajc (changes requested):

harryajc (approved):

PR #4404, opened 2026-08-01
TP-Link Omada SG2210XMP-M2 Smart Switch

TwentySixer:
Definition for TP-Link Omada SG2210XMP-M2 Smart Switch


device-types/TP-Link/SG2210XMP-M2.yaml
```diff
@@ -0,0 +1,61 @@
+---
+manufacturer: TP-Link
+model: SG2210XMP-M2
+slug: tp-link-sg2210xmp-m2
+part_number: SG2210XMP-M2
+u_height: 1
+is_full_depth: false
```
harryajc, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

harryajc (approved):

PR #4421, opened 2026-08-08
Add Crestron TSS-770 device-type and TSW-770-LB Module

clacyNS:
Added the manufacturer Crestron and then added a device-type and a module related to the device.


device-types/Crestron/tss-770.yaml
```diff
@@ -0,0 +1,34 @@
...
+manufacturer: Crestron
+model: TSS-770
+slug: crestron-tss-770
+part_number: TSS-770
+u_height: 0
+is_full_depth: false
+subdevice_role: parent
+airflow: passive
+weight: 14.10
+weight_unit: oz
```
harryajc, line 11: To link to image to the device you need to add `front_image: true` you will need to update the image name to match the current schema <VALUE_IN_SLUG>.front.<acceptable_format>

device-types/Crestron/tss-770.yaml
```diff
@@ -0,0 +1,34 @@
...
+  Wi-Fi 802.11a/b/g/n/ac, Bluetooth beacon. Not rack-mountable.
+
+interfaces:
+  - name: Ethernet
+    type: 100base-tx
+    mgmt_only: false
+    poe_mode: pd
+    poe_type: type2-ieee802.3at
+
+  - name: USB1
```
harryajc, line 24: I think given how netbox models things the best way might be to model these usb ports as `console-server-ports` and then model the ones on the other device as `console-ports`

harryajc (changes requested):

PR #4454, opened 2026-08-11
Add Ubiquiti UniFi U6 Long-Range device type

visezion:


harryajc:
This device already exists in this repo

PR #4480, opened 2026-08-19
UPDATED

bbrk364:


bbrk364:
PATCH

walday:
This represents a massive, impressive body of work spanning three years, but it requires thorough verification:
1) A large number of files have been renamed—following a migration strategy from underscores to dashes—resulting in duplicates (e.g., `3_com` vs. `3-com` equipment).
2) Some of the proposed changes involve merely adding empty comments or decomposing existing entries (e.g., `device-types/ADVA/FSP-150-XG480-25G-100G.yaml`).
3) Certain hardware items (e.g., APC ATS, Cisco ISR) have been entered incorrectly regarding the manufacturer's model designation and effectively duplicate existing devices.
I suggest breaking this commit down into smaller parts and taking a careful, considered approach to merging such a large dataset.

danner26:
As @walday please break this commit down into smaller chunks to make it easier to review. Thank you @bbrk364

PR #4488, opened 2026-08-21
Add 72 AXIS IP camera device types

ch-bas:
Adds **72 AXIS IP network cameras** to the existing `device-types/AXIS/` directory (M, P, Q, and F/FA series — bullets, domes, box, panoramic, thermal, and PTZ). Companion to my Hikvision (#4485), Dahua (#4486), and Uniview (#4487) PRs.

### Included per device
- **PoE interface** — `type` from the datasheet's `RJ45 …BASE-T` connector row (`100base-tx`, or `1000base-t` for the 13 Gigabit models), `poe_mode: pd` + `poe_type` from the datasheet's stated 802.3 standard (af/at/bt).
- **DC power port** (dc-terminal, `maximum_draw`) where the camera supports DC input.
- **microSD module bay** where present.
- **weight**, and a datasheet link in `comments`.

### Sourcing
All specs come from the official Axis datasheet linked in each file's `comments`, and fields the datasheet doesn't state are **omitted rather than estimated**.

### Validation
`pytest tests/definitions_test.py`, `yamllint`, and the `yamlfmt` pre-commit hook all pass for the 72 files. The three existing `device-types/AXIS/` entries are untouched, and no slug/filename collides with them.

### Disclosure
Generated from [cctv-database.com](https://cctv-database.com), a CC0 camera-spec database I maintain where every entry is manually verified against the manufacturer datasheet. Happy to adjust naming/fields to match the library's conventions.


harryajc (changes requested):

ch-bas:
Thanks for the review, @harryajc! I've added `weight`/`weight_unit` to all 68 AXIS files that were missing it — every value is the camera-only weight taken from the official Axis datasheet for that model (grams; kg figures converted). Where a datasheet lists weight only "with weathershield" (the integral outdoor LVE/VE domes and bullets), I used that figure since the shield isn't a removable mount. Nothing was estimated.

I also merged the latest `master` in to clear the earlier `validate-inputs` failure (the branch had fallen behind), so all checks are green now.

All 72 AXIS files now carry a weight. Ready for another look — thanks!

harryajc (approved):

PR #4493, opened 2026-08-24
Adding Synology RS3618XS

samsclub91:


.github/workflows/sync-slim.yml
harryajc, line 1: What is the reason for this workflow file?

tests/known-modules.json
harryajc, line 1: These json files should not be modified during a PR

harryajc (changes requested):

PR #4498, opened 2026-08-27
hpe-msa-2050-san-controller elevation images

Domoninic:


device-types/HPE/MSA-2050-SAN-Controller.yaml
harryajc, line 1: You have linked these images to the controller where they should be linked to the actual chassis which is "MSA-2050-Enclosure.yaml"

harryajc (changes requested):

harryajc (approved):

PR #4505, opened 2026-08-28
Cisco ucs c200 m2

edaleand:


device-types/Cisco/UCS-C200-M2.yaml
```diff
@@ -0,0 +1,22 @@
+---
+manufacturer: Cisco
+model: UCS-C200-M2
+slug: cisco-ucs-c200-m2
+part_number: UCS-C200M2
+u_height: 1
+is_full_depth: true
```
harryajc, line 10: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

device-types/Cisco/UCS-C200-M2.yaml
edaleand: I have added the data.

harryajc (approved):

PR #4507, opened 2026-08-28
Add Nvidia Q3400-RA, SN5610 and NVLink 5 switch; complete the QM9700 port list

dietergorecki:
Adds three NVIDIA switch definitions and fixes an existing one. Component lists were verified against running hardware and cross-checked with vendor datasheets; anything neither source states has been omitted rather than guessed.

### New definitions

- **Q3400-RA** — Quantum-X800, 4U, 144x 800Gb/s XDR over 72 twin-port OSFP cages. Port names follow NVOS (`swNpM`). `fnm1` is the dedicated in-band fabric-node management port used by UFM. 8x C20 PSUs at 875W. PN `920-9B36F-00RX-8S0`.
- **SN5610** — Spectrum-4, 2U, 64x 800GbE OSFP. Power-optimised variant of the SN5600, with 4 PSU bays per the HPE SKU. Breakout sub-ports are deliberately **not** modelled, as splitting is a deployment choice rather than fixed hardware.
- **NVLink 5 Switch** — the NVL72 NVLink switch tray (9 per rack). A deliberately thin definition: only the out-of-band management ports are modelled, because the NVLink ports are blind-mate backplane connections to the compute trays rather than field-cabled ports. Fed by the rack busbar, hence `is_powered: false`. **Happy to drop this one if you would rather not carry a definition with no data ports.**

### Fix to the existing QM9700

The current `Nvidia/QM9700.yaml` has a few problems, which this PR corrects:

1. **Two ports are missing.** It defines 62 interfaces, but the QM9700 has 64 (32 twin-port OSFP cages x 2). `P22.1` and `P23.0` are absent.
2. **No management interface** is defined.
3. **Ports are typed as Ethernet** (`400gbase-x-osfp`) on a switch that is InfiniBand-only.
4. **Port names do not match the OS.** MLNX-OS presents the ports as `IB1/<cage>/<port>`, not `P<n>.<n>`.

The replacement defines all 64 ports as `infiniband-ndr` named `IB1/1/1` through `IB1/32/2`, plus `mgmt0`, and keeps the existing chassis attributes, PSU bays and console port. Since this renames interfaces, it is a breaking change for anyone who has already instantiated the old definition — happy to split it into its own PR, or drop it, if you would prefer.

Note: I had also prepared an SN5600 definition but it turned out to be functionally identical to the one already in the library, so it is not included here.

### Validation

`yamllint` and `pytest tests/definitions_test.py` both pass locally against current `master`.

### DIFF
--- a/device-types/Nvidia/NVLink-5-Switch.yaml
+++ b/device-types/Nvidia/NVLink-5-Switch.yaml
@@ -1,0 +1,21 @@
+---
+manufacturer: Nvidia
+model: NVLink 5 Switch
+slug: nvidia-nvlink-5-switch
+u_height: 1
+is_full_depth: true
+comments: |
+  NVIDIA fifth-generation NVLink switch tray, as deployed in GB200/GB300 NVL72
+  racks (9 trays per rack). Only the out-of-band management interfaces are
+  modelled: the NVLink ports themselves are backplane cartridge connections to
+  the compute trays rather than field-cabled ports.
+  The tray is fed by the NVL72 rack busbar and has no power inlet of its own.
+is_powered: false
+interfaces:
+  - name: bmc
+    type: 1000base-t
+    mgmt_only: true
+  - name: eth0
+    type: 1000base-t
+  - name: eth1
+    type: 1000base-t
--- a/device-types/Nvidia/Q3400-RA.yaml
+++ b/device-types/Nvidia/Q3400-RA.yaml
@@ -1,0 +1,341 @@
+---
+manufacturer: Nvidia
+model: Q3400-RA
+slug: nvidia-q3400-ra
+part_number: 920-9B36F-00RX-8S0
+u_height: 4
+is_full_depth: true
+airflow: front-to-rear
+weight: 60
+weight_unit: kg
+comments: |
+  NVIDIA Quantum-X800 Q3400-RA. 144x 800Gb/s XDR InfiniBand ports over 72 twin-port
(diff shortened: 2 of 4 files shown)


device-types/Nvidia/Q3400-RA.yaml
```diff
@@ -0,0 +1,341 @@
...
+  NVIDIA Quantum-X800 Q3400-RA. 144x 800Gb/s XDR InfiniBand ports over 72 twin-port
+  OSFP cages, powered by the Quantum-3 ASIC. fnm1 is the dedicated in-band fabric
+  node management port used by NVIDIA UFM.
+  Product page: https://www.nvidia.com/en-us/networking/products/infiniband/quantum-x800/
+  User manual: https://networking-docs.nvidia.com/xdrswitcheshw/introduction
+console-ports:
+  - name: uart
+    type: rj-45
+power-ports:
+  - name: PSU1
```
harryajc, line 21: Are these FRUs? If so they should be modelled as `module-bays`

device-types/Nvidia/NVLink-5-Switch.yaml
```diff
@@ -0,0 +1,21 @@
+---
+manufacturer: Nvidia
+model: NVLink 5 Switch
+slug: nvidia-nvlink-5-switch
+u_height: 1
+is_full_depth: true
+comments: |
```
harryajc, line 7: If this is a tray within a larger device, then this should be modelled as a child device.

harryajc (changes requested):

PR #4508, opened 2026-08-28
Add Supermicro ARS-121GL-NB3-LCC-R1 GB300 NVL72 compute tray

dietergorecki:
Adds the liquid-cooled 1U GB300 compute tray used in the Supermicro `SRS-GB300-NVL72` rack (18 trays per rack).

Each tray carries 2x NVIDIA GB300 Grace Blackwell Superchips (2x 72-core Grace CPU, 4x B300 GPU), 4x ConnectX-8 800Gb/s HCAs and 1x BlueField-3 DPU.

### Notes

- The nine interfaces were verified against running hardware and are consistent across every tray checked.
- `is_powered: false` — the tray draws from the rack power shelves over the busbar and has no power inlet of its own.
- Interface names are as the OS presents them (`ibp3s0`, `ibP2p3s0`, `ibP16p3s0`, `ibP18p3s0` for the ConnectX-8 XDR HCAs; `enP22s22f0np0`/`f1np1` for the BlueField-3 ports).
- Weight is 32 kg; Supermicro does not publish a per-tray weight in the rack datasheet.

Data sheet: https://www.supermicro.com/datasheet/datasheet_SuperCluster_GB300_NVL72.pdf

### Validation

`yamllint` and `pytest tests/definitions_test.py` both pass locally.

### DIFF
--- a/device-types/Supermicro/ARS-121GL-NB3-LCC-R1.yaml
+++ b/device-types/Supermicro/ARS-121GL-NB3-LCC-R1.yaml
@@ -1,0 +1,46 @@
+---
+manufacturer: Supermicro
+model: ARS-121GL-NB3-LCC-R1
+slug: supermicro-ars-121gl-nb3-lcc-r1
+part_number: ARS-121GL-NB3-LCC-R1
+u_height: 1
+is_full_depth: true
+weight: 32
+weight_unit: kg
+comments: |
+  Liquid-cooled 1U GB300 compute tray for the Supermicro SRS-GB300-NVL72 rack
+  (18 trays per rack). Each tray carries 2x NVIDIA GB300 Grace Blackwell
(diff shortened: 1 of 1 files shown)


device-types/Supermicro/ARS-121GL-NB3-LCC-R1.yaml
```diff
@@ -0,0 +1,46 @@
...
+manufacturer: Supermicro
+model: ARS-121GL-NB3-LCC-R1
+slug: supermicro-ars-121gl-nb3-lcc-r1
+part_number: ARS-121GL-NB3-LCC-R1
+u_height: 1
+is_full_depth: true
+weight: 32
+weight_unit: kg
+comments: |
+  Liquid-cooled 1U GB300 compute tray for the Supermicro SRS-GB300-NVL72 rack
```
harryajc, line 11: If this goes inside a larger device, then this should probably be modelled as a child device

harryajc (changes requested):

PR #4527, opened 2026-08-29
Add support for new port mapping schema and migrated all current devices

### LINKED ISSUE #4350: All device types missing mapping between front and rear ports
Hi,

i'm not sure this is working as intended but when importing (which showed no errors), every device i looked at in a current netbox (currently using `v4.6.0-Docker-5.0.1`) with the `deviceypte-library` is missing the mapping between front and backend ports, which is kinda cumbersome to reconnect everything by hand.

every patch panel, every UTP plug, everything, means all device are shown as unconnected as long as one does not fix the 1:1 connection from front to backend ports.

screenshot for reference from an generic 4 UTP plug as imported

Image

in contrast screenshot of a fixed up generic 2 UTP plug as used

Image

As said, i'm not sure if it's intended to be this way, but as i found the rear ports specified in the `.yaml` files like in https://github.com/netbox-community/devicetype-library/blob/HEAD/device-types/Generic/wall-box-2-utp-plugs.yaml and it gave no error at import time i'd thought i should report it.

And btw, MANY thx for this repo, saved us an enourmous amount of time to get going and might also be the biggst reason AI is already pretty good at just outputting device type yamls for devices currently not included :)!

harryajc:


harryajc:
Including commits from #4194 this fixes all files that have front or rear ports to make sure that there won't be any issues with importing these devices

harryajc:
Merged under different PR

PR #4551, opened 2026-09-03
Add hardened Arista CCS-710 series

colinmacgiolla:


device-types/Arista/CCS-710HXP-20TNH-4S.yaml
```diff
@@ -0,0 +1,112 @@
...
+model: CCS-710HXP-20TNH-4S
+slug: arista-ccs-710hxp-20tnh-4s
+part_number: CCS-710HXP-20TNH-4S
+u_height: 0
+weight: 4.2
+weight_unit: kg
+is_full_depth: false
+airflow: passive
+comments: '[Arista CCS-710HXP Data Sheet](https://www.arista.com/assets/data/pdf/Datasheets/CCS-710HXP-Datasheet.pdf)'
+module-bays:
```
harryajc, line 12: Looking at the image this seems to be dc-terminal style power ports rather than module bays given that there is no slot for a power supply.

harryajc (changes requested):

harryajc (approved):

PR #4557, opened 2026-09-04
Create IR1101-K9.yaml

bezops:
Cisco Catalyst IR1101 Rugged Series Router

### DIFF
--- a/device-types/Cisco/IR1101-K9.yaml
+++ b/device-types/Cisco/IR1101-K9.yaml
@@ -28,6 +28,8 @@ interfaces:
     description: Cellular interface
 module-bays:
   - name: Expansion Module
+    position: '1'
+    description: Expansion module slot
 power-ports:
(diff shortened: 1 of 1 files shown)


device-types/Cisco/IR1101-K9.yaml
```diff
@@ -0,0 +1,35 @@
+---
+manufacturer: Cisco
+model: IR1101-K9
+slug: cisco-ir1101-k9
+part_number: IR1101-K9
+u_height: 0
+is_full_depth: false
```
harryajc, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

bezops (commented):
added weight / weight_unit

PR #4566, opened 2026-09-08
Update Cisco Catalyst 9200CX, C9400 and C9600 device types

f-kessler:
Addition of C9200CX-8UP-2XH and Update of interface names on C9200CX-8UXG-2X


.gitignore
harryajc, line 1: What is the reason for this file change?

harryajc (changes requested):

.gitignore
f-kessler: This change went by mistake into the commit. I will try to remove it. Sorry.

harryajc (approved):

PR #4571, opened 2026-09-09
Add 588 Hanwha IP camera device types

ch-bas:
Add 588 Hanwha IP camera device types covering outdoor/indoor PTZ, fixed, IR, and vandal-resistant models across the QND, QNO, QNV, SNO, SNV, XND, XNO, XNV, and related series.

Each definition includes:
- Manufacturer, model, slug, part number
- Weight (g)
- PoE class → poe_mode/poe_type mapped per IEEE 802.3 standard, or DC power port where applicable
- Ethernet interface speed (100/1000 base-T)
- Manufacturer datasheet link in comments

## Disclosure
These definitions were generated from cctv-database.com, a CC0 camera-spec database I maintain, where every entry is manually verified against the manufacturer PDF. Happy to adjust naming or fields to match the library's conventions. More vendors can follow as separate PRs if this is welcome.


harryajc (changes requested):
A Few of these devices have AC and DC power in. Is this correct, I've never seen a CCTV camera with a C14 power port on it.

ch-bas:
Fixed — removed all AC (C14) power ports. PoE+AC cameras now rely on PoE only. The 20 large PTZ cameras that were purely AC-powered (no PoE, no DC — e.g. PNP-9200RH, SNP-3301) now use `type: hardwired` to indicate direct-wired mains power, which seems more accurate than a C14 connector.

ch-bas:
AC power ports (iec-60320-c14) have been removed from all camera device types. DC-only cameras retain their DC terminal port; PoE-only cameras use interface-level PoE attributes; AC-only PTZ cameras now use `type: hardwired`.

PR #4572, opened 2026-09-09
Add 351 i-PRO IP camera device types

ch-bas:
Add 351 i-PRO IP camera device types covering the M-series, WV-S, WV-U, WV-X, WV-NF, WV-SC, and related series (outdoor/indoor fixed, PTZ, fisheye, multisensor).

Each definition includes:
- Manufacturer, model, slug, part number
- Weight (g)
- PoE class → poe_mode/poe_type mapped per IEEE 802.3 standard where class is known; DC/AC power ports where applicable
- Manufacturer datasheet link in comments

2 models already in the library (WV-S66600-Z3L, WV-U15550-V3L) were skipped to avoid duplicate slugs.

## Disclosure
These definitions were generated from cctv-database.com, a CC0 camera-spec database I maintain, where every entry is manually verified against the manufacturer PDF. Happy to adjust naming or fields to match the library's conventions. More vendors can follow as separate PRs if this is welcome.


harryajc (changes requested):
A Few of these devices have AC and DC power in. Is this correct, I've never seen a CCTV camera with a C14 power port on it.

ch-bas:
Fixed — removed all AC (C14) power ports. You're right that CCTV cameras connect to mains through an external adapter, not directly via a C14 port on the body. PoE+AC cameras now rely on PoE only; no i-PRO cameras in this set were purely AC-only.

harryajc (approved):

PR #4575, opened 2026-09-09
Apc smt2200rm2u front and rear elevations

vicbreen:


elevation-images/APC/apc-smt2200rm2u.rear.jpg
harryajc, line 1: Like the other image provided there should be no empty or white space around the device elevation.

harryajc (changes requested):

harryajc (approved):

PR #4612, opened 2026-09-21
Add TL-SG1008P

Trolldemorted:


device-types/TP-Link/TL-SG1008P.yaml
```diff
@@ -0,0 +1,35 @@
+---
+manufacturer: TP-Link
+model: TL-SG1008P
+slug: tp-link-tl-sg1008p
+comments: '[TP-Link TL-SG1008P](https://www.tp-link.com/en/business-networking/soho-switch-poe/tl-sg1008p/#specifications)'
+is_full_depth: false
+u_height: 0
```
harryajc, line 7: Hello, thank you for your contribution! In an attempt to have the most accurate and complete data we can, I have a few additional data points we would like you to add if possible.

harryajc (changes requested):

device-types/TP-Link/TL-SG1008P.yaml
Trolldemorted: Hey, thanks for the feedback!

The SG1008P is a small desktop switch, it is not a 19" device with a dedicated airflow direction. It also has multiple hardware revisions so multiple different weights are given, should I just pick the weight from one of the statements of a vendor?

grafik

harryajc: The best option is probably the latest. For airflow if it is fanless then you should use `airflow: passive`

